PR #23, opened 2021-03-16
Bugfix/error when data_server_port (disregarded key) is empty in instrument configuration

dsavchenko:


volodymyrss:
thanks! this makes sense. But this this is really a bug fix, the incomplete transition leaving unused keys needs to be completed.

volodymyrss (approved):
thanks, this is ok, but it's a patch to larger incomplete transition

dsavchenko:
> thanks! this makes sense. But this this is really a bug fix, the incomplete transition leaving unused keys needs to be completed.

Ok, I will now deal with other keys. This one was just explicitly disregarded

volodymyrss:
> > thanks! this makes sense. But this this is really a bug fix, the incomplete transition leaving unused keys needs to be completed.
> 
> Ok, I will now deal with other keys. This one was just explicitly disregarded

normally, there are few others also disregarded, I outlined here:

https://github.com/oda-hub/dispatcher-app/issues/24

volodymyrss (changes requested):
actually, I am not sure these additional ones ones can be circumvented.
It's better that the dispatcher crashes on starting without them.

Although, it seems they would need to belong to the plugin?

dsavchenko:
As far as I understand, this class deals with the configuration of an instrument. It should not crash the dispatcher itself.

Anyway, None values are processed in the constructor, so let them go there.

volodymyrss:
if instrument is misconfigured it should crash. 
Error of configuration should not go unnoticed! One way or another. You can also suggest another way to clearly indicate problematic configuration.
Or is it not problematic?

volodymyrss:
since it is a little bit implicit here what is the situation that causes the problem, could you maybe write a test then, so we see what we should define as expected behavior?

PR #29, opened 2021-03-16
make DataServerConf more instrument-agnostic 

dsavchenko:
This allows using not only hardcoded config keys in dataserver config dict. The error raises only if one of the explicitly required keys is missed.


burnout87:
What about providing the list of required keys directly with the configuration of the queried instrument? 
Then pass such a list to the constructor of the DataServerConf. What do you think?

Then I noticed that the constructor will check the _dispatcher_mnt_point_ parameter, the _data_server_remote_cache_ and the __data_server_port__. Can those three treated the same way those included in the list of required keys are ?

thanks

dsavchenko:
> What about providing the list of required keys directly with the configuration of the queried instrument?
> Then pass such a list to the constructor of the DataServerConf. What do you think?

The logic here is that the keys in the input are themselves probably come from the configuration of the instrument. If we expect error in config file, it can be also in this list of required keys. 

Another way could be to hardcode this list somewhere in the plugin and pass it when calling this class.

But the idea was to put here the keys that are unambiguously required by any instrument plugin.

dsavchenko:
> Then I noticed that the constructor will check the _dispatcher_mnt_point_ parameter, the _data_server_remote_cache_ and the **data_server_port**. Can those three treated the same way those included in the list of required keys are ?

These keys require some special behaviour (migrated from previous realization)
e.g.
*data_server_port* just disregarded but currently only throw warning, not exception
*dispatcher_mnt_point* causes the creation of the directory 
*data_server_remote_cache* is named differently from object attribute
etc.

burnout87:
Ok, so are those keys always necessary? Because if that is so, then those can stay as they are right now ()and being initialized), but more instrument-specific ones being loaded with the configuration by filling-up te required_keys list.

What do you think?

dsavchenko:
I think that they are not required universally. That's why there are 'is not None' checks. 
But we should ask @volodymyrss

burnout87:
> I think that they are not required universally. That's why there are 'is not None' checks.
> But we should ask @volodymyrss

What I meant was actually related as to if it is ok to check for the presence of those parameters in the **DataServerConf** init, since they are not instrument specific. The instument specific ones are loaded inside the required_keys list

dsavchenko:
My intentions are completely opposite. Instrument-independent keys (common to any instr.) in required_keys. 
Then (possibly instrument specific, but could work simarly with different instruments) keys requiring special attention. 
Then just all others.

I currently don't want to change the interface of the class (like requiring to provide the list of keys). It will pull out changes to all plugins using it

burnout87:
Ok, now I see, it makes sense. I wanted to use the same approach, just used different terminology.

cdci_data_analysis/configurer.py
```diff
@@ -33,89 +33,62 @@
...
 
-
-# TODO: this looks rather specific to INTEGRAL?
-
 class DataServerConf:
 
-    def __init__(self, data_server_url,
-                 data_server_remote_cache=None,
-                 dispatcher_mnt_point=None,
-                 dummy_cache=None,
```
volodymyrss, line 44: dummy_cache  is not entirely optional. in the sense that we rather want this to be set in production.
it might be useful to allow it to be None in test. But it should at least show a warning, or ideally raise a problem if server is not started in some "debug" mode (the same mode that reports extra details in run_analysis issues)

volodymyrss, line 44: dummy_cache is not entirely optional. also products_url. they are also used by all instruments.

It is true that current implementation would allow it anyway, and fail in dummy request, and make a problem with product url.
so we rather want them to be always set in production, but it might be useful to allow it to be None in test (or in current Kyiv deployment).  

But this situation should at least show a warning, or ideally raise a problem if server is not started in some "debug" mode (the same mode that reports extra details in run_analysis issues)

 data_server_remote_cache and dispatcher_mnt_point are the only two INTEGRAL specific here.

tests/test_configurer.py
```diff
@@ -0,0 +1,36 @@
...
+                 'data_server_remote_cache': 'reduced/ddcache',
+                 'dummy_cache': 'data/dummy_prods'
+                 }
+    bad_dict = {'data_server_url': 'https://data-server:5000',
+                 'data_server_port': '5000',
+                 'dispatcher_mnt_point': 'data',
+                 'data_server_cache': 'reduced/ddcache',
+                 'dummy_cache': 'data/dummy_prods'
+                 }
+    ugly_dict = {'data_server_host': 'data-server',
```
volodymyrss, line 18: nice! but if we provide data_server_host and data_server_url it will say nothing of the wrong value, right?

volodymyrss (changes requested):
Thanks! I think it's a good change, but it's a bit too dynamic. Would it be useful to make some constrains?

cdci_data_analysis/configurer.py
```diff
@@ -33,89 +33,62 @@
...
-            self.dispatcher_mnt_point = os.path.abspath(dispatcher_mnt_point)
+        if conf.get('dispatcher_mnt_point', None) is not None:
+            self.dispatcher_mnt_point = os.path.abspath(conf.pop('dispatcher_mnt_point'))
             FilePath(file_dir=self.dispatcher_mnt_point).mkdir()
         else:
             self.dispatcher_mnt_point = None
 
-        self.products_url = products_url
-
         if self.dispatcher_mnt_point is not None and self.data_server_remote_path is not None:
```
volodymyrss, line 71: Ironically considering the intent to reduce instrument dependence, the only parameter that are still treated explicitly are INTEGRAL (clearly since special treatment had to be preserved, unfortunatly, at least for now)

cdci_data_analysis/configurer.py
```diff
@@ -33,89 +33,62 @@
...
-            #self.data_server_cache = os.path.join(self.dispatcher_mnt_point, self.data_server_remote_path)
             self.data_server_cache = os.path.join(
                 self.dispatcher_mnt_point, self.data_server_remote_path)
         else:
             self.data_server_cache = None
 
+        #optional config keys
+        for key in conf:
+            self.__setattr__(key, conf[key])
```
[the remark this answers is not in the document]
burnout87: To preserve the compactness of the code, we could define, as indeed suggested, a place for each instrument (plugin seems indeed the best place) where the instrument-specific keywords are defined.

And we could use the same approach for those mandatory ones, simply placing those somewhere else in a dispatcher configuration.

Then, we could separate the loading of those two sets of keys, (eg general and instrument-specific), and treat them in separate ways.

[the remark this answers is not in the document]
volodymyrss: indeed!

there are two key classifications: required/optional and dispatcher/plugin. 2x2=4 options in total. 
But I do not think there should be any key which is not validated by some of these lists.

It is probably a healthy practice if typos, mistakes, misunderstandings in the config  are noticed.

We could even use config schema (e.g. with python schema-config, which addresses other concerns, e.g. secret and templating handling, or similar) instead of defining programmatically.
But at this point it would be easier not to go into that, unless someone is curious. But we might want to finish this request.

At this point, we can just classify currently set values as follows:

**required, common**
data_server_url 

**required in production**
dummy_cache
products_url

**required, integral, special treatment**
dispatcher_mnt_point
data_server_remote_cache

**to discard**:
data_server_port

everything else may be validated with extra, optional (to preserve interface) lists provided by the plugin

volodymyrss:
I realize the requested changes could be much, if it is beyond @dsavchenko  intent. We could as well merge this one as it is, create **an issue for the following change**, and make another feature.

volodymyrss:
I realize the requested changes could be much, if it is beyond @dsavchenko  intent. We could as well merge this one as it is, create **an issue for the following change**, and make another feature, maybe even by @burnout87 

So please @dsavchenko let us know what you think: are you interested to discuss these further complications in this PR and/or implement these developments?

dsavchenko:
> I realize the requested changes could be much, if it is beyond @dsavchenko intent. We could as well merge this one as it is, create **an issue for the following change**, and make another feature, maybe even by @burnout87
> 
> So please @dsavchenko let us know what you think: are you interested to discuss these further complications in this PR and/or implement these developments?

I will try to implement at least some of the comments.

dsavchenko:
I have made some changes to the class. Unit tests still need to be updated.

burnout87 (commented):
Can we consider the option of having those INTEGRAL specific parameters coming also from the list of required ones and treated in the same way? so, not to have those hard-coded? Perhaps not now, but for the future might become useful.

cdci_data_analysis/configurer.py
```diff
@@ -33,89 +33,97 @@
...
+                            f"None value of the required configuration key {key} is only allowed in debug mode")
+                    logger.warning(
+                        f"required configuration key {key} is None")
+                self.__setattr__(key, value)
+            except KeyError as e:
+                logger.error(
+                    f"problem constructing {self}: {key} configuration key is required")
+                raise e
+
+        if conf.pop('data_server_port', None) is not None:
```
volodymyrss, line 100: minor, but why not make a list of two obsolete keys and slighly reduce code redundancy?

cdci_data_analysis/configurer.py
```diff
@@ -33,89 +33,97 @@
...
-            logger.warning(
-                f"problem constructing {self}: dispatcher_mnt_point is None: {dispatcher_mnt_point}")
+        logger.info("building config from %s", conf)
 
-        self.data_server_url = data_server_url
 
-        # dummy prods local cache
-        self.dummy_cache = dummy_cache
+        try:
+            self.__setattr__('data_server_url', conf.pop('data_server_url'))
```
volodymyrss, line 55: just out of curiosity, why __setattr__ here?

cdci_data_analysis/configurer.py
```diff
@@ -33,89 +33,97 @@
...
+    def __init__(self, required_keys=None, allowed_keys=None, **kwargs):
+        if required_keys is None:
+            #temporary hardcode to preserve interface
+            required_keys = ['data_server_url', 'dummy_cache']
 
-        if data_server_url is None:
-            logger.warning(
-                f"problem constructing {self}: data_server_url is None: {data_server_url}")
+        if allowed_keys is None:
+            #temporary hardcode to preserve interface
```
volodymyrss, line 44: I might be wrong or misunderstand, but is the default interface-preserving set of allowed but optional keys a different one? specifically, those before?

cdci_data_analysis/configurer.py
```diff
@@ -33,89 +33,97 @@
...
+        # NOTE: these are required for integral, so need to be processed before for-cycle for the other required ones
+        if conf.get('data_server_remote_cache', None) is not None:
+            # path to dataserver cache
+            self.data_server_remote_path = conf.pop('data_server_remote_cache')
+        else:
+            self.data_server_remote_path = None
 
-        if dispatcher_mnt_point is not None:
-            self.dispatcher_mnt_point = os.path.abspath(dispatcher_mnt_point)
+        if conf.get('dispatcher_mnt_point', None) is not None:
```
volodymyrss, line 70: frankly I would take out the entire special treatment part in a separate function, and maybe get rid of it later when feasible. to make this clear and nice function more clear and nice

and also put it at the end. the two integral-specific keys are also allowed by optional, by interface-preserving standard

volodymyrss:
> Can we consider the option of having those INTEGRAL specific parameters coming also from the list of required ones and treated in the same way? so, not to have those hard-coded? Perhaps not now, but for the future might become useful.

indeed, I wrote something similar in the review and then saw this comment.

cdci_data_analysis/configurer.py
dsavchenko: I do not understand this comment

cdci_data_analysis/configurer.py
dsavchenko: Copy-pasted) I will change this to an assignment for clarity

cdci_data_analysis/configurer.py
dsavchenko: Of course. Will do it

cdci_data_analysis/configurer.py
dsavchenko: It could be taken into a separate method. But it must be called in the constructor before other required keys, not at the end. the reason for this is that these keys should be pop-ed from dict to not being processed in general for-loop

burnout87: I agree, it is good to have it in a separate method, in order to get rid of it completely in the future

volodymyrss (approved):

dsavchenko:
Should @burnout87 also approve this to finally merge it?

burnout87 (approved):
There you go!

volodymyrss:
great, thanks!

PR #30, opened 2021-03-23
actually compatible with previous plugin interfaces

volodymyrss:
turns out it was not really yet, but almost there!

### DIFF
--- a/cdci_data_analysis/configurer.py
+++ b/cdci_data_analysis/configurer.py
@@ -4,13 +4,15 @@ import traceback
 from builtins import (bytes, str, open, super, range,
                       zip, round, input, int, pow, object, map, zip)
 
-from cdci_data_analysis import conf_dir
-from cdci_data_analysis.analysis.io_helper import FilePath
-import yaml
+from typing import Union, List
 
 import sys
 import os
 import logging
+
+from cdci_data_analysis import conf_dir
+from cdci_data_analysis.analysis.io_helper import FilePath
+import yaml
 
 __author__ = "Andrea Tramacere"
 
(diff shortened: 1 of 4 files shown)


cdci_data_analysis/configurer.py
```diff
@@ -46,6 +46,9 @@ def __init__(self, required_keys=None, allowed_keys=None, **kwargs):
         if allowed_keys is None:
             #temporary hardcode smth to preserve interface
             allowed_optional_keys = ['data_server_cache']
+
+            # preserving interface requires listing these keys, currently provided by all plugins (to be updated)
+            allowed_optional_keys += ['data_server_remote_cache', 'dispatcher_mnt_point']
```
dsavchenko, line 51: These keys are implicitly allowed in process_integral_keys
So this will change nothing in the behaviour, but at least to have it declared explicitly is good for clarity.

tests/test_configurer.py
```diff
@@ -74,4 +74,15 @@ def test_dsconf_pass_keys():
...
+def test_dsconf_legacy_plugin_keys():
+    conf_dict = {
+                 'data_server_url': 'bacon',
+                 'data_server_port': 'eggs',
+                 'data_server_remote_cache': 'spam',
+                 'dispatcher_mnt_point': 'spam',
+                 'dummy_cache': 'spam!',
+                }
+
+    conf = DataServerConf.from_conf_dict(conf_dict)
```
dsavchenko, line 88: What is the addition compared to test_dsconf_integral?

cdci_data_analysis/configurer.py
volodymyrss: but it does change the interface!

if the key is None, since None is handed in process_integral_keys the same way as unset key, and then it fails because it is not allowed.
and None is routinely provided by plugins, apparently.

it could be handled by replacing get with pop, but this means that yet another argument of process_integral_keys will be mutating, leading for possible further confusion.

volodymyrss: I show it with the updated test below.

tests/test_configurer.py
volodymyrss: Ah I put the test quickly at the end, which is unfortunate, since usually I would show the problem with a failing test and then make a fix. Here is I started the right way: https://github.com/oda-hub/dispatcher-app/pull/31 - it is just for the test. Then we can deal with the fix of #30  here.

dsavchenko (approved):

burnout87 (approved):

volodymyrss:
thanks for the reviews!

PR #31, opened 2021-03-24
test for interface incompatibility, two instances

volodymyrss:
let's start bug fixes with a test! else it's unclear


dsavchenko (approved):
Now it is clear!

volodymyrss:
@burnout87 could you please review this one? It's not just for approve but also to keep in touch. Since you were involved.
Sorry if you wanted to do it later, this is not urgent, just want to be sure the request is not forgotten.

burnout87 (approved):
Much more clear now, perhaps some more test for integral can be added in the future.
Sorry for the delay.

Issue #19, opened 2021-03-16
implement filtering of requests


volodymyrss:
following meeting on 2021-03-15, and the description in https://github.com/oda-hub/dispatcher-app/blob/staging-1-3/interfaces.md#usage we can start implementing the scheme for role-based handling of the requests. There several flows described there, we need to clarify them for ourselves and see how to address them progressively.

burnout87:
Here some ideas:

According to what explained [here](https://github.com/oda-hub/dispatcher-app/blob/master/plugin-interface.md), but also implemented [here](https://github.com/oda-hub/dispatcher-app/pull/29), two approaches can be followed for the user-role interaction dispatcher-plugins:

1. provide an implementation of the instrument **configuration** loading within the plugin, and a list of required query-role pairs inside the instrument configuration: these roles will be then loaded when the instrument configuration is actually loaded within the dispatcher. This can become quite tricky the moment we want to implement a more fine-grained configuration that could specify the required role for each query


2. Expand the **instrument factory** by defining an object that provides a number of pairs role-query, to pass to the instantiation of the instrument object, and to use anytime a specific query for a given instrument is performed: checking the role within the token along with the requested query will let a certain query, of a certain instrument, assessed if it can be performed for a user (a similar object already exists: [example](https://github.com/oda-hub/dispatcher-plugin-integral/blob/fe2ccb7279b90d0c378a5e359919931cc873a5bf/cdci_osa_plugin/osa_isgri.py#L89)). However, here we would have to change and expand every plugin code

burnout87:
In order to limit the possibilities directly over the interface for a certain user, the login procedure should be handled within the dispatcher in order to apply (if needed) a _restricted_ view for a user depending on its role.

Does this make any sense?

Perhaps @motame can shed some light on this point...

volodymyrss:
>  1.    provide an implementation of the instrument configuration loading within the plugin, and a list of required query-role pairs inside the instrument configuration: these roles will be then loaded when the instrument configuration is actually loaded within the dispatcher. This can become quite tricky the moment we want to implement a more fine-grained configuration that could specify the required role for each query

I think it is rarely necessary or to have fine-grained control by query but not detailed control by query parameters.

So I would reduce this option to a single config setting: the roles required by any queries in this plugin. Perhaps we should assume by default that all roles listed need to be satisfied.
It may be a universally required option. For full-public plugins the value will be empty list of required roles.

This way, older plugins may be loaded without change.

[...]

Issue #25, opened 2021-03-16
clarify exceptions returned by the dispatcher


volodymyrss:
a lot of the time in the current dispatcher the exception flows are not particularly untested, and occasionally fail within the exception handling.   Due to time constrains and organizational priorities no sufficient refactoring was made for this purpose.

More recently, some effort was made to make sure  returned errors as somewhat meaningful, but this effort is far from complete.

In general, dispatcher has two sorts of exception situations: 
* some with the query responses, with things like [this](https://github.com/oda-hub/dispatcher-app/blob/master/cdci_data_analysis/flask_app/dispatcher_query.py#L772). This kind contains also exceptions from backends.
* and some handled by the flask high-level interception like this https://github.com/oda-hub/dispatcher-app/blob/staging-1-3/cdci_data_analysis/flask_app/app.py#L153

we would want to explain which kinds of exceptions can happen - most of all with exception class hierarchy. 
Two hierarchies - for each of the two kinds.

And we need to be sure that
* user gets meaningful but not too confusing/revealing message.
* we get more detailed message (in sentry and in log)

I would also argue that we should very clearly distinguish unexpected errors from expected.

volodymyrss:
I previously note some general experience and principles of our condition in here: https://github.com/oda-hub/doc-ops-reporting#issue-handling-principles . It derives from the user stories and practical experience of operating and using these platforms.

burnout87:
Starting some investigation that will help me in understanding more the architecture and how to improve the exceptions handling.

One first question:

- [Here](https://github.com/oda-hub/dispatcher-app/blob/bc2be600665c090bd0b9172a4b2cb4b1aad48dc2/cdci_data_analysis/flask_app/dispatcher_query.py#L773) a new job is built, set it as failed, but not used for anything else: is there some other reason why building a job is needed in that case? Also, such a function, currently, is used only in case of a failed token validation...perhaps it can be used somewhere else. Something I am trying to understand.

volodymyrss:
> Starting some investigation that will help me in understanding more the architecture and how to improve the exceptions handling.

honorable goal!
 
> One first question:
> 
>     * [Here](https://github.com/oda-hub/dispatcher-app/blob/bc2be600665c090bd0b9172a4b2cb4b1aad48dc2/cdci_data_analysis/flask_app/dispatcher_query.py#L773) a new job is built, set it as failed, but not used for anything else: 

[...]

PR #32, opened 2021-03-24
Role based queries

burnout87:
First implementation of what discussed #19

### DIFF
--- a/cdci_data_analysis/analysis/exceptions.py
+++ b/cdci_data_analysis/analysis/exceptions.py
@@ -24,7 +24,7 @@ class APIerror(Exception):
 
 class BadRequest(APIerror):
     def __init__(self, message, status_code=None, payload=None):
-        super().__init__(message, status_code=400, payload=payload)
+        super().__init__(message, status_code=status_code, payload=payload)
 
 
 class RequestNotUnderstood(BadRequest):
@@ -37,7 +37,8 @@ class RequestNotUnderstood(BadRequest):
 
 
 class RequestNotAuthorized(BadRequest):
-    pass
+    def __init__(self, message):
+        super().__init__(message, status_code=403)
 
 
 class MissingParameter(RequestNotUnderstood):
--- a/cdci_data_analysis/analysis/exceptions.py
+++ b/cdci_data_analysis/analysis/exceptions.py
@@ -33,6 +33,9 @@ class RequestNotUnderstood(BadRequest):
     it is unclear for us and we want user to please clarify!
     """
 
+class RequestNotAuthorized(BadRequest):
+    pass
+
 class MissingParameter(RequestNotUnderstood):
     pass
 
--- a/cdci_data_analysis/analysis/exceptions.py
+++ b/cdci_data_analysis/analysis/exceptions.py
@@ -21,9 +21,11 @@ class APIerror(Exception):
     def __str__(self):
         return repr(self)
 
+
 class BadRequest(APIerror):
     def __init__(self, message, status_code=None, payload=None):
         super().__init__(message, status_code=400, payload=payload)
+
 
 class RequestNotUnderstood(BadRequest):
     """
@@ -33,13 +35,14 @@ class RequestNotUnderstood(BadRequest):
     it is unclear for us and we want user to please clarify!
     """
 
+
 class RequestNotAuthorized(BadRequest):
     pass
+
 
 class MissingParameter(RequestNotUnderstood):
     pass
 
-###
 
 class UnfortunateRequestResults(Exception):
     """
@@ -48,6 +51,7 @@ class UnfortunateRequestResults(Exception):
     it is not bad request
     it is unclear for us and we want user to please clarify!
     """
+
 
 class ProblemDecodingStoredQueryOut(Exception):
     """
--- a/cdci_data_analysis/analysis/instrument.py
+++ b/cdci_data_analysis/analysis/instrument.py
@@ -177,7 +177,6 @@ class Instrument:
         if query_out.status_dictionary['status']==0:
             query_out=self.set_catalog_from_fronted(par_dic, request,back_end_query,logger=logger,verbose=verbose,sentry_client=sentry_client)
 
-
         #set input products
         if query_out.status_dictionary['status'] == 0:
             try:
@@ -252,7 +251,7 @@ class Instrument:
     def check_instrument_query_role(self, query_name, roles):
         query_obj = self.get_query_by_name(query_name)
         if not query_obj.check_query_roles(roles):
-            raise RequestNotAuthorized("Roles %s cannot execute query %s", roles, query_name)
+            raise RequestNotAuthorized(f"Roles {roles} cannot execute query {query_name}")
         else:
             return True
 
--- a/cdci_data_analysis/analysis/instrument.py
+++ b/cdci_data_analysis/analysis/instrument.py
@@ -161,25 +161,23 @@ class Instrument:
                   api=False,
                   **kwargs):
 
-        self._current_par_dic=par_dic
-
         if logger is None:
             logger = self.get_logger()
 
-        #set pars
-        query_out=self.set_pars_from_form(par_dic,verbose=verbose,sentry_client=sentry_client)
+        # set pars values from the input parameters
+        query_out = self.set_pars_from_form(par_dic, verbose=verbose, sentry_client=sentry_client)
 
-        if verbose ==True:
+        if verbose:
             self.show_parameters_list()
 
-        #set catalog
+        # set catalog
         if query_out.status_dictionary['status']==0:
-            query_out=self.set_catalog_from_fronted(par_dic, request,back_end_query,logger=logger,verbose=verbose,sentry_client=sentry_client)
+            query_out=self.set_catalog_from_fronted(par_dic, request,back_end_query, logger=logger, verbose=verbose,sentry_client=sentry_client)
 
-        #set input products
+        # set input products
         if query_out.status_dictionary['status'] == 0:
             try:
-                query_out = self.set_input_products_from_fronted(par_dic, request,back_end_query,logger=logger,verbose=verbose,sentry_client=sentry_client)
+                query_out = self.set_input_products_from_fronted(par_dic, request,back_end_query, logger=logger,verbose=verbose,sentry_client=sentry_client)
             except Exception as e:
                 # FAILED
                 query_out.set_failed(product_type,
@@ -189,6 +187,7 @@ class Instrument:
                                      excep=e)
 
         self.logger.info('--> par dict', par_dic)
+        # logger.info('--> par dict', par_dic)
 
         if dry_run:
             job.set_done()
@@ -235,7 +234,7 @@ class Instrument:
                     # logger.error("run_query failed: %s", traceback.format_exc())
                     query_out.set_failed(product_type, logger=logger, sentry_client=sentry_client, excep=e)
 
-        #adding query parameters to final products
+        # adding query parameters to final products
         query_out.set_analysis_parameters(par_dic)
         query_out.set_api_code(par_dic)
         query_out.dump_analysis_parameters(out_dir,par_dic)
--- a/cdci_data_analysis/analysis/instrument.py
+++ b/cdci_data_analysis/analysis/instrument.py
@@ -186,8 +186,8 @@ class Instrument:
                                      sentry_client=sentry_client,
                                      excep=e)
 
-        self.logger.info('--> par dict', par_dic)
-        # logger.info('--> par dict', par_dic)
+        # self.logger.info('--> par dict', par_dic)
+        logger.info('--> par dict', par_dic)
 
         if dry_run:
             job.set_done()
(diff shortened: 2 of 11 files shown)


cdci_data_analysis/analysis/exceptions.py
```diff
@@ -33,10 +35,14 @@ class RequestNotUnderstood(BadRequest):
     it is unclear for us and we want user to please clarify!
     """
 
+
+class RequestNotAuthorized(BadRequest):
```
volodymyrss, line 39: this could as well return 403 status? more conventional

burnout87: Yes, makes sense.

cdci_data_analysis/analysis/instrument.py
```diff
@@ -69,26 +67,20 @@ def __init__(self,
                  data_server_query_class=None,
                  query_dictionary={}):
 
-        #name
-        self.name=instr_name
-
+        # name
+        self.name = instr_name
+        # logger
+        self.logger = logging.getLogger(repr(self))
```
volodymyrss, line 73: we should be careful that this does not lead to unconstrained growth of loggers, for every request. rigth now, it happens in other case - see logging_tree output.

just general point, here it might be fine

cdci_data_analysis/analysis/instrument.py
```diff
@@ -165,9 +145,8 @@ def test_has_input_products(self, config,instrument,logger=None):
         if self.data_server_query_class is not None:
             return self.data_server_query_class(config=config,instrument=self).test_has_input_products(instrument,logger=logger)
 
-
-
-    def run_query(self,product_type,
+    def run_query(self, product_type,
+                  query_name,
```
volodymyrss, line 149: why did this change? I sorry, I dont follow

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -1172,8 +1179,18 @@ def run_query(self, off_line=False, disp_conf=None):
...
             else:
-
+                query_name = self.instrument.get_product_query_name(product_type)
+                if not self.public:
+                    roles = []
+                    if self.token is not None and self.decoded_token is not None:
+                        roles = self.get_token_roles()
+                    # assess the permissions for the query execution
+                    try:
+                        self.instrument.check_instrument_query_role(query_name, roles)
```
volodymyrss, line 1189: it will need par_dict I think, to make filtering by them

volodymyrss (changes requested):
several points listed. is that can we have an example of filtering based on parameters, a fake one, in empty instrument?

cdci_data_analysis/analysis/exceptions.py
burnout87: I am struggling to understand where to set the final staus_code (403 in this case) for the response that should be returned. Is there somewhere a similar case? I noticed that every response returned when building this returns 200

volodymyrss: A change in there https://github.com/oda-hub/dispatcher-app/blob/e2fa1fe1622bbd9171e23a1cf48f4b084a7bb6f5/cdci_data_analysis/flask_app/dispatcher_query.py#L566 might be needed.

alternatively, we can just let it go there https://github.com/oda-hub/dispatcher-app/blob/e2fa1fe1622bbd9171e23a1cf48f4b084a7bb6f5/cdci_data_analysis/flask_app/app.py#L158

which already handles status codes for exceptions

for auth, second option can do.

but first option may be improved to handle status in any case.

two cases relevant discussed in there:

https://github.com/oda-hub/dispatcher-app/issues/25

burnout87: The first option is probably a better ( I was already trying that way), have to figure our better what I am missing.

As for the second, in this case, that errorhandler is not executed, since we directly return the Response object.

cdci_data_analysis/analysis/instrument.py
burnout87: Reason why I did this is because there were spme _print_ within the class. So it is something to keep an eye in the future for sure, but currently I used it only to replace the print that were already there.

cdci_data_analysis/analysis/instrument.py
burnout87: I moved it here since I was in need to extract the name of the query from the product type, and in order not to repeat the same operation twice I decide to pass it as an argument. I will probably roll back, since I realized that there is not such a big advantage with it.

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: You mean to have more information for evaluate the permission of the user over a certain query?

cdci_data_analysis/analysis/instrument.py
volodymyrss: of course, and it makes sense. 
but note an example of this: https://github.com/oda-hub/dispatcher-app/blob/master/cdci_data_analysis/flask_app/dispatcher_query.py#L269
 it can be avoided though

cdci_data_analysis/flask_app/dispatcher_query.py
volodymyrss: yes, we need to also filter by parameters of the specific request.

cdci_data_analysis/analysis/instrument.py
burnout87: That looks like a proper _logger_ configuration, so perhaps some reformatting of the code and in particular on how the logger is initiated (at least for the _dispatcher_quey_) is needed.

cdci_data_analysis/analysis/instrument.py
burnout87: #33 opened for the purpose

cdci_data_analysis/analysis/instrument.py
burnout87: #33  opened for the purpose

burnout87:
> several points listed. is that can we have an example of filtering based on parameters, a fake one, in empty instrument?

Working on it now

volodymyrss:
> > > several points listed. is that can we have an example of filtering based on parameters, a fake one, in empty instrument?
> 
> > Working on it now
> 
> So with the last commit I created a new type of numerical query which accepts only one parameter (to play around a bit, so to understand more in-depth the architecture), then the user-role is checked against those provided.
> 
> A test is also there

Ok!
To be sure we are on the same page, once you are ready for us to look again, you re-request review, ok? Then the "change requested" status will change, and it will appear on my https://github.com/pulls/review-requested
Conversely, do look on your own https://github.com/pulls/review-requested to see if there are any reviews you are happy to close.
Thanks!

burnout87:
> > > > several points listed. is that can we have an example of filtering based on parameters, a fake one, in empty instrument?
> > 
> > 
> > > Working on it now
> > 
> > 
> > So with the last commit I created a new type of numerical query which accepts only one parameter (to play around a bit, so to understand more in-depth the architecture), then the user-role is checked against those provided.
> > A test is also there
> 
> Ok!
> To be sure we are on the same page, once you are ready for us to look again, you re-request review, ok? Then the "change requested" status will change, and it will appear on my https://github.com/pulls/review-requested
> Conversely, do look on your own https://github.com/pulls/review-requested to see if there are any reviews you are happy to close.
> Thanks!


If you have nothing else to add, would you approve it?
Also @dsavchenko , what do you think?

volodymyrss:
> > > > > several points listed. is that can we have an example of filtering based on parameters, a fake one, in empty instrument?
> > > 
> > > 
> > > > Working on it now
> > > 
> > > 
> > > So with the last commit I created a new type of numerical query which accepts only one parameter (to play around a bit, so to understand more in-depth the architecture), then the user-role is checked against those provided.
> > > A test is also there
> > 
> > 
> > Ok!
> > To be sure we are on the same page, once you are ready for us to look again, you re-request review, ok? Then the "change requested" status will change, and it will appear on my https://github.com/pulls/review-requested
> > Conversely, do look on your own https://github.com/pulls/review-requested to see if there are any reviews you are happy to close.
> > Thanks!
> 
> If you have nothing else to add, would you approve it?

I will look on it now, thanks.

cdci_data_analysis/analysis/queries.py
```diff
@@ -257,25 +245,26 @@ def print_form_dictionary_list(self,l):
...
 
         for par in self._parameters_list:
             l.append(par.reprJSON())
 
-
         return json.dumps(l)
 
+    # Check if the given query cn be executed given a list of roles extracted from the token
+    def check_query_roles(self, roles, par_dic):
+        return True, []
```
volodymyrss, line 258: what's the second argument? In general, I think it is a bit confusing when tuples are returned without annotation. I like to use dict in a case like this with informtive keys.
Or type annotations (with typing module).  But in this case it is indeed not clear  for me what it means without reading more.
but maybe just a comment would do, for us

cdci_data_analysis/plugins/dummy_instrument/data_server_dispatcher.py
```diff
@@ -60,4 +60,42 @@ def get_dummy_products(self, instrument, config=None, **kwargs):
...
+        return []
+
+    def process_product_method(self, instrument, prod_list, api=False, **kw):
+        query_out = QueryOutput()
+        return query_out
+
+    # example with the general user role
+    def check_query_roles(self, roles, par_dic):
+        if 'p' in par_dic.keys():
+            # not sure this is actually the best way to obtain a certain parameter value
```
volodymyrss, line 96: me neither.
there is this parameter class there, maybe it's good to use that. but it's not critical here.

volodymyrss (approved):
I have no particularly relevant comments.
If it passes tests should be ok, I might miss something.

I general, we should probably converge on style, maybe make a single PR on it.
Else it's a bit harder to see which changes are functional.

volodymyrss:
We might or might not always review all 3 of us: I think it is always useful if done fast (so we do not wait for reviews more than necessary), but it depends if @dsavchenko is interested, it's extra work for him obviously.
At least you and I need to review all, since you develop this principally and I take care of our core deployment and product needs.

Maybe let's wait for @dsavchenko here, and ask his confirmation if he wants to help with reviews for parts he did not (yet?) touch.

dsavchenko (approved):
Looks ok in general. I have no comments added to the @volodymyrss ones. But I did not dive really deep into it.

volodymyrss:
> Looks ok in general. I have no comments added to the @volodymyrss ones. But I did not dive really deep into it.

cool! 

So, but, you want to be reviewer for all things? It's really kind of a duty, it's great if you can.

cdci_data_analysis/analysis/queries.py
burnout87: Since we want to show the list of the required role(s) for a given query in case this is unauthorized (ie [](https://github.com/oda-hub/dispatcher-app/blob/2be8a88c68a621ba1caf462aeb070054c80d2fb0/cdci_data_analysis/analysis/instrument.py#L254)), I return this directly from the method that is supposed to check for it (eg [another example](https://github.com/oda-hub/dispatcher-app/blob/2be8a88c68a621ba1caf462aeb070054c80d2fb0/cdci_data_analysis/plugins/dummy_instrument/data_server_dispatcher.py#L94)).

But I agree that actually, a dictionary is more informative.

dsavchenko:
> Maybe let's wait for @dsavchenko here, and ask his confirmation if he wants to help with reviews for parts he did not (yet?) touch.

I may be not very helpful in some cases, I am still not sure I understand the internals. 
But generally, I am interested to be more or less involved in what's going on.

cdci_data_analysis/analysis/queries.py
volodymyrss: Indeed, makes sense, and I think it was even me who asked for this functionality!
But I did not remember.

volodymyrss:
> > Maybe let's wait for @dsavchenko here, and ask his confirmation if he wants to help with reviews for parts he did not (yet?) touch.
> 
> I may be not very helpful in some cases, I am still not sure I understand the internals.
> But generally, I am interested to be more or less involved in what's going on.

Ok. Let's also discuss on the Monday meeting, since it has to do with duties and time allocation.

cdci_data_analysis/plugins/dummy_instrument/data_server_dispatcher.py
burnout87: Yes, and that class is actually used later 
[](https://github.com/oda-hub/dispatcher-app/blob/2be8a88c68a621ba1caf462aeb070054c80d2fb0/cdci_data_analysis/analysis/instrument.py#L168)

I will see if it can be used so to make it more uniform and consistent

burnout87: Uses it now, much more consistent and cleaner approach

cdci_data_analysis/analysis/queries.py
burnout87: It returns now a dictionary, easier to handle and less error-prone

volodymyrss (approved):
Seems ok! But I did not look deep. I think we can go with this.

dsavchenko (approved):

PR #34, opened 2021-03-25
Issues on deployment

volodymyrss:


burnout87:
What is the goal of this ? Is there some issue on deplyment?

volodymyrss:
> What is the goal of this ? Is there some issue on deplyment?

Indeed there are! but consider that this is draft (not ready for review) and review is not requested. I will probably separate it in a couple of different ones.

burnout87:
All right! Let me know so I can also get familiar with the deployment process

volodymyrss:
Well it's really about using it more more practically. Coverage is not too high yet. Need to add more tests for specific cases, appearing with real plugins, etc. Let's see on Monday.

volodymyrss:
I will close this and re-create

PR #35, opened 2021-03-26
more linting and coverage

volodymyrss:


dsavchenko (approved):

burnout87:
This is cool! Extremely powerful plugin!

burnout87 (approved):

volodymyrss:
yeah, it's very useful! I like a lot tools for static code analysis, they are very helpful. We can add a bit more too, later.
For tests, it's extremely useful, but dispatcher is developing a working test set only since half a year or so, so it's just about time.

PR #37, opened 2021-03-26
coverage instead of pytest-cov

volodymyrss:


volodymyrss:
It did not report coverage correctly for some reason! I changed the approach. Strange, the previous one works in other places.

PR #39, opened 2021-03-29
Log refactoring

burnout87:
Refactoring ongoing

### DIFF
--- a/tests/test_server_basic.py
+++ b/tests/test_server_basic.py
@@ -299,22 +299,18 @@ def test_empty_instrument_request(dispatcher_live_fixture):
 
 def test_no_instrument(dispatcher_live_fixture):
     server = dispatcher_live_fixture
-    print("constructed server:", server)
-
-    c=requests.get(server + "/run_analysis",
-                   params=dict(
-                   image_type="Real",
-                   product_type="image",
-                   E1_keV=20.,
-                   E2_keV=40.,
-                   T1="2008-01-01T11:11:11.0",
-                   T2="2008-06-01T11:11:11.0",
+    logger.info("constructed server:", server)
+    c = requests.get(server + "/run_analysis",
+                     params=dict(
+                         image_type="Real",
+                         product_type="image",
+                         E1_keV=20.,
+                         E2_keV=40.,
+                         T1="2008-01-01T11:11:11.0",
+                         T2="2008-06-01T11:11:11.0",
                 ))
 
-    print("content:", c.text)
-
-    jdata=c.json()
-
+    logger.info("content:", c.text)
     assert c.status_code == 400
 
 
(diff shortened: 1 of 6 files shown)


burnout87:
@volodymyrss Remember if this TODO was discussed at some point in time? 
[https://github.com/oda-hub/dispatcher-app/blob/37a0eb86b8f92ebb5f935f69bb30154b16914a72/cdci_data_analysis/flask_app/dispatcher_query.py#L134](https://github.com/oda-hub/dispatcher-app/blob/37a0eb86b8f92ebb5f935f69bb30154b16914a72/cdci_data_analysis/flask_app/dispatcher_query.py#L134)

I think some log might still be useful, but perhaps without the scratch_dir

volodymyrss:
> @volodymyrss Remember if this TODO was discussed at some point in time?
> 
> https://github.com/oda-hub/dispatcher-app/blob/37a0eb86b8f92ebb5f935f69bb30154b16914a72/cdci_data_analysis/flask_app/dispatcher_query.py#L134-L138
> 
> I think some log might still be useful, but perhaps without the scratch_dir

I think it would be still good to store log in the individual scratch directory of the job. It allows to inspect the job for each request easily near data. In some cases it is still useful.
Except now it did not work, and caused some issues, I do not remember, perhaps with overcreation of loggers.

tests/test_server_basic.py
```diff
@@ -44,6 +44,20 @@
...
+
+    c=requests.get(server + "/api/meta-data",
+                   params={**default_params},
+                )
+
+    logger.info(f"content: {c.text}")
+    jdata=c.json()
+
+    assert c.status_code == 200
+
```
volodymyrss, line 66: can we add a test to check if log was created? because the problem was that it was not, that's why it was commented, I think

volodymyrss (changes requested):

tests/test_server_basic.py
burnout87: It was crashing mainly because the job_id was not set Non for that specific case, I will add the check

volodymyrss: and also that it does not create a new logger for each request. That's what was happening at some points too.
I find it is convenient to inspect loggers with logging_tree, but simply logging would do too.

volodymyrss: it will be necessary but not sufficient, since it will not catch an problem which happened in the wild. but it might be unclear what I am trying to explain with the logging, so let me maybe to that part myself.

volodymyrss:
I would not necessarily deal with this for 21.05

volodymyrss:
I just merged master and better coverage testing to see how far it is, and it is not too close anymore. Perhaps it's too much effort right now. I hope no merges were destructive.

burnout87:
> I just merged master and better coverage testing to see how far it is, and it is not too close anymore. Perhaps it's too much effort right now. I hope no merges were destructive.

Thanks! It is not necessary at the moment. So it can be labelled as an enhancement and we can deal with that at a later stage after the release.

PR #42, opened 2021-03-31
sometimes 'instrument' attribute is not set

volodymyrss:
this happened to me in live deployment. I did not produce a test, which would be advisabe

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -46,7 +46,7 @@ jobs:
 
     - name: Test with pytest
       run: |
-        python -m coverage run -m pytest  tests/test_server_basic.py -sv --full-trace --log-cli-level=DEBUG -m 'not isgri_plugin and not dda'
+        python -m coverage run --source=cdci_data_analysis -m pytest tests -sv --full-trace --log-cli-level=DEBUG -m 'not isgri_plugin and not dda'
 
     # I wonder how this interacts with the Github App. But this is not only for PR. Also this does not always work
     - name: Codecov
--- a/tests/test_job_management.py
+++ b/tests/test_job_management.py
@@ -13,6 +13,7 @@ def test_construct_for_callback(app):
     assert query
 
 
+@pytest.mark.xfail
 def test_callback_without_prior_run_analysis(dispatcher_live_fixture):
     server = dispatcher_live_fixture
     print("constructed server:", server)
--- a/tests/test_job_management.py
+++ b/tests/test_job_management.py
@@ -2,16 +2,6 @@ import requests
 
 import pytest
 
-def test_construct_for_callback(app):
-    from cdci_data_analysis.flask_app.dispatcher_query import InstrumentQueryBackEnd
-
-    query = InstrumentQueryBackEnd(
-              app, 
-              instrument_name='mock', 
-              data_server_call_back=True)
-
-    assert query
-
 def test_callback_without_prior_run_analysis(dispatcher_live_fixture):
     server = dispatcher_live_fixture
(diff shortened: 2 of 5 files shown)


burnout87:
Did you see which set of input parameters causes a crash?

volodymyrss:
yeah, we should add a test for it.

volodymyrss:
sorry my last comment was made before I saw yours.
So, I think it happens when a "callback" call happens.

it starts from here:

https://github.com/oda-hub/dispatcher-app/blob/master/cdci_data_analysis/flask_app/app.py#L264


and goes here:

https://github.com/oda-hub/dispatcher-app/blob/master/cdci_data_analysis/flask_app/dispatcher_query.py#L172

at which time the instrument is not set and this happens.

it is a bit tricky since callback calls are updates from the backend. it is quite possible to immitate them in the test, but it will take construction of this particular sequence of calls

volodymyrss:
I realzied it is more relevant that instrument is not set. I just changed the title

volodymyrss:
we really need to make here a test for the callback endpoint. if you want - you can do, that way you have an opportunity to understand it better yourself. if not - I can do, then you can understand with the example. let me know

burnout87:
> we really need to make here a test for the callback endpoint. if you want - you can do, that way you have an opportunity to understand it better yourself. if not - I can do, then you can understand with the example. let me know

Yes, otherwise we will not manage to close this PR. I will do it, since I anyway wanted to understand the flow when a call back is requested, this is the best opportunity

volodymyrss:
I changed the coverage computation too, so it's a bit complex. But the test shows the issue now.

volodymyrss:
@burnout87 does the test make sense?

.github/workflows/python-app.yml
```diff
@@ -46,7 +46,7 @@ jobs:
 
     - name: Test with pytest
       run: |
-        python -m coverage run -m pytest  tests/test_server_basic.py -sv --full-trace --log-cli-level=DEBUG -m 'not isgri_plugin and not dda'
+        python -m coverage run --source=cdci_data_analysis -m pytest tests -sv --full-trace --log-cli-level=DEBUG -m 'not isgri_plugin and not dda'
```
burnout87, line 49: Does this guarantee that for the coverage evaluation, only the sources within cdc_data_analysis are considered?
If so, then it's really good!

tests/test_job_management.py
```diff
@@ -27,12 +27,31 @@ def test_callback_without_prior_run_analysis(dispatcher_live_fixture):
...
+                        async_dispatcher=False,
+                    ))
+
+    print(c.text)    
+
+    assert c.status_code == 200        
+
+    c = requests.get(server + "/call_back",
+                   params={
+                       'job_id': 'test-job-id',
```
burnout87, line 42: Is this job-id completely randomly choosen or does it have a purpose?

volodymyrss: no purpose in this case. As we advance the job management testing, it will correspond to a directory which must exist.
Right now it just covers a case  which otherwise leads to an unhandled exception.

.github/workflows/python-app.yml
volodymyrss: Yes, else you can see the [report](https://codecov.io/gh/oda-hub/dispatcher-app/tree/ec557543c3415dbc6ebf79e2cbaf8e44a96f4754)  it includes other things, e.g. tests themselves.

burnout87 (approved):
Thanks for the effort! Looking forward to trying those new tests

volodymyrss:
thanks for reviewing!

PR #47, opened 2021-04-01
copied verbatim part proposed for removal by CF

volodymyrss:


burnout87:
I might be missing something, but isn't this PR a duplication of #48  ?

volodymyrss:
> I might be missing something, but isn't this PR a duplication of #48 ?

it is not!

volodymyrss:
> I might be missing something, but isn't this PR a duplication of #48 ?

you might be confused by the fast that this one here at first contained stray commit (it was reverted almost immediately, but you must have seen it). Sorry for that!
But they are not the same.
Also this is a draft, and has no reviewers assigned. Let's not think about it much now. Even if it was in fact duplicated, it would not be a problem.

Purpose of each PR is indicated in their titles, and they are clearly distinct.

Issue #20, opened 2021-03-16
send email upon completion of long requests


volodymyrss:
token contains email. If request took a while - and email can be useful. There could be other factors to control email delivery. TBD

burnout87:
Which email address should be used?

Is there already something like info-at-astro etc etc ?

volodymyrss:
the email needs to be sent to the user - i.e. the one found in the token.

volodymyrss:
but it's a good idea to send one to us too. We have an email, like astroooda-feedback, we should ask @motame , I forgot.
we can also send a backup one to team@odahub.io

burnout87:
Sorry, I didn't specify, which email should be used as a sender? That's why I gave the example wit info at ...

volodymyrss:
yes, well, we still should send it to info@astro smth. for backup. 
as for the from, you can as well use team@odahub.io . It will work now for reply, and it includes you. And we see if we find some alternative.

volodymyrss:
oh, "from", "reply-to" emails as well as SMTP details should be configurable. They depend on the deployment. For experiment, you can even use localhost. Or gmail. For pytest - maybe [this](https://pypi.org/project/pytest-localserver/)?

burnout87:
Yes, those should be configurable, depending on the deployment, like we did for the secret key

Will look into it, I also found [this](https://realpython.com/python-send-email/).

burnout87:
Something there was no time to discuss during the last meeting: how can we define a "long" request?
I thought of two possible scenarios:

- Execution time exceeds a certain threshold
- Asynchs requests

Any other situation to take into account?

ferrigno:
I think that every request should be considered as long, UNLESS products are ready and can be directly retrieved at the first call. I do not know exactly if this is equivalent to asynchronous request, you should check in the dispatcher, I believe.

volodymyrss:
> Something there was no time to discuss during the last meeting: how can we define a "long" request?
> I thought of two possible scenarios:
> 
>     * Execution time exceeds a certain threshold

Yes. Maybe we can store this user choice in token.
By the way, note that this will probably only happen during "callback" requests, i.e. when user is not active.

> 
>     * Asynchs requests

This should only apply to async request, since sync ones can never be long really - they will timeout.

> 
> 
> Any other situation to take into account?

volodymyrss:
[...]

PR #51, opened 2021-04-03
Integralallsky plugin

volodymyrss:


volodymyrss:
@burnout87 here I test compatibility with one of the plugins.

cdci_data_analysis/analysis/queries.py
```diff
@@ -545,6 +542,9 @@ def get_query_products(self,instrument,job,run_asynch,query_type='Real',logger=N
             else:
                 debug_message=''
 
+            if os.environ.get('DISPATCHER_DEBUG', 'yes') == 'yes':
```
volodymyrss, line 545: do we still need this? I think maybe yes?

burnout87: It is set here:
[https://github.com/oda-hub/dispatcher-app/blob/6330adeb99832ec23d05943e705badb87badd10a/tests/test_configurer.py#L5](https://github.com/oda-hub/dispatcher-app/blob/6330adeb99832ec23d05943e705badb87badd10a/tests/test_configurer.py#L5)

and used here:
[https://github.com/oda-hub/dispatcher-app/blob/6330adeb99832ec23d05943e705badb87badd10a/tests/test_configurer.py#L32](https://github.com/oda-hub/dispatcher-app/blob/6330adeb99832ec23d05943e705badb87badd10a/tests/test_configurer.py#L32)

cdci_data_analysis/analysis/queries.py
```diff
@@ -529,7 +524,9 @@ def get_query_products(self,instrument,job,run_asynch,query_type='Real',logger=N
             logger.error("passing request issue: %s", e)
             raise
 
-        except Exception as e: # noo!!!
+        except Exception as e: # TODO: could we avoid these? they make error tracking hard
+            logger.exception("failed to get query products")
```
burnout87, line 528: This is indeed a bit too generic is such an important stage of the whole processing. Perhaps the definition of a dedicated exception could help in tracking down the error

volodymyrss: indeed. This is also related to https://github.com/oda-hub/dispatcher-app/issues/25 . But I do not want to address the entire challenge in this PR.

cdci_data_analysis/analysis/queries.py
burnout87: So I would say that is is good to have it for testing purposes

volodymyrss: absolutely. My concern here was this change is present in this PR, I thought it is already there. But it's ok, it's fine

cdci_data_analysis/analysis/instrument.py
```diff
@@ -166,6 +166,9 @@ def run_query(self, product_type,
         if logger is None:
             logger = self.get_logger()
 
+        #  this was removed by 2f5b5dfb7e but turns out it is used by some plugins, see test_server_plugin_integral_all_sky
+        self._current_par_dic=par_dic
```
burnout87, line 170: Which case brakes?
Because as far as I  remember, _current_par_dic is no longer used anywhere

volodymyrss: it says it in the comment: this very case here test_server_plugin_integral_all_sky

volodymyrss: here: 

https://github.com/oda-hub/dispatcher-plugin-integral-all-sky/blob/master/dispatcher_plugin_integral_all_sky/spiacs_lightcurve_query.py#L311

maybe this breaks encapsulation, it needs to be determined if it is actually necessary to support this interface to plugins.

burnout87: It is used for the definition of a product, but honestly I don't know the details why:
[https://github.com/oda-hub/dispatcher-app/blob/6330adeb99832ec23d05943e705badb87badd10a/cdci_data_analysis/analysis/products.py#L327-L373](https://github.com/oda-hub/dispatcher-app/blob/6330adeb99832ec23d05943e705badb87badd10a/cdci_data_analysis/analysis/products.py#L327-L373)

volodymyrss: > It is used for the definition of a product, but honestly I don't know the details why:

oh it is used to add url to fits files. This url is a unique reference to the product, and it can be resolved to actually get the product from the frontend. 

It is essential that these urls are built and maintained. It is also the url I mentioned in [there](https://github.com/oda-hub/dispatcher-app/issues/20#issuecomment-818571887).

What is unclear, however, is why should this possibly be done customly in  this plugin, by accessing presumably internal (_.*) attribute. All plugins should do this equally, and hence it can be done in the dispatcher base class.

But as of now, looking in other plugins, the same approach is used everywhere: 
(e.g. [polar](https://github.com/oda-hub/dispatcher-plugin-polar/blob/master/cdci_polar_plugin/polar_lightcurve_query.py#L243))

But, before we do any refactoring, this functionality should be maintained.

This PR here also serves as an example of the kind of near-production cases we need before continious deployment can be done safely.

burnout87 (commented):
Jut one comment about the use of the new marker

volodymyrss:
> This marker should be included also here ?
[https://github.com/oda-hub/dispatcher-app/blob/6330adeb99832ec23d05943e705badb87badd10a/.github/workflows/python-app.yml#L47-L49](https://github.com/oda-hub/dispatcher-app/blob/6330adeb99832ec23d05943e705badb87badd10a/.github/workflows/python-app.yml#L47-L49) 

you mean, to exclude it from tests? This one is fast, it passes tests, let's let be. Maybe we could mark it as "network"

burnout87 (approved):

PR #56, opened 2021-04-14
Sending mail completion of a request

burnout87:
The first implementation currently does not support the evaluation of a long request.

A test has been provided which starts a local smtpd server that only acknowledges the ending of an email and then discards those, without actually forwarding it to the receiver.

smtp server settings can be configured, along with the sender email address to be used. These will vary on the deployment


burnout87:
- An email is sent in case of query concluded with the status 'done' or 'failed', at the moment, the email text is quite standard, just mentioning that the request was completed with the relative state
- Within the configurations the following parameters can be set:
  - smtp_server (eg localhost)
  - sender_mail (eg 'team@odahub.io')
  - cc_receivers_mail (eg ['team@odahub.io'] )
  - smtp_port (eg 1025)
  - mail_password
- The receiver email address is extracted from the user token
- From the user token also a threshold time duration that, if exceeded, will trigger the email sending (a default value is applied otherwise)
- Some tests is also there, for that purpose a simple local smtp server is started (this only acknowledge the fact that the email has been sent, this is not actually the forwarded)
- As already discussed, depending on the deployment, smtp server will vary

I am not entirely sure on how the _async_  case should be handled, I guess I will udnerstand soon

volodymyrss (changes requested):
Thanks! The added functionalities are useful.

However, several things were missing from our request which made the required use case logic unclear:

We really care about sending email when user is not interactively engaged with the platform. Right now, you'll send the email when a little bit longer interactive request, i.e. longer execution of `run_query`. In practice, this can never really be too long to justify email on `done`. Also interactive user can just see the error returned, no need to send an email. And if the request is interrupted, the connection will break and email will not be sent.

We really care about sending email on completion for very long requests, like > 30min, since these are the requests which are likely to be non-interactive.

I noted it briefly in [here](https://github.com/oda-hub/dispatcher-app/issues/20#issuecomment-818565648) but it was not sufficiently clear, sorry.

So, we really need to send an email on `callback` call, when job is done or failed.

In addition, we would want to send an email (additionally configurable) on `failed` and  `new` request of any kind. This will allow user to come back to the request. This particular function might end up as a lot of emails, so it's better be extra configurable, and we need to be a bit careful. But they are needed.

If you want, in  this PR we can close on basic email functionality, and deal with necessary async (callback, job handling)  features after.
Or we can continue here.

cdci_data_analysis/analysis/tokenHelper.py
```diff
@@ -12,6 +13,16 @@
     return decoded_token['name'] if 'name' in decoded_token else ''
 
 
+def get_token_user_mail(decoded_token):
+    # extract user
+    return decoded_token['sub'] if 'sub' in decoded_token else ''
+
+
+def get_token_user_threshold_mail(decoded_token):
+    # extract user
```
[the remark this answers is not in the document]
burnout87: Yes indeed

cdci_data_analysis/analysis/products.py
```diff
@@ -152,6 +152,21 @@
         self.status_dictionary['comment'] = str(comment)
         self.status_dictionary['warning'] = str(warning)
 
+    def set_status_warning(self, warning):
```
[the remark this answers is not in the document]
burnout87: I simply wanted to have the possibility to change a single entry, but it makes actually have those enclosed in a single one, 5 small ones is actually a bit redundant

cdci_data_analysis/analysis/tokenHelper.py
```diff
@@ -12,6 +13,16 @@
     return decoded_token['name'] if 'name' in decoded_token else ''
 
 
+def get_token_user_mail(decoded_token):
+    # extract user
+    return decoded_token['sub'] if 'sub' in decoded_token else ''
+
+
+def get_token_user_threshold_mail(decoded_token):
```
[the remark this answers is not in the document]
burnout87: the threshold_mail is a parameter included within the user token, it is used to set a time limit threshold that if exceeded will trigger the email sending, it is to cover the case of very long requests (eg > 30min)

burnout87:
> Thanks! The added functionalities are useful.
> 
> However, several things were missing from our request which made the required use case logic unclear:
> 
> We really care about sending email when user is not interactively engaged with the platform. Right now, you'll send the email when a little bit longer interactive request, i.e. longer execution of `run_query`. In practice, this can never really be too long to justify email on `done`. Also interactive user can just see the error returned, no need to send an email. And if the request is interrupted, the connection will break and email will not be sent.
> 
> We really care about sending email on completion for very long requests, like > 30min, since these are the requests which are likely to be non-interactive.
> 
> I noted it briefly in [here](https://github.com/oda-hub/dispatcher-app/issues/20#issuecomment-818565648) but it was not sufficiently clear, sorry.
> 
> So, we really need to send an email on `callback` call, when job is done or failed.
> 
> In addition, we would want to send an email (additionally configurable) on `failed` and `new` request of any kind. This will allow user to come back to the request. This particular function might end up as a lot of emails, so it's better be extra configurable, and we need to be a bit careful. But they are needed.
> 
> If you want, in this PR we can close on basic email functionality, and deal with necessary async (callback, job handling) features after.
> Or we can continue here.

Thanks for the comments, I am now focusing on the async side of this feature. I realize it is taking me a bit more than I was expected, and I will have more questions since it is a part of the dispatcher I haven't used much yet.

Wat kind of extra-configuration would you see as ecessary?

volodymyrss:
> 
> Wat kind of extra-configuration would you see as ecessary?

I would say two kind of email preferences:

* if email is sent on done (if timeout is satisfied, which can be 0)
* if email is sent when new job is created. Actually this happens when backend responds "submitted" which is returned to the client.

burnout87:
I was also forgetting if you are curious about the problem I was encountering with the subprocess creation, I put here the code I was trying to run but didn't succeed:

[https://github.com/oda-hub/dispatcher-app/blob/3c024cad60122be2d77acafd5c3ae5d6805bdfc0/tests/conftest.py#L58-L107](https://github.com/oda-hub/dispatcher-app/blob/3c024cad60122be2d77acafd5c3ae5d6805bdfc0/tests/conftest.py#L58-L107)

cdci_data_analysis/flask_app/app.py
```diff
@@ -201,7 +201,9 @@ def common_exception_payload():
 
     if payload['debug_mode'] == "yes":
         payload['config'] = {
-            'dispatcher-config': remove_nested_keys(app.config['conf'].as_dict(),  ['sentry_url', 'logstash_host', 'logstash_port', 'secret_key'])
+            'dispatcher-config': remove_nested_keys(app.config['conf'].as_dict(),
+                                                    ['sentry_url', 'logstash_host', 'logstash_port', 'secret_key',
+                                                     'mail_password',])
```
volodymyrss, line 206: mail_password is SMTP server password, yes? If yes, it's good to be specific - there are many other possible mail-related passwords. Also not all SMTP servers would need password, e.g. those running on localhost (and also as that used currently to send feedback) usually do not.

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -502,8 +506,12 @@
         query_out = None
 
         _, self.config_data_server = self.set_config()
-        
-        job = job_factory(self.par_dic['instrument_name'],
+        instrument_name = time_request = ""
+        if 'instrument_name' in self.par_dic:
```
volodymyrss, line 510: you can use things like:

```python
instrument_name = self.par_dic.get('instrument_name', '')
```

which is reasonably readable and a lot shorter. 
Or you intentionally avoided it?

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -526,11 +534,42 @@
...
         logger.warn('-----> set status to %s', status)
 
-        job.write_dataserver_status(
-            status_dictionary_value=status, full_dict=self.par_dic)
+        if self.is_email_to_send_callback(status):
+            try:
+                self.send_email(status,
+                                instrument=instrument_name,
+                                time_request=time_request)
+                self.par_dic['mail_status'] = 'mail sent'
```
volodymyrss, line 542: you use `par_dict` to store outcome of the request processing? will it not cause issues in some places, e.g. since `par_dict` is used to build `job_id`? In general `par_dict` was meant to be more or less the *input parameter dict*.

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -526,11 +534,42 @@
...
+                self.logger.warning("query dismissed by token validation")
+                return resp
+            timeout_threshold_mail = tokenHelper.get_token_user_timeout_threshold_mail(self.decoded_token)
+            if timeout_threshold_mail is None:
+                # set it to the a default value, from the configuration
+                timeout_threshold_mail = self.app.config.get('conf').mail_sending_timeout_threshold
+            # in case the request was long and 'done'
+            # or if failed
+            # or when the job was created ('submitted')
+            return (duration_query > timeout_threshold_mail and status == 'done') or status == 'submitted' or status == 'failed'
```
volodymyrss, line 570: I am not sure `submitted` ever happens during callback, except perhaps due to an issue. Generally it will only happen during `run_analysis` request. So email might be sent in both calls.

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -1255,9 +1285,79 @@
...
                                               off_line=off_line,
                                               api=api)
-
         return resp
 
+    def send_email(self, status="done",
+                   instrument="",
+                   time_request=""):
+        server = None
+        self.logger.info("Sending completion mail")
```
volodymyrss, line 1294: completeion or submission?

volodymyrss (changes requested):

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: Thanks I will use this approach, much better indeed

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: I thought about it, an alternative could be not to store anything, and just assume that the email was sent correctly. In case of problems, then additional information should be stored.

cdci_data_analysis/flask_app/app.py
burnout87: Thanks, will update!

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: What about storing this information inside the monitor of the job? In this way, the input list of parameter won't be affected, and the status of the email still logged

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: I will include also the case `submitted` for `run_analysis`

burnout87: For my understanding, we want to have an email when the query status becomes `submitted`? This happens here:
 [https://github.com/oda-hub/dispatcher-app/blob/b9191b653203859745117d008cc41de524d2d044/cdci_data_analysis/flask_app/dispatcher_query.py#L1195-L1205](https://github.com/oda-hub/dispatcher-app/blob/b9191b653203859745117d008cc41de524d2d044/cdci_data_analysis/flask_app/dispatcher_query.py#L1195-L1205)

So, afterwards, the user will be receiving updates by a number of `call_back` call?

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -526,11 +542,47 @@
...
-        job.write_dataserver_status(
-            status_dictionary_value=status, full_dict=self.par_dic)
+        if self.is_email_to_send_callback(status):
+            try:
+                self.send_email(status,
+                                instrument=instrument_name,
+                                time_request=time_request)
+                job.write_dataserver_status(status_dictionary_value=status,
+                                            full_dict=self.par_dic,
+                                            email_status='email sent')
```
volodymyrss, line 552: nice, this is good approach!

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -801,9 +853,11 @@
 
         if self.token is not None:
             try:
-                validate = self.validate_query_from_token()
-                if validate:
-                    pass
+                if self.validate_query_from_token():
```
volodymyrss, line 856: hm I missed befor, but still `(self, )` above here

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -1174,6 +1216,19 @@
                     else:
                         query_new_status = 'submitted'
                         job.set_submitted()
+                        mail_sending_job_submitted = self.app.config.get('conf').mail_sending_job_submitted
```
volodymyrss, line 1219: I think this might be in the token too. Users may want to configure this possibility

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -1174,6 +1216,19 @@
...
+                        # send submitted email
+                        if mail_sending_job_submitted:
+                            try:
+                                time_request = self.par_dic.get('time_request', '')
+                                self.send_email('submitted',
+                                                instrument=self.instrument.name,
+                                                time_request=time_request)
+                                # store an additional information about the sent email
+                                query_out.set_status_field('email_status', 'email sent')
+                            except MailNotSent as e:
```
volodymyrss, line 1229: we should be sure the problem is sent to sentry. we have to be certain to react on this

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -1255,9 +1311,79 @@
                                               job_monitor=job_monitor,
                                               off_line=off_line,
                                               api=api)
-
         return resp
 
+    def send_email(self, status="done",
+                   instrument="",
+                   time_request="",
+                   products_url=""):
```
volodymyrss, line 1319: it will be provided later, yes?

burnout87: yes...on the way!

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: I will need some clarification on this

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: Will be included also in the token, along with all the email sending related parameters

tests/test-conf.yaml
```diff
@@ -4,7 +4,7 @@ dispatcher:
     dispatcher_url: 0.0.0.0
     dispatcher_port: 8001
     dispatcher_service_url: http://localhost:8001
-    sentry_url: "https://2ba7e5918358439485632251fa73658c@sentry.io/1467382"
+    sentry_url: "https://2703d13672264153a19e86f2cf5a415b@sentry.io/5729456"
```
volodymyrss: hm, did you accidentially store your real URL?

burnout87: yes, reverting now

volodymyrss: yes, but making a commit replacing it does not actually remove the token from the git history.

You will be always able to access it by specific commit, since history is preserved:

https://github.com/oda-hub/dispatcher-app/blob/117c0d82184cc6ab4ac86e2a96d7c9daa0773c8f/tests/test-conf.yaml

There are special tools to purge something from the entire tree, but it will require rewriting history, and a force-push, which is unfortunate.
Also, the token has already been exposed in the public, and there are robots specifically searching for these on github.
I think the safest and easiest way is to expire this particular token in the sentry interface, and create a new one.

tests/conftest.py
```diff
@@ -25,6 +25,88 @@ def app():
...
+    controller = Controller(handler, hostname='127.0.0.1', port=1025)
+    # Run the event loop in a separate thread
+    controller.start()
+
+    yield controller
+
+    print("will stop the mail server")
+    controller.stop()
+
+    # # with subprocess
```
volodymyrss, line 58: the asyncio version looks good to me, but you want to try also the subprocess one? maybe it is interesting, if so could you make another fixture out of it, with the same interface as the existing one? So we test it right away.

in principle subprocess was vital for server, since it is good to see the exact command used for running it.

tests/test-conf.yaml
```diff
@@ -8,3 +8,11 @@ dispatcher:
     logstash_host: 10.194.169.75
     logstash_port: 5001
     secret_key: 'secretkey_test'
+    smtp_server: 'localhost'
+    sender_mail: 'team@odahub.io'
+    cc_receivers_mail: ['team@odahub.io']
+    smtp_port: 1025
+    smtp_server_password: ''
+    mail_sending_timeout: True
+    mail_sending_timeout_threshold: 1800
```
volodymyrss, line 17: I would call it `mail_sending_timeout_default_threshold` for clarity. It's how it is used, yes?

Also, small suggestion, why not make another dict layer for mail, for example? There are quite many keys. It will be more readable.

volodymyrss: by the way, there is example config there in conf_dir, it would need to be updated, although it does not affect immediately any functionality:

https://github.com/oda-hub/dispatcher-app/blob/master/cdci_data_analysis/config_dir/conf_env.yml.example

tests/test_job_management.py
```diff
@@ -1,95 +1,253 @@
...
 
     print("response from run_analysis:", json.dumps(c.json(), indent=4))
 
     session_id = c.json()['session_id']
     job_id = c.json()['job_monitor']['job_id']
 
-    #TODO ensure it is submitted
-
     job_monitor_json_fn = f'scratch_sid_{session_id}_jid_{job_id}/job_monitor.json'
+    # the aliased version might have been created
```
volodymyrss, line 70: when I tested I assumed the area is empty, and alias is not created. It is true that it's better to ensure that, or workaround like you did.
I wonder if it is better to enforce this property, e.g. by setting some random parameter.
That way both path: aliased and not, would be tested clearly separately.
But I would not want this request to request extra work. Maybe lets add a TODO? What do you think?

tests/test_job_management.py
```diff
@@ -1,95 +1,253 @@
...
+        job_monitor_call_back_failed_json_fn_aliased)
+    assert c.status_code == 200
+    # read the json file
+    if os.path.exists(job_monitor_call_back_failed_json_fn):
+        f = open(job_monitor_call_back_failed_json_fn)
+    else:
+        f = open(job_monitor_call_back_failed_json_fn_aliased)
+
+    jdata = json.load(f)
+    assert jdata['email_status'] == 'email sent'
```
volodymyrss, line 160: could we make the smtp server fixture write the email in a file, read it here and verify it is as expected?

tests/conftest.py
```diff
@@ -25,6 +25,88 @@ def app():
...
+        async def handle_DATA(self, server, session, envelope):
+            try:
+                peer = session.peer
+                mail_from = envelope.mail_from
+                rcpt_tos = envelope.rcpt_tos
+                data = envelope.content
+                print(f"mail server: Receiving message from: {peer}")
+                print(f"mail server: Message addressed from: {mail_from}")
+                print(f"mail server: Message addressed to: {rcpt_tos}")
+                print(f"mail server: Message length : {len(data)}")
```
volodymyrss, line 42: here I would write to file and verify the content later in the test. Or maybe interpret from the output.

cdci_data_analysis/analysis/products.py
```diff
@@ -205,9 +208,10 @@
 
         if message is None:
             message = '%s' % message_prepend_str
-            message += 'failed: %s' % (failed_operation)
+            message += ' failed: %s' % (failed_operation)
             if extra_message is not None:
                 message += 'message: %s' % (extra_message)
+            message.strip()
```
volodymyrss, line 214: this does nothing, I guess you wanted message = message.strip()

volodymyrss (changes requested):

tests/conftest.py
burnout87: You can check now, a test with the subprocess is there, and the error is indeed generated

tests/test-conf.yaml
burnout87: good idea, configuration options are growing

tests/test_job_management.py
burnout87: One question: the reason why the aliased folder is created is only because there was already a folder created with the same session_id and job_id?

volodymyrss: only the same job_id is required. The purpose is that if there is already a running job, the user can re-attach to it. That URL we insert in the email will lead the user to the same request.
But session ID may change, if new browser window is started.
By the way, we should prevent aliasing between different authorized users:

https://github.com/oda-hub/dispatcher-app/issues/59

burnout87: Ok, I will add as a TODO, it is not strictly related to this PR

tests/test-conf.yaml
burnout87: I will just re-create the project...

volodymyrss: It's a known issue, many people end up doing it at some point. 
Recreating project does it, but sentry specifically supports expiring tokens for this or similar reasons.

tests/conftest.py
burnout87: @volodymyrss did you man age to reproduce the same error ?

tests/test_job_management.py
burnout87: It is an option. But it would work only within the specific context of this test ie using a localhost server. In a real context, we don't have the option to inspect log files generated from an SMTP server. 

An alternative would be to check that the email_status was written in the json output file...

volodymyrss: I mean precisely for the test, that's why I commented on the test code, sorry if I was unclear.

checking email_status would be also useful, but verifying content of the email as delivered to the smtp server, and stored by the test server, is useful, since email is delivered to the user, and needs to be well checked.

tests/conftest.py
burnout87: From the context of this fixture, there is no knowledge about the session_id nor the job_id, so the scratch folder for the specific test cannot be located (maybe some information can be extracted by inspecting the content of the mail, though this wouldn't be complete). 

Any idea?

Another approach would be to just write over another file inside the test folder, and perhaps move this inside the scratch folder at the end of the test...not very elegant, but it would work...what do you think?

volodymyrss: > Another approach would be to just write over another file inside the test folder, and perhaps move this inside the scratch folder at the end of the test...not very elegant, but it would work...what do you think?

This file is meant to represent the events/state of the SMTP  server. Test then verifies that the SMTP server got what it should.
So there is not reason why state of the SMTP server should be in scratch directory of a particular job. 
In fact, it would not be elegant if it was.
To avoid relying on a convention on the stored filename between conftest and tests,  I'd make the fixture report it's content as an attribute.
See how caplog fixture does it.

volodymyrss (approved):

burnout87:
Thanks

volodymyrss:
thanks to you!

PR #62, opened 2021-04-23
Prevented aliasing different users

burnout87:
Will generate the job_idtaking into account also the user_id, so that two users generating the very same request yield differen scratch folders.

Wokring on some tests now.

### DIFF
--- a/cdci_data_analysis/flask_app/dispatcher_query.py
+++ b/cdci_data_analysis/flask_app/dispatcher_query.py
@@ -578,10 +578,13 @@ class InstrumentQueryBackEnd:
         job_monitor_status_json_file_aliased = f'scratch_sid_{session_id}_jid_{job_id}_aliased/query_output.json'
         request_url = ""
         file = None
-        if os.path.exists(job_monitor_status_json_file):
-            file = open(job_monitor_status_json_file)
-        elif os.path.exists(job_monitor_status_json_file_aliased):
-            file = open(job_monitor_status_json_file_aliased)
+        if self.scratch_dir:
+            file = open(self.scratch_dir + '/query_output.json')
+        else:
+            if os.path.exists(job_monitor_status_json_file_aliased):
+                file = open(job_monitor_status_json_file_aliased)
+            elif os.path.exists(job_monitor_status_json_file):
+                file = open(job_monitor_status_json_file)
         if file:
             jdata = json.load(file)
             if 'prod_dictionary' in jdata and 'analysis_parameters' in jdata['prod_dictionary']:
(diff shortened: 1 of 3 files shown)


volodymyrss (changes requested):
I think it needs to be rebased or merged with #56

burnout87:
Yes I think the same. So I merge with #56  and then compared with the master? These should be the correct steps ?

volodymyrss:
there are several ways, rebase is possible, but you could do the same you did for #56 when #60 was merged before. You can merge master on this branch.

image
https://github.com/oda-hub/dispatcher-app/network

burnout87:
> there are several ways, rebase is possible, but you could do the same you did for #56 when #60 was merged before. You can merge master on this branch.
> 
> image
> https://github.com/oda-hub/dispatcher-app/network


In any case, before I can do anything I think you need to approve...

volodymyrss:
> In any case, before I can do anything I think you need to approve...

I will be a lot easier for me to compare when it will be mergable. This merge on feature branch would be a modification to  prevented-aliasing-different-users branch, a change to it, which is why I set it to `request changes` - the requested change is to make it compatible with master.
Please let me know if it does not make sense, we can discuss and call too.

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -565,9 +574,15 @@
 
     def generate_request_url_call_back(self, products_url, session_id, job_id) -> str:
         job_monitor_status_json_file = f'scratch_sid_{session_id}_jid_{job_id}/query_output.json'
+        # to be handled now, with the job_id generated taking into account only the user_id
+        job_monitor_status_json_file_aliased = f'scratch_sid_{session_id}_jid_{job_id}_aliased/query_output.json'
         request_url = ""
+        file = None
         if os.path.exists(job_monitor_status_json_file):
             file = open(job_monitor_status_json_file)
+        elif os.path.exists(job_monitor_status_json_file_aliased):
```
volodymyrss, line 583: hm, I think maybe aliased should take the priority if it exists? Since it is the one that corresponds to the currently processed request.
But then if this query_output is written in non-aliased directory at the moment, it should be written there.

tests/test_server_basic.py
```diff
@@ -87,7 +87,76 @@ def test_empty_request(dispatcher_live_fixture):
     logger.info(jdata['config'])
 
 
-def test_valid_token(dispatcher_live_fixture,):
+def test_same_request_different_users(dispatcher_live_fixture):
```
volodymyrss, line 90: nice!

volodymyrss (changes requested):
small change, maybe even not necessary for this case, but since you addressed the part of query_output.json, and it is related to the present PR.

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: Good point.

However, I just realized that the path of the scratch directory in use is actually saved as a member of the InstrumentQueryBacvkEnd object..I guess I can simply use that one...will test neo

volodymyrss (approved):

PR #63, opened 2021-04-26
new example options

volodymyrss:
example config was not updated for the most recently renamed field


burnout87 (approved):
For some reason, I was sure I did rename that option

PR #69, opened 2021-04-27
Reusable fixture, for reuse in plugins and api

volodymyrss:


cdci_data_analysis/pytest_fixtures.py
```diff
@@ -0,0 +1,251 @@
...
+        smtp_server: 'localhost'
+        sender_mail: 'team@odahub.io'
+        cc_receivers_mail: ['team@odahub.io']
+        smtp_port: 1025
+        smtp_server_password: ''
+        email_sending_timeout: True
+        email_sending_timeout_default_threshold: 1800
+        email_sending_job_submitted: True
+    """)
+
```
burnout87, line 164: Like this, don't you end up writing within the 'test-dispatcher-conf.yaml' file the same content?

burnout87 (commented):
Will this approach completely replace the 'tests/test-conf.yaml' file?

volodymyrss:
> Will this approach completely replace the 'tests/test-conf.yaml' file?

yes

cdci_data_analysis/pytest_fixtures.py
volodymyrss: yes. but it's easier to do like this that to distribute and then discover yet another non-code file. see how discovery of the test*yaml was rather cumbersome

burnout87: Nice, it will just have to be aligned with the latest version of the configuration

burnout87 (approved):

PR #75, opened 2021-04-29
'None' (string, not python None) appears in callback for public requests, and is causing failures

volodymyrss:
I attempted to make a test, but it's not complete.

Could you please have a look?
I think the problem is more or less clear, `token='None'` and `time_request='None'` appear in callbacks for public request, and cause trouble in the constructor (ValueError on converting time to float, jwt.exceptions.DecodeError for token).

### DIFF
--- a/tests/test_job_management.py
+++ b/tests/test_job_management.py
@@ -155,8 +155,9 @@ def test_email_callback_after_run_analysis(dispatcher_live_fixture, dispatcher_l
 
 
 @pytest.mark.parametrize("default_values", [True, False])
-def test_email_callback_after_run_analysis(dispatcher_live_fixture, dispatcher_local_mail_server, default_values):
->>>>>>> d33b3be25e99e64dd854a2e7e57a6198f3079f71
+@pytest.mark.parametrize("time_request_none", [True, False])
+def test_email_callback_after_run_analysis(dispatcher_live_fixture, dispatcher_local_mail_server, default_values, time_request_none):
+# >>>>>>> d33b3be25e99e64dd854a2e7e57a6198f3079f71
     # TODO: for now, this is not very different from no-prior-run_analysis. This will improve
 
     server = dispatcher_live_fixture
@@ -300,8 +301,9 @@ def test_email_callback_after_run_analysis(dispatcher_live_fixture, dispatcher_l
         f = open(job_monitor_call_back_done_json_fn_aliased)
 
     jdata = json.load(f)
-    if default_values:
-        # email not supposed to be sent (request is short)
+    if default_values or time_request_none:
+        # email not supposed to be sent because request is too short
+        # or the time of the original request was not sent
         assert 'email_status' not in jdata
     else:
         assert jdata['email_status'] == 'email sent'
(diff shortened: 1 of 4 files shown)


tests/test_job_management.py
```diff
@@ -90,6 +90,73 @@ def test_public_async_request(dispatcher_live_fixture, dispatcher_local_mail_ser
...
+        f = open(job_monitor_call_back_done_json_fn_aliased)
+
+    jdata = json.load(f)
+    assert 'email_status' not in jdata
+    assert 'call_back_status' in jdata
+    assert jdata['call_back_status'] == 'parameter missing during call back: original request time not available'
+
+
+@pytest.mark.parametrize("default_values", [True, False])
+def test_email_callback_after_run_analysis(dispatcher_live_fixture, dispatcher_local_mail_server, default_values):
```
burnout87, line 158: Is the same test repeated twice? Or perhaps there's some copy and paste error?

burnout87 (changes requested):

tests/test_job_management.py
```diff
@@ -90,6 +90,73 @@ def test_public_async_request(dispatcher_live_fixture, dispatcher_local_mail_ser
 @pytest.mark.parametrize("default_values", [(True), False])
 @pytest.mark.parametrize("time_request_none", [True, False])
 def test_email_callback_after_run_analysis(dispatcher_live_fixture, dispatcher_local_mail_server, default_values, time_request_none):
+#def test_time_request(dispatcher_live_fixture, dispatcher_local_mail_server): #VS: sorry, was not sure what to do with name?
```
burnout87, line 93: I had made to separate test to test the two parameters separately. But better in this way, more testing has been actually performed. I change also the way the time of the request was initialized

burnout87 (changes requested):

burnout87:
Marginal change in the way the email is formatted in case the time of the request is not available

tests/test_job_management.py
volodymyrss: it's some merge issue. anyway, I did not mean to complete the fix, I could have made  an issue. Sorry for being unclear. But now it is done I guess

volodymyrss:
hm, test failed. but it worked locally. I can not work on it tomorrow. Could you finish it up when possible? Thanks!

burnout87:
> hm, test failed. but it worked locally. I can not work on it tomorrow. Could you finish it up when possible? Thanks!

Yes I will work on it

tests/test_job_management.py
```diff
@@ -124,6 +132,7 @@ def test_email_callback_after_run_analysis(dispatcher_live_fixture, dispatcher_l
         # makes more sense to have it sent directly with the request,
         # though it is not considered for the url encoding
         # to confirm
+        # VS: I do not really understand, why is it useful in the request? can dispatcher not compute it? Relying on time from another locations may cause issues
```
burnout87, line 134: Do you mean to have it acquired on the dispatcher's side?

It indeed makes sense, it will help also in avoinding issues related to requests coming from different time zones?

volodymyrss: exactly! timezones, time sync, client errors and manipulations, etc

volodymyrss:
> > hm, test failed. but it worked locally. I can not work on it tomorrow. Could you finish it up when possible? Thanks!
> 
> Yes I will work on it

thanks! I might have made some problem with merging. But the principle of solution works.

burnout87:
> > > hm, test failed. but it worked locally. I can not work on it tomorrow. Could you finish it up when possible? Thanks!
> > 
> > 
> > Yes I will work on it
> 
> thanks! I might have made some problem with merging. But the principle of solution works.

It should be ok now. I can merge and deal with #76

PR #76, opened 2021-04-30
Time request in dispatcher

burnout87:
@volodymyrss  With this, I will address the time-request setting within the dispatcher


volodymyrss (approved):
Thanks!
It might be more issues will come up when more tests are added to plugins.

PR #84, opened 2021-05-04
clean up of some tests, some questions

burnout87:
did a first clean-up of some tests, that are now moved into the dispatcher-plugin related projects

### DIFF
--- a/tests/test_plugins.py
+++ b/tests/test_plugins.py
@@ -1,3 +1,4 @@
+### Are these needed? Some of those seem already present within other test modules
 
 from builtins import (str, open, range)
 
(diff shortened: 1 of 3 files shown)


tests/test_plugins.py
```diff
@@ -1,3 +1,4 @@
+### Are these needed? Some of those seem already present within other test modules
```
volodymyrss, line 1: that was long ago, kind of futures (https://python-future.org/imports.html)

No more necessary since we do not support python2 anymore.

volodymyrss (approved):

burnout87:
@volodymyrss Removed also the 'test_server_plugin_integral_all_sky' tests

volodymyrss (approved):
it's interesting that there is less coverage now. It means that remaining plugin-independent tests did not entirely cover what plugins use.

Previously, the purpose of the plugins-based tests was to show features plugin use before more generic tests defining dispatcher interface are adopted. Hence the plugin tests were fixing interface by plugin use - which is a bit vague but useful.

But the drop is not large and let's not focus on it.

PR #85, opened 2021-05-05
some updates on common test request

volodymyrss:
small updates to looping functionlity; should be later moved to another file for better importing


burnout87 (approved):
Thanks, actually I had already started to do exactly this. But you have been way faster

volodymyrss:
> Thanks, actually I had already started to do exactly this. But you have been way faster

Yeah sorry I needed to do my LC test which is needed for some eventual test case, would not want it remain broken..

volodymyrss:
thanks!

burnout87:
> > Thanks, actually I had already started to do exactly this. But you have been way faster
> 
> Yeah sorry I needed to do my LC test which is needed for some eventual test case, would not want it remain broken..

no problem! I had in mind exactly the same approach

PR #87, opened 2021-05-06
dict for factories

burnout87:
Instead of the list of factories, a dictionary pairing the name of the instrument and the relative factory is provided.

set_instrument calls the factory only for the needed instrument

### DIFF
--- a/cdci_data_analysis/plugins/importer.py
+++ b/cdci_data_analysis/plugins/importer.py
@@ -55,16 +55,20 @@ cdci_plugins_dict = {
 }
 
 instrument_factory_list = []
+instrument_name_dict = {}
 # pre-load the empty instrument factory
 
 instrument_factory_list.append(empty_instrument.my_instr_factory)
 instrument_factory_list.append(empty_async_instrument.my_instr_factory)
+instrument_name_dict['empty'] = empty_instrument.my_instr_factory
+instrument_name_dict['empty-async'] = empty_async_instrument.my_instr_factory
 
 for plugin_name in cdci_plugins_dict:
     logger.info("found plugin: %s", plugin_name)
 
     try:
         e = importlib.import_module(plugin_name+'.exposer')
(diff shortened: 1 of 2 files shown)


cdci_data_analysis/plugins/importer.py
```diff
@@ -55,16 +55,20 @@
...
 instrument_factory_list.append(empty_async_instrument.my_instr_factory)
+instrument_name_dict['empty'] = empty_instrument.my_instr_factory
+instrument_name_dict['empty-async'] = empty_async_instrument.my_instr_factory
 
 for plugin_name in cdci_plugins_dict:
     logger.info("found plugin: %s", plugin_name)
 
     try:
         e = importlib.import_module(plugin_name+'.exposer')
+        instrument_name_dict.update(e.instr_name_dict)
```
volodymyrss, line 71: this changes dispatcher-plugin interface, breaking it for existing plugins. There should be a reasonable default, compatible with what was before, for plugins which do not implement this.

burnout87: indeed, I need also a small change in the various plugins

burnout87: if we don't want to change the plugins interface, then this is not a solution

volodymyrss:
> if we don't want to change the plugins interface, then this is not a solution 

most plugins can not possible break on instrument constructor.
we would only need to change one plugin which, at this time, might break.
and preserve other plugin interface.
but maybe it's too complex.

burnout87:
> can not possible break

ok then let's leave this solution aside for the moment, maybe it will be useful sometime in the future

volodymyrss:
> > can not possible break
> 
> ok then let's leave this solution aside for the moment, maybe it will be useful sometime in the future

ok, sure. One solution would be to fallback to previous mechanism if dict is not available. But it's not critical anyway. It's not even in the current milestone.

PR #89, opened 2021-05-07
failing test

burnout87:
failing test for an invalid token

scratch_dir is not created, and request not processed as public

### DIFF
--- a/tests/test_server_basic.py
+++ b/tests/test_server_basic.py
@@ -189,6 +189,45 @@ def test_valid_token(dispatcher_live_fixture):
     logger.info(json.dumps(jdata, indent=4))
 
 
+@pytest.mark.xfail
+def test_invalid_token_scartch_dir_creation(dispatcher_live_fixture, ):
+    server = dispatcher_live_fixture
+
+    logger.info("constructed server: %s", server)
+    # let's generate an expired token
+    exp_time = int(time.time()) - 500
+    # expired token
+    token_payload = {
+        **default_token_payload,
+        "exp": exp_time
+    }
+    encoded_token = jwt.encode(token_payload, secret_key, algorithm='HS256')
+
+    params = {
+        **default_params,
+        'product_type': 'dummy',
+        'query_type': "Dummy",
+        'instrument': 'empty',
+        'token': encoded_token
+    }
+
+    jdata = ask(server,
+                params,
+                max_time_s=50,
+                expected_query_status=None,
+                expected_status_code=403
+                )
+
+    assert jdata['error_message'] == 'token expired'
+    logger.info("Json output content")
+    logger.info(json.dumps(jdata, indent=4))
+
+    # certain output information are not even returned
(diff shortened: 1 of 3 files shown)


volodymyrss:
oh you mean when token is invalid? Sorry I misunderstood as "no valid token is provided" == public request which does work.

If token is invalid, the request should **not** be treated as public, and should be returned immediately.
It is important so that users do not miss message and accidentally start processing under unexpected permissions.
Actually, maybe let's revert this test so that the directory is not actually created. But it will be similar to existing tests.
Also, we could add in a message something like "you can still make a public request, please remove the token to do so!"

burnout87:
> oh you mean when token is invalid? Sorry I misunderstood as "no valid token is provided" == public request which does work.

Ok, yes, sorry for the confusion. 

> 
> If token is invalid, the request should **not** be treated as public, and should be returned immediately.
> It is important so that users do not miss message and accidentally start processing under unexpected permissions.

And that's what is happening.

> Actually, maybe let's revert this test so that the directory is not actually created. But it will be similar to existing tests.
> Also, we could add in a message something like "you can still make a public request, please remove the token to do so!"

burnout87:
> Also, we could add in a message something like "you can still make a public request, please remove the token to do so!"

Currently, the returned error message just mentions that the token is invalid/expired. But for the user it is very useful

tests/test_server_basic.py
```diff
@@ -189,6 +189,45 @@ def test_valid_token(dispatcher_live_fixture):
...
+
+    assert jdata['error_message'] == 'token expired'
+    logger.info("Json output content")
+    logger.info(json.dumps(jdata, indent=4))
+
+    # certain output information are not even returned
+    assert 'session_id' in jdata
+    assert 'job_monitor' in jdata
+    assert 'job_id' in jdata['job_monitor']
+
```
volodymyrss, line 229: so but can we check that scratch is not created? To capture the behavior

burnout87: A bit tricky to capture. 

One option I see, is to count the number of the scratch folders at the beginning of the test, and then at the end. This number does not have to change.

What do you think of that?

volodymyrss: That's a possibility.

burnout87: Within the given workflow, will it ever happen to have multiple tests running in parallel?

volodymyrss: Doing that might cause issues right now anyway.

By the way, to avoid that, and in general, it would be nice to create all these scratch directories in a temporary directory. 
It can be done in the fixture, and hence apply to all tests. https://github.com/oda-hub/dispatcher-app/issues/90
It might break tests which find files I guess. So the fixture needs to return it's location too.
So maybe I suggest not to do this now, and assume tests are not parallel. Just add a pytest.mark that it is not safe for parallel.

volodymyrss (approved):

PR #91, opened 2021-05-07
roles can be specified as list

burnout87:
previsously only one case was covered

### DIFF
--- a/cdci_data_analysis/analysis/tokenHelper.py
+++ b/cdci_data_analysis/analysis/tokenHelper.py
@@ -3,8 +3,12 @@ import jwt
 
 def get_token_roles(decoded_token):
     # extract role(s)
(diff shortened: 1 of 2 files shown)


volodymyrss (commented):
can they be? why? when?

burnout87:
When I wrote the first tests for the roles, I actually specified as strings.

[https://github.com/oda-hub/dispatcher-app/blob/349fba7f8b8803c97f4403d37487e21be5ce6fc1/tests/test_server_basic.py#L238-L240](https://github.com/oda-hub/dispatcher-app/blob/349fba7f8b8803c97f4403d37487e21be5ce6fc1/tests/test_server_basic.py#L238-L240)

I am not sure if it's that necessary to support also this format, but for sure the list must be.

volodymyrss (approved):

PR #93, opened 2021-05-07
test valid token oda_api

burnout87:


volodymyrss (approved):

volodymyrss (approved):

burnout87:
> > So, what was missing was the last version of the oda_api?
> 
> Apparently.

So, shouldn't we also change the requirements?

[https://github.com/oda-hub/dispatcher-app/blob/cb8e85c49aa5fc8e7f9ca0c027c376c7e9b71fb4/requirements.txt#L15-L17](https://github.com/oda-hub/dispatcher-app/blob/cb8e85c49aa5fc8e7f9ca0c027c376c7e9b71fb4/requirements.txt#L15-L17)

volodymyrss:
> > > So, what was missing was the last version of the oda_api?
> > 
> > 
> > Apparently.
> 
> So, shouldn't we also change the requirements?
> 
> https://github.com/oda-hub/dispatcher-app/blob/cb8e85c49aa5fc8e7f9ca0c027c376c7e9b71fb4/requirements.txt#L15-L17

requirements one is not really used since it is installed after from master.
but I added master one now.

volodymyrss:
looks like our changes were incompatible. Should I let you do it?
We can do it like here: https://github.com/oda-hub/dispatcher-plugin-antares

burnout87:
> looks like our changes were incompatible. Should I let you do it?
> We can do it like here: https://github.com/oda-hub/dispatcher-plugin-antares

yes I 'll do it

setup.py
```diff
@@ -16,9 +16,11 @@
 
 f = open("./requirements.txt",'r')
 install_req=f.readlines()
```
volodymyrss, line 18: we discussed this here: https://github.com/oda-hub/dispatcher-plugin-antares/pull/8 and highlighted that `install_requires` which is [meant to serve a bit different purpose than requirements.txt](https://packaging.python.org/discussions/install-requires-vs-requirements/).

setup.py
```diff
@@ -16,9 +16,11 @@
...
-# install_req = [req.replace('\n', '').strip() for req in install_req]
 f.close()
 
+
+install_req += ['oda_api', 'dispatcher-plugin-integral-all-sky']
```
volodymyrss, line 22: if anything, `install_req` is meant to be more narrow than `requirements.txt`

see [there](https://packaging.python.org/discussions/install-requires-vs-requirements/)
```
Whereas install_requires requirements are minimal, requirements files often contain an exhaustive listing of pinned versions for the purpose of achieving repeatable installations of a complete environment.
```

volodymyrss (changes requested):

burnout87:
`install_req` now has a minimal approach, and follows the same approach used in the `dispatcher-plugin-antares`. 

Tests succeed now.

setup.py
volodymyrss: Only really the things used by the bare dispatcher in operations, and only references to pypi. Maybe it's worth revising if all packages listed are actually used.
tests_require equivalent can contain things needed for testing dispatcher, as long as they are in pypi.
Then, master branch of some packages, as well as plugins (needed for the environment) can be installed over it, from requirements.txt . I guess.

volodymyrss (approved):

PR #96, opened 2021-05-11
extend parameters list

burnout87:
simply for the empty instrument numerical query


burnout87:
Just to include some parameters (max_pointings, osa_version, data_rights).

But I am not 100% sure this is the correct place, probably a change in the plugin is required

volodymyrss:
Indeed, I suspect dispatcher-app should not know about osa_version except possibly (and even then only arguably) in a test example.

burnout87:
This was started, but then realized was not needed here

PR #98, opened 2021-05-12
more requirements

volodymyrss:


setup.py
```diff
@@ -20,7 +20,27 @@
...
+    "bokeh==0.12.16",
+    "json_tricks",
+    "flask-restx",
+    "six",
+    "werkzeug<=0.16.1",
+    "python-shell-colors",
+    "logging_tree",
+    "celery",
+    "redis",
+    "pyjwt"
```
burnout87, line 43: Is this the list coming from the requirements, less some that are not strictly necessary?

burnout87 (commented):
just a question

setup.py
volodymyrss: except these ones are actually strictly necessary, since the module can not be even imported without them.

burnout87 (approved):

PR #104, opened 2021-05-16
Allow non tls smtp relay

volodymyrss:


volodymyrss:
on isolated network, non-tls smtp relay may exist safely. Would be also possible to configure it with an option. But trying is a way too.

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -1430,7 +1436,10 @@ def send_email(self, status="done",
             server = smtplib.SMTP(smtp_server, port)
             # just for testing purposes, not ssl is established
             if smtp_server != "localhost":
-                server.starttls(context=context)
+                try:
+                    server.starttls(context=context)
+                except Exception as e:
+                    self.logger.warning(f'unable to start TLS: {e}')
```
burnout87, line 1442: Good! Didn't think about it.

burnout87 (approved):

PR #106, opened 2021-05-17
fixtures with state

volodymyrss:


volodymyrss:
the change in fixtures is not breaking existing approach, and it speeds up currently available 144 tests (maybe too much) from >1h to ~8 min.

cdci_data_analysis/pytest_fixtures.py
```diff
@@ -352,27 +367,78 @@ def follow_output():
...
+    return dict(
+        url=service, 
+        pid=p.pid
+        )        
 
-    
-    print("will keep service alive a bit, for async")
-    time.sleep(0.5)
+@pytest.fixture
+def dispatcher_long_living_fixture(pytestconfig, dispatcher_test_conf_fn):
```
burnout87, line 376: I like the idea! But where do you use this fixture? In place of `dispatcher_live_fixture`?

.github/workflows/python-app.yml
```diff
@@ -46,7 +46,7 @@ jobs:
 
     - name: Test with pytest
       run: |
-        python -m pytest tests --cov=cdci_data_analysis -sv --full-trace --log-cli-level=DEBUG -m 'not isgri_plugin and not dda'
+        DISPATCHER_MOCK_KB=yes python -m pytest tests --cov=cdci_data_analysis -sv --full-trace --log-cli-level=DEBUG -m 'not isgri_plugin and not dda'
```
burnout87, line 49: What is the usage of this env_var?

burnout87 (commented):
couple of questions

cdci_data_analysis/pytest_fixtures.py
volodymyrss: yes, behaves exactly the same. Just we need to be cautious that this way dispatcher initialization is not tested every time, and there could be some accumulating issues. So it's best to reserve it for really demanding tests.

burnout87 (approved):

PR #107, opened 2021-05-17
removed some fields for products url generation, renamed import, nami…

burnout87:
Adjusted also the naming of some functions. Should now be more consistent with its purpose

### DIFF
--- a/cdci_data_analysis/flask_app/dispatcher_query.py
+++ b/cdci_data_analysis/flask_app/dispatcher_query.py
@@ -571,11 +571,11 @@ class InstrumentQueryBackEnd:
             if self.is_email_to_send_callback(status, time_original_request):
 
                 # build the products URL
-                request_url = self.generate_request_url_call_back(self.config.products_url, session_id, self.job_id)
+                products_url = self.generate_products_url_from_file(self.config.products_url, session_id, self.job_id)
                 self.send_email(status,
                                 instrument=instrument_name,
                                 time_request=time_original_request,
-                                request_url=request_url)
+                                request_url=products_url)
                 job.write_dataserver_status(status_dictionary_value=status,
                                             full_dict=self.par_dic,
                                             email_status='email sent')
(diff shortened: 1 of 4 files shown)


burnout87:
The token, as well as session_id and job_id, are not part the product_url within the email, but it might still be part of the product url from the frontend. Therefore the frontend needs to be updated accordingly

volodymyrss:
> The token, as well as session_id and job_id, are not part the product_url within the email, but it might still be part of the product url from the frontend. Therefore the frontend needs to be updated accordingly

We need to check if frontend actually generates it. It thought it does not, and it really should not. It's better if there is a single function doing it, and it is in the dispatcher already.
Oh, the frontend receives par_dict from dispatcher, this list should be cleaned too.

burnout87:
> > The token, as well as session_id and job_id, are not part the product_url within the email, but it might still be part of the product url from the frontend. Therefore the frontend needs to be updated accordingly
> 
> We need to check if frontend actually generates it. It thought it does not, and it really should not. It's better if there is a single function doing it, and it is in the dispatcher already.
> Oh, the frontend receives par_dict from dispatcher, this list should be cleaned too.

I am not familiar with the frontend code, so perhaps I looked in the wrong place, but it seems like that such url is generated.

The list of ignored fields should be updated, to make things easy of course.

[https://github.com/oda-hub/frontend-astrooda/blob/86b9874b1a2d927f9dbf25c940e46ee7f26adfe1/js/astrooda.instrument.js#L1277-L1294](https://github.com/oda-hub/frontend-astrooda/blob/86b9874b1a2d927f9dbf25c940e46ee7f26adfe1/js/astrooda.instrument.js#L1277-L1294)

volodymyrss (changes requested):
hm, ok, but a test?

burnout87:
> hm, ok, but a test?

The current test in `test_job_management` currently checks the product url, just not directly but using the content of thje email

volodymyrss:
> > > The token, as well as session_id and job_id, are not part the product_url within the email, but it might still be part of the product url from the frontend. Therefore the frontend needs to be updated accordingly
> > 
> > 
> > We need to check if frontend actually generates it. It thought it does not, and it really should not. It's better if there is a single function doing it, and it is in the dispatcher already.
> > Oh, the frontend receives par_dict from dispatcher, this list should be cleaned too.
> 
> I am not familiar with the frontend code, so perhaps I looked in the wrong place, but it seems like that such url is generated.
> 
> The list of ignored fields (`ignore_params_url`) should be updated, to make things easy of course.
> 
> https://github.com/oda-hub/frontend-astrooda/blob/86b9874b1a2d927f9dbf25c940e46ee7f26adfe1/js/astrooda.instrument.js#L1277-L1294

indeed! Good point.
I started to doubt it fully comes from the dispatcher when I did not see it in the response.
It gives you an idea what is ignored.
What is unfortunate is that ignore_params_url should be now synchronized with its done by generate_products_url_from_par_dict.
Maybe in the future we can make sure they come from the same place.
But for now, can we make it as you suggest and make a TODO comment?
Feel free to make a PR on frontend-astrooda!
We have now couple of PRs on there, we should agree with @motame tomorrow how to handle them efficiently.

burnout87:
> > `test_job_management`
> 
> but it does not test that newly ignored keys are ignored.

I check now that the product url I expect does not contain the ignored parameters, before checking the entire content of the received email.

volodymyrss (approved):

PR #117, opened 2021-05-19
response schema

volodymyrss:


volodymyrss:
Purpose of request for review in this case is so that we are all aware what these status fields mean.

cdci_data_analysis/flask_app/app.py
```diff
@@ -221,6 +227,26 @@ def common_exception_payload():
...
+    message = fields.Str(desciption="if query_status == 'failed', shown in waitingDialog in red")     
+    error_message = fields.Str(desciption="if query_status == 'failed', shown in waitingDialog in red")     
+    debug_message = fields.Str(desciption="if query_status == 'done' but exit_status.status != 0, shown in waitingDialog in red")     
+    comment = fields.Str(desciption="always, shown in waitingDialog in yellow")     
+    warning = fields.Str(desciption="")     
+    
+
+class QueryOutJSON(Schema):
+    query_status = fields.Str(
+                        validate=OneOf(["done", "failed"]),
```
burnout87, line 242: `submitted` status not considered?

cdci_data_analysis/flask_app/app.py
```diff
@@ -221,6 +227,26 @@ def common_exception_payload():
     return payload
 
 
+
+class ExitStatus(Schema):
+    status = fields.Int(validate=OneOf([0, 1]))
+    message = fields.Str(desciption="if query_status == 'failed', shown in waitingDialog in red")
```
burnout87, line 233: I guess it's description

burnout87 (changes requested):

burnout87:
> Purpose of request for review in this case is so that we are all aware what these status fields mean.

I guess, that more work within the dispatcher will have to be done in order to output objects of this format, am I right?

volodymyrss:
> > Purpose of request for review in this case is so that we are all aware what these status fields mean.
> 
> I guess, that more work within the dispatcher will have to be done in order to output objects of this format, am I right?

not really, this is a description of json output (and input) schema, in openapi/swagger, which is annotation for restful and otherwise http services.

So what will be needed is just switching on the validation using these schemas.
I am just concerned that if we switch them on now something will not comply.
So we can be concerned with this after.
But already now, this gives an idea of the description needed for current goal: communication with frontend.

cdci_data_analysis/flask_app/app.py
volodymyrss: you guess correctly! It would fail if it was applied.

cdci_data_analysis/flask_app/app.py
volodymyrss: I add, but it does not really matter, since it is not currently exhaustive: only parts needed for current cause - reporting on frontend - are described.

burnout87 (approved):

dsavchenko (approved):

PR #119, opened 2021-05-20
test email with oda_api

burnout87:
probably not the cleanest way, but I think it shows that emails are sent when they should and when they should not

### DIFF
--- a/cdci_data_analysis/analysis/queries.py
+++ b/cdci_data_analysis/analysis/queries.py
@@ -500,6 +500,8 @@ class ProductQuery(BaseQuery):
 
                 if job_status=='done':
                     job.set_done()
+                elif job_status == 'failed':
+                    job.set_failed()
                 else:
                     job.set_submitted()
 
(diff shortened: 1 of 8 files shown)


burnout87:
I also applied a change to how the status of a job was applied, but I am not sure if it makes sense of the logic behind the first time this was implemented

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -209,6 +209,7 @@ def __init__(self, app, instrument_name=None, par_dic=None, config=None, data_se
             query_out.set_query_exception(
                 e, 'InstrumentQueryBackEnd constructor', extra_message='InstrumentQueryBackEnd constructor failed')
 
+            #TODO I guess these comments are no longer needed
```
volodymyrss, line 212: indeed

volodymyrss (approved):

cdci_data_analysis/analysis/queries.py
```diff
@@ -500,6 +500,8 @@ def get_query_products(self,instrument,job,run_asynch,query_type='Real',logger=N
 
                 if job_status=='done':
                     job.set_done()
+                elif job_status == 'failed':
+                    job.set_failed()
```
burnout87, line 504: forgot to ask, is this ok to be done here ?

volodymyrss: I guess so.

PR #122, opened 2021-05-20
Email "To" field was empty due to different assumption on email field.

volodymyrss:
I put both.

### DIFF
--- a/cdci_data_analysis/analysis/tokenHelper.py
+++ b/cdci_data_analysis/analysis/tokenHelper.py
@@ -19,7 +19,11 @@ def get_token_user(decoded_token):
 
 def get_token_user_email_address(decoded_token):
     # extract user email address
-    return decoded_token['sub'] if 'sub' in decoded_token else ''
+    if 'sub' in decoded_token:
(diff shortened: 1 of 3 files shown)


cdci_data_analysis/analysis/tokenHelper.py
```diff
@@ -19,7 +19,11 @@ def get_token_user(decoded_token):
 
 def get_token_user_email_address(decoded_token):
     # extract user email address
-    return decoded_token['sub'] if 'sub' in decoded_token else ''
+    if 'sub' in decoded_token:
+        return decoded_token['sub']
+    if 'email' in decoded_token:
+        return decoded_token['email']
+    return ''
```
burnout87, line 26: In which was this happening? Did the token used have a different format?

burnout87 (changes requested):

cdci_data_analysis/analysis/tokenHelper.py
volodymyrss: Yes, token may have different format.

PR #124, opened 2021-05-20
more log on mail sending, removed duplicate check

volodymyrss:


volodymyrss:
or maybe not. Unclear. Need to define

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -669,7 +669,7 @@ def is_email_to_send_run_completion(self, status):
     def is_email_to_send_callback(self, status, time_original_request):
         if not self.public:
             # in case the request was long and 'done'
-            if status == 'done':
+            if status in ['done', 'task_complete']:
```
burnout87, line 672: Is `task_complete` a status that can be returned? I ask just because I have never seen 'til now. Does it have the same meaning as `done`?

burnout87 (changes requested):

cdci_data_analysis/flask_app/dispatcher_query.py
volodymyrss: There are many statuses meaning different things. Particular choice will end up in different email rate. So maybe it's best configurable too. Later.

burnout87: I was about indeed to suggest a configurable approach

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -679,11 +682,18 @@ def is_email_to_send_callback(self, status, time_original_request):
...
                 email_sending_timeout = tokenHelper.get_token_user_sending_timeout_email(self.decoded_token)
                 if email_sending_timeout is None:
                     email_sending_timeout = self.app.config.get('conf').email_sending_timeout
 
-                return email_sending_timeout and duration_query > timeout_threshold_email and status == 'done'
+                logger.info("email_sending_timeout: %s", email_sending_timeout)
+                logger.info("duration_query > timeout_threshold_email %s", duration_query > timeout_threshold_email)
+                logger.info("email_sending_timeout and duration_query > timeout_threshold_email %s", email_sending_timeout and duration_query > timeout_threshold_email)
+
+                return email_sending_timeout and duration_query > timeout_threshold_email
```
burnout87, line 696: Thanks

burnout87 (approved):

PR #129, opened 2021-05-21
Check if in debug mode when importing plugins

burnout87:


tests/test_configurer.py
```diff
@@ -2,14 +2,6 @@
 import os
 import pytest
 
-@pytest.fixture
-def dispatcher_debug(monkeypatch):
-    monkeypatch.setenv('DISPATCHER_DEBUG_MODE', 'yes')
-
-@pytest.fixture
-def dispatcher_nodebug(monkeypatch):
```
volodymyrss, line 10: oh nice!
Can we test that this has the desired effect on the instrument list?

volodymyrss (changes requested):

tests/test_server_basic.py
```diff
@@ -93,6 +93,41 @@ def test_empty_request(dispatcher_live_fixture):
...
+
+    print("content:", c.text)
+
+    jdata=c.json()
+
+    assert c.status_code == 400
+
+     # parameterize this
+    assert jdata['installed_instruments'] == [] or \
+           jdata['installed_instruments'] == ['spi_acs']
```
volodymyrss, line 112: is this one still needed? we might have moved all plugin tests out?

volodymyrss (changes requested):

tests/test_server_basic.py
burnout87: ops I will adjust it

cdci_data_analysis/flask_app/app.py
```diff
@@ -208,14 +208,16 @@ def common_exception_payload():
...
+        'DISPATCHER_DEBUG_MODE', 'no')  # change the default
 
-    if payload['debug_mode'] == "yes":
-        payload['config'] = {
-            'dispatcher-config': remove_nested_keys(app.config['conf'].as_dict(),
-                                                    ['sentry_url', 'logstash_host', 'logstash_port', 'secret_key',
-                                                     'smtp_server_password'])
-        }
+    # TODO why only in debug_mode ?
+    # if payload['debug_mode'] == "yes":
```
burnout87, line 214: Is ok to remove the nested keys in any case? Before it was happening only in case of debug_mode

burnout87 (commented):
one question regarding the usage of the debug_mode

cdci_data_analysis/flask_app/app.py
volodymyrss: it should be safe

volodymyrss (approved):

PR #133, opened 2021-05-21
Email formatting

burnout87:
Also a reformatting of all the email functinoalities.

Dispatcher_query was becoming way big


burnout87:
Ideally we could also consider whether or not to have the `job_id` within the email subject.

volodymyrss:
we should have job_id within subject. Maybe indeed adding "job_id" before it for clarity

volodymyrss:
I am adding also:
* html url
* parameter table
* time spent 
* small api code

burnout87:
It is not there in the current format, but also the username.

cdci_data_analysis/flask_app/templates/submitted.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
+<html lang="en">
+<body>
+
+Dear user,
+
+you receive this message because you submitted a request to get from the service provided by University of Geneva available at the URL {}.
```
dsavchenko, line 7: "University of Geneva" should be variable (though default)

burnout87: You mean that will change depending on the deployment? Makes sense.

volodymyrss:
maybe also to add url to make an issue

cdci_data_analysis/analysis/email_helper.py
```diff
@@ -0,0 +1,220 @@
...
+        context = ssl.create_default_context()
+        #
+        # Try to log in to server and send email
+        server = smtplib.SMTP(smtp_server, port)
+        # just for testing purposes, not ssl is established
+        if smtp_server != "localhost":
+            try:
+                server.starttls(context=context)
+            except Exception as e:
+                logger.warning(f'unable to start TLS: {e}')
```
dsavchenko, line 127: Also, need an option for when the server requires SSL over 465 from the beginning, by using SMTP_SSL 
In this case, simple SMTP may just be stuck for a long time.

cdci_data_analysis/analysis/email_helper.py
```diff
@@ -0,0 +1,220 @@
...
+            'time_request': time_request,
+            'request_url': request_url,
+            'api_code_no_token': api_code_no_token,
+        }
+    }
+
+    template = env.get_template('email.html')
+    email_body_html = template.render(**email_data)
+    
+    email_subject = re.search("<title>(.*?)</title>", email_body_html).groups()[0]
```
dsavchenko, line 85: Not important, but why not .group(1)?

dsavchenko (changes requested):

cdci_data_analysis/analysis/email_helper.py
volodymyrss: no particular reason. Is it not unfortunate groups are counted from 1?..

cdci_data_analysis/analysis/email_helper.py
volodymyrss: do you have a place where this is run? if yes and you need it, please make an issue.

volodymyrss:
Rebranded into #143 , to maintain review workflow

PR #134, opened 2021-05-21
check number emails done and submitted

burnout87:


burnout87:
More tests are needed

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -648,12 +648,15 @@ def is_email_to_send_run_completion(self, status):
...
+            email_jobs_folders = glob.glob(f'scratch_sid_*_jid_{self.job_id}*/email_history')
+            if len(email_jobs_folders) >= 1:
+            # if os.path.exists(self.scratch_dir + '/email_history'):
+                submitted_email_files = glob.glob(f'scratch_sid_*_jid_{self.job_id}*/email_history/email_submitted_*.email')
                 if len(submitted_email_files) >= 1:
-                    last_submitted_email_sent = submitted_email_files[len(submitted_email_files) - 1]
-                    f_name, f_ext = os.path.splitext(os.path.basename(last_submitted_email_sent))
-                    time_last_email_submitted_sent = float(f_name.split('_')[3])
+                    times = []
+                    [times.append(float(os.path.splitext(os.path.basename(f))[0].split('_')[2])) for f in submitted_email_files]
```
volodymyrss, line 658: is this really a more undestandable version of two-line loop? :)

volodymyrss: if you do want to use list comprehesion, even if it is a long line, why not combine two lines and avoid append?

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -689,7 +692,13 @@ def is_email_to_send_callback(self, status, time_original_request):
                 logger.info("duration_query > timeout_threshold_email %s", duration_query > timeout_threshold_email)
                 logger.info("email_sending_timeout and duration_query > timeout_threshold_email %s", email_sending_timeout and duration_query > timeout_threshold_email)
 
-                return email_sending_timeout and duration_query > timeout_threshold_email
+                number_done_emails_ok = True
+                if os.path.exists(self.scratch_dir + '/email_history'):
+                    done_email_files = glob.glob(self.scratch_dir + '/email_history/email_done_*.email')
+                    if len(done_email_files) >= 1:
+                        logger.info("number of done emails sent: %s", len(done_email_files))
```
volodymyrss, line 699: we should issue a warning here, to sentry. this is somehow anomalous, and we need to be aware

tests/test_job_management.py
```diff
@@ -500,15 +511,198 @@ def test_email_submitted(dispatcher_live_fixture, dispatcher_local_mail_server):
...
+
+    assert os.path.exists(smtp_server_log)
+    f_local_smtp = open(smtp_server_log)
+    f_local_smtp_jdata = json.load(f_local_smtp)
+
+    assert len(f_local_smtp_jdata) == 3
+
+
+@pytest.mark.not_safe_parallel
+def test_email_submitted_multiple_requests(dispatcher_live_fixture, dispatcher_local_mail_server):
```
volodymyrss, line 547: there is some repetition on inspecting emails, maybe it can be taken out into some common functionality (similar to these ask/loop functions). It will be then easier to read the individual tests and also they can be reused in plugins.
A minor suggestion, maybe not now.

volodymyrss (changes requested):
Small suggestions

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: I tried to compact a number of instructions. Sorry I didn't get what you mean with the combination of two lines instead of the append.

volodymyrss: I mean, if you do want to use list comprehension, is this not going to do what you want:

```
times = [ float(os.path.splitext(os.path.basename(f))[0].split('_')[2])) for f in submitted_email_files ]
```

while also using generated list.

Relying on side-effects of list comprehension is a bit unusual and arguably quite unpythonic. 
E.g. see some community discussion [here](https://stackoverflow.com/questions/5753597/is-it-pythonic-to-use-list-comprehensions-for-just-side-effects) and similar.

Maybe I am missing your intent.

burnout87: True, let's not going unpythonic. Didn't consider these, actually very important, aspect of the side-effects that could actually be dangerous.

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -586,6 +590,11 @@ def run_call_back(self, status_kw_name='action') -> typing.Tuple[str, typing.Uni
                                             email_status='email sent')
             else:
                 job.write_dataserver_status(status_dictionary_value=status, full_dict=self.par_dic)
+        except MultipleDoneEmail as e:
```
volodymyrss, line 593: this should not, however, break the current situation when the product is eventually recieved by the user, after several "submit-done" cycles. So write_dataserver_status should still happen, perhaps with a note that it's duplicated done.

burnout87:
Any additional comments @volodymyrss ?

volodymyrss (approved):
ok I think, as far  as I looked.

PR #144, opened 2021-05-31
Update email.html

volodymyrss:


volodymyrss:
@ferrigno you can also just edit the file in github browser directly, and then select "make PR" radio instead of "make" change.
Please have a look if this what you meant!

PR #151, opened 2021-05-31
add short-lived token and extra note about it

volodymyrss:
image


ferrigno:
New stuff is OK, but 

Note that you still have "the following this url" 

I also suggest to remove PS and the following sentence with 

"Python API code"

ferrigno:
Consider this as approved with the corrections suggested above

burnout87 (approved):

PR #156, opened 2021-05-31
test download products

burnout87:
Now, in case the download products is requested, the instrument is set to `mock`,

On the frontend side. the instrument should, ideally, not be set

### DIFF
--- a/cdci_data_analysis/flask_app/dispatcher_query.py
+++ b/cdci_data_analysis/flask_app/dispatcher_query.py
@@ -231,6 +231,9 @@ class InstrumentQueryBackEnd:
         # this takes care of various strange objects which can not be properly represented
         return format_hash(json.dumps(o))
 
+    def calculate_job_id(self, par_dic):
+        return u'%s' % (self.make_hash(par_dic))
+
     # not job_id??
     def generate_job_id(self, kw_black_list=['session_id', 'job_id', 'token']):
         self.logger.info("\033[31m---> GENERATING JOB ID <---\033[0m")
@@ -254,7 +257,7 @@ class InstrumentQueryBackEnd:
             # token has not been considered, but the user id will be (if availaable)
             _dict['sub'] = tokenHelper.get_token_user_email_address(self.decoded_token)
 
-        self.job_id = u'%s' % (self.make_hash(_dict))
+        self.job_id = self.calculate_job_id(_dict)
 
         self.logger.info(
             '\033[31mgenerated NEW job_id %s \033[0m', self.job_id)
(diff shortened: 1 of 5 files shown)


volodymyrss (changes requested):
Please add a test, else we will break it again.

burnout87:
> Please add a test, else we will break it again.

Yes, I was planning it...

volodymyrss:
> > Please add a test, else we will break it again.
> 
> Yes, I was planning it...

Sorry was it unclear! Since review requested and no draft status.

cdci_data_analysis/pytest_fixtures.py
```diff
@@ -434,6 +434,25 @@ def dispatcher_long_living_fixture(pytestconfig, dispatcher_test_conf_fn, dispat
...
+    )
+    DispatcherJobState.remove_scratch_folders(job_id=job_id)
+    scratch_dir_path = f'scratch_sid_{session_id}_jid_{job_id}'
+    # set the scratch directory
+    os.makedirs(scratch_dir_path)
+
+    with open(scratch_dir_path + '/test.fits.gz.recovered', 'wb') as fout:
+        fout.write(os.urandom(20))
+
+    yield scratch_params
```
volodymyrss, line 455: I would also return the data written, and later compare. It will be send through flask and all, could be some confusion.

burnout87: it is exactly what I am trying to do

burnout87:
I am not sue I handled the returned file via Flask, in the best way, so I am open to improvements.

tests/test_server_basic.py
```diff
@@ -221,6 +220,40 @@ def test_valid_token(dispatcher_live_fixture):
...
+            'session_id': session_id,
+            'job_id': job_id
+        }
+
+    c = requests.get(server + "/download_products",
+                     params=params)
+
+    print("content:", c.text.encode())
+
+    assert c.status_code == 200
```
volodymyrss, line 245: c.content is bytes of response, it can be inspected and compared with ` empty_products_files_fixture['content']`

burnout87: that is actually some test code, that serves no purpose for this test

volodymyrss: ok, but do you test the result of the request?

burnout87: Yes, it check that the content returned matches

volodymyrss: But where is the returned content?

tests/test_server_basic.py
```diff
@@ -221,6 +220,55 @@ def test_valid_token(dispatcher_live_fixture):
...
+
+    # download the output, read it and then compare it
+    with gzip.open(f'scratch_sid_{session_id}_jid_{job_id}/output_test', 'wb') as fout:
+        fout.write(empty_products_files_fixture['content'])
+    data_downloaded = ''
+    with gzip.open(f'scratch_sid_{session_id}_jid_{job_id}/output_test', 'rb') as fout:
+        data_downloaded = fout.read()
+
+    assert data_downloaded == empty_products_files_fixture['content']
+
```
volodymyrss, line 270: you are writing `empty_products_files_fixture['content']` to a file, reading it back, and verifying that this did not change the content, right?

This process is not involving flask or dispatcher, so they are not tested.

burnout87: My bad. Adjusting

burnout87:
Now it actually verify the downloaded content matches with the written one, but I am still not entirely convinced I am reading/writing correctly.

tests/test_server_basic.py
```diff
@@ -221,6 +220,55 @@ def test_valid_token(dispatcher_live_fixture):
...
+    params = {
+            'instrument': instrument,
+            # since we are passing a job_id
+            'query_status': 'ready',
+            'file_list': 'test.fits.gz',
+            'download_file_name': 'output_test',
+            'session_id': session_id,
+            'job_id': job_id,
+            'token': encoded_token
```
volodymyrss, line 254: is token verified? is there an invalid token case?

volodymyrss (changes requested):
One thing that I think is still lacking is test of token verification, and making sure token corresponds to the jobid.

tests/test_server_basic.py
burnout87: Yes it is, but I could add a case/test with an expired one

burnout87:
> One thing that I think is still lacking is test of token verification, and making sure token corresponds to the jobid.

What do you mean with token corresponding to the job_id ?

volodymyrss:
> > One thing that I think is still lacking is test of token verification, and making sure token corresponds to the jobid.
> 
> What do you mean with token corresponding to the job_id ?

Now, it will be possible for anyone with a valid token to request any valid job_id, right?

job_id is specific to each user. It would be possible to check that job_id matches the analysis parameters found in scratch dir + user details.
Or we can leave it for later, with a TODO.

burnout87:
Test also with a user different from the one used for the `job_id` of the request.
Not entirely sure how to handle the `off_line` and `api` parameters.

tests/test_server_basic.py
burnout87: more tests are now there: one with authorized/matching user and one with an unauthorized/unmatching user

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -419,19 +425,33 @@ def prepare_download(self, file_list, file_name, scratch_dir):
...
-
-        tmp_dir, target_file = self.prepare_download(
-            file_list, file_name, self.scratch_dir)
-        #print('downlaoding scratch dir', self.scratch_dir)
         try:
+            # TODO not entirely sure about these
+            self.off_line = False
+            self.api = False
+
+            if not self.public:
```
volodymyrss, line 433: and if not public, data from any job_id can be downloaded?

burnout87: Adding also this check

burnout87: So, in case the request is public, a check of the provided `job_id`, against the one calculated with the analysis_parameter from the scratch directory (and without any token since this is not sent) is performed.

burnout87: Does it make sense?

volodymyrss (approved):

PR #162, opened 2021-06-02
validation job_id for run_analysis and call_back

burnout87:
Is it helpful to have this check also during a `call_back` ?

### DIFF
--- a/cdci_data_analysis/analysis/exceptions.py
+++ b/cdci_data_analysis/analysis/exceptions.py
@@ -37,7 +37,7 @@ class RequestNotUnderstood(BadRequest):
 
 
 class RequestNotAuthorized(BadRequest):
-    def __init__(self, message, debug_message):
+    def __init__(self, message, debug_message=''):
         self.debug_message = debug_message
         super().__init__(message, status_code=403)
 
--- a/cdci_data_analysis/analysis/exceptions.py
+++ b/cdci_data_analysis/analysis/exceptions.py
@@ -37,7 +37,8 @@ class RequestNotUnderstood(BadRequest):
 
 
 class RequestNotAuthorized(BadRequest):
-    def __init__(self, message):
+    def __init__(self, message, debug_message):
+        self.debug_message = debug_message
         super().__init__(message, status_code=403)
 
 
--- a/cdci_data_analysis/flask_app/app.py
+++ b/cdci_data_analysis/flask_app/app.py
@@ -273,7 +273,6 @@ def run_analysis():
     request_summary = log_run_query_request()
 
     try:
-        # t0 = _time.time()
         t0 = g.request_start_time
         query = InstrumentQueryBackEnd(app)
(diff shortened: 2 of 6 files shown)


volodymyrss (changes requested):
oh, could we add here more informative message following 

https://github.com/oda-hub/doc-review-guidelines/issues/27

volodymyrss:
> Is it helpful to have this check also during a `call_back` ?

call_back does not provide parameters, no? what would you compare?

burnout87:
> > Is it helpful to have this check also during a `call_back` ?
> 
> call_back does not provide parameters, no? what would you compare?

It would read, within the scratch folder, the analysis_parameters.json and calculate the job_id from it, and then compare the calculated one with the provided one (with, of course, the token considered)

volodymyrss:
> > > Is it helpful to have this check also during a `call_back` ?
> > 
> > 
> > call_back does not provide parameters, no? what would you compare?
> 
> It would read, within the scratch folder, the analysis_parameters.json and calculate the job_id from it, and then compare the calculated one with the provided one (with, of course, the token considered)

this would serve quite different purpose than that with run_analysis. with run_analysis, it is actually possible for client to send incorrect parameters with job_id.
this actually happened.

comparison of `job_id` and `analysis_parameters.json` found in directory with that job_id will only fail if directory is written incorrectly, or call back url was constructed incorrectly.
in principle, this is all internal dispatcher business, but it's good to have for verification.
If not complex, is useful. But then I would rather focus on more impactful issues.

burnout87:
> > > > Is it helpful to have this check also during a `call_back` ?
> > > 
> > > 
> > > call_back does not provide parameters, no? what would you compare?
> > 
> > 
> > It would read, within the scratch folder, the analysis_parameters.json and calculate the job_id from it, and then compare the calculated one with the provided one (with, of course, the token considered)
> 
> this would serve quite different purpose than that with run_analysis. with run_analysis, it is actually possible for client to send incorrect parameters with job_id.
> this actually happened.
> 
> comparison of `job_id` and `analysis_parameters.json` found in directory with that job_id will only fail if directory is written incorrectly, or call back url was constructed incorrectly.
> in principle, this is all internal dispatcher business, but it's good to have for verification.
> If not complex, is useful. But then I would rather focus on more impactful issues.

the check was actually quite easy to implement

volodymyrss (approved):

burnout87:
After some thoughts I decided to remove the check during `call_back`, since it is out of the scope from the current issue we are addressing, and also to avoid breaking something else

volodymyrss:
> After some thoughts I decided to remove the check during `call_back`, since it is out of the scope from the current issue we are addressing, and also to avoid breaking something else

Ok! We can always leave a TODO in code  and an issue on github, with reference to this PR. This will allow us to come back to it later easily!

burnout87:
#166 for future implementation

PR #163, opened 2021-06-02
Corrupted long lines in email: URLs, API code

volodymyrss:
emails generally can not contain lines longer than 999 characters.
different SMTP servers will deal with these differently: 
* some will respond with error, 
* some, apparently, automatically introduce new line 

The latter  may cause an issue if it is added in the middle of data,  e.g. in some random place in json 
We need:

* compress (zip) some long parameters, e.g. selected_catalog. They cause issues in other cases already.
* request by shortcut (job_d): but it is clear that it is not generally possible to derive parameters from job_id,  works like a URL shortener, i.e. to reverse this process. We should add a warning in this case.
* to detect this and be clear we can not send these long lines. they are not often usable as URLs anyway

Test of email api code and url also highlighted difference between ignored keys in these two kind of requests, in part related to some duplication of job_id generation.

### DIFF
--- a/cdci_data_analysis/configurer.py
+++ b/cdci_data_analysis/configurer.py
@@ -233,7 +233,7 @@ class ConfigEnv(object):
                                      disp_dict['email_options']['email_sending_timeout'],
                                      disp_dict['email_options']['email_sending_timeout_default_threshold'],
                                      disp_dict['email_options']['email_sending_job_submitted'],
-                                     disp_dict['email_options'].get('email_sending_job_submitted_default_interval', 0)
+                                     disp_dict['email_options']['email_sending_job_submitted_default_interval']
                                      )
 
         # not used?
--- a/cdci_data_analysis/configurer.py
+++ b/cdci_data_analysis/configurer.py
@@ -233,7 +233,7 @@ class ConfigEnv(object):
                                      disp_dict['email_options']['email_sending_timeout'],
                                      disp_dict['email_options']['email_sending_timeout_default_threshold'],
                                      disp_dict['email_options']['email_sending_job_submitted'],
-                                     disp_dict['email_options']['email_sending_job_submitted_default_interval']
+                                     disp_dict['email_options'].get('email_sending_job_submitted_default_interval', 0)
                                      )
 
         # not used?
--- a/cdci_data_analysis/flask_app/dispatcher_query.py
+++ b/cdci_data_analysis/flask_app/dispatcher_query.py
@@ -81,7 +81,7 @@ class InstrumentQueryBackEnd:
     def instrument_name(self, instrument_name):
         self._instrument_name = instrument_name
 
-    def __init__(self, app, instrument_name=None, par_dic=None, config=None, data_server_call_back=False, verbose=False, get_meta_data=False):
+    def __init__(self, app, instrument_name=None, par_dic=None, config=None, data_server_call_back=False, verbose=False, get_meta_data=False, download_products=False):
         # self.instrument_name=instrument_name
 
         self.logger = logging.getLogger(repr(self))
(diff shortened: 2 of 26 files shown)


cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -448,12 +437,16 @@ def validate_job_id(self):
...
 
+
     def download_products(self,):
         try:
             # TODO not entirely sure about these
             self.off_line = False
             self.api = False
 
+            # TODO: extract this into separate call?            
+            request_par_dic = self.get_request_par_dic()
```
dsavchenko, line 448: unused var

cdci_data_analysis/plugins/dummy_instrument/data_server_dispatcher.py
```diff
@@ -17,9 +17,7 @@
 
 from __future__ import absolute_import, division, print_function
 
-import urllib
-from builtins import (bytes, str, open, super, range,
-                      zip, round, input, int, pow, object, map, zip)
+import os
```
dsavchenko, line 20: If dropping this part of 2-to-3 transition, why not also drop future imports?

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -607,13 +601,14 @@ def run_call_back(self, status_kw_name='action') -> typing.Tuple[str, typing.Uni
...
                     job_id=self.job_id,
+                    session_id=self.par_dic['session_id'],
                     status=status,
                     instrument=instrument_name,
                     product_type=product_type,
                     time_request=time_original_request,
                     request_url=products_url,
                     api_code=DispatcherAPI.set_api_code(original_request_par_dic, 
-                                                        url=self.app.config['conf'].products_url),
+                                                        url=self.app.config['conf'].products_url + "/dispatch-data"),
```
burnout87, line 611: Will it always be the case to have `dispatch_data` as final part of the url ? I guess this depends on the deployment...

dsavchenko (approved):

cdci_data_analysis/configurer.py
```diff
@@ -233,7 +233,7 @@ def __init__(self, cfg_dict, origin=None):
                                      disp_dict['email_options']['email_sending_timeout'],
                                      disp_dict['email_options']['email_sending_timeout_default_threshold'],
                                      disp_dict['email_options']['email_sending_job_submitted'],
-                                     disp_dict['email_options']['email_sending_job_submitted_default_interval']
+                                     disp_dict['email_options'].get('email_sending_job_submitted_default_interval', 0)
```
burnout87, line 236: This interval has a default value which should be set in the configuration, and used in case it is not provided with the token. Is this for security reasons?

burnout87 (commented):
a couple of questions

cdci_data_analysis/flask_app/dispatcher_query.py
volodymyrss: It does not, it is how frontend-astrooda refers to the dispatcher, it's also hardcoded in all instruments.
e.g. there:
https://github.com/oda-hub/frontend-astrooda/blob/master/instruments/astrooda_isgri/astrooda_isgri.inc#L22

since it is a single unique alias, there is no reason for it to be configurable.

products_url is frontend, where users click. dispatch-data makes it dispatcher.
I considered to add a comment there since it's not obvious. I will do it.

cdci_data_analysis/configurer.py
volodymyrss: It might not be set, it was not set in example I guess.  It's not wrong to have sensible defaults in some cases.
But ok I revert for consistency.

cdci_data_analysis/plugins/dummy_instrument/data_server_dispatcher.py
volodymyrss: We can just do it all at once in a single commit, once enough tests are implemented.
Similarly, we can make a single commit for style alignment.
Commits mix features and style and larger then necessary.

PR #170, opened 2021-06-04
More clear message

burnout87:
Before it was related to the download of a product, now is more general, since the validation of the `job_id` happens both during a download as well as when `run_analysys`

### DIFF
--- a/cdci_data_analysis/flask_app/dispatcher_query.py
+++ b/cdci_data_analysis/flask_app/dispatcher_query.py
@@ -445,7 +445,7 @@ class InstrumentQueryBackEnd:
                 if not self.public:
                     debug_message += " for your user account email"
                 logstash_message(self.app, {'origin': 'dispatcher-call-back', 'event': 'unauthorized-user'})
-                raise RequestNotAuthorized("user not authorized to download the requested product", debug_message=debug_message)
+                raise RequestNotAuthorized("Request not authorized", debug_message=debug_message)
 
     def download_products(self,):
         try:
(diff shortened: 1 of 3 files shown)


dsavchenko (approved):

volodymyrss (approved):

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -446,7 +446,7 @@ def validate_job_id(self):
                 if not self.public:
                     debug_message += " for your user account email"
                 logstash_message(self.app, {'origin': 'dispatcher-call-back', 'event': 'unauthorized-user'})
-                raise RequestNotAuthorized("user not authorized to download the requested product", debug_message=debug_message)
```
volodymyrss, line 449: oh, sorry I overlooked and noticed while testing something else, this is not just called when download, right?

burnout87: Exactly! Also when `run_analysis', and perhaps in the future it will be during `call_back` (and for emails I saw something...?)

volodymyrss: it is already used in all kinds of places  where `not authorized to download the requested` is not right, since user is not trying to download anything. Maybe previous, more generic, message was better in this case.

burnout87: The current one is actually quite generic (`Request not authorized`) unless I misunderstood

volodymyrss: it says there in the first argument `not authorized to download the requested`, right?

burnout87: That is the old message, I have now re-written

[https://github.com/oda-hub/dispatcher-app/blob/03b473cdc785d5387f6e4e5f772b7dfd3cd8de07/cdci_data_analysis/flask_app/dispatcher_query.py#L449](https://github.com/oda-hub/dispatcher-app/blob/03b473cdc785d5387f6e4e5f772b7dfd3cd8de07/cdci_data_analysis/flask_app/dispatcher_query.py#L449)

volodymyrss: ok, I looked in a wrong place

PR #171, opened 2021-06-07
Fix for download problem by not adding parameter user_catalog_file to analysis_parameters.json, if it is None

dsavchenko:


dsavchenko:
> Shall we apply the check also when generating the `job_id` and creating the `analysis_parameters.json` ?

I am not sure if the situation possible when some parameter being empty intentionally?

burnout87:
> > Shall we apply the check also when generating the `job_id` and creating the `analysis_parameters.json` ?
> 
> I am not sure if the situation possible when some parameter being empty intentionally?

Exactly, probably is not necessary. It would just be an additional security check.

dsavchenko:
Ok, empty parameter values in the request are being converted into empty string in the file, not null. Then it is safe t implement these additional 'is not None' checks

burnout87:
> Ok, empty parameter values in the request are being converted into empty string in the file, not null. Then it is safe t implement these additional 'is not None' checks

In this way we are also 'consistent' with the way we treat the request parameters

burnout87 (approved):

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -232,7 +232,8 @@ def format_hash(x): return hashlib.md5(
         return format_hash(json.dumps(o))
 
     def calculate_job_id(self, par_dic):
-        return u'%s' % (self.make_hash(par_dic))
```
volodymyrss, line 235: I am not too sure why we have both `make_hash` and `calculate_job_id` but ok, it's not part of this PR.

cdci_data_analysis/analysis/instrument.py
```diff
@@ -551,7 +551,8 @@ def set_catalog_from_fronted(self,par_dic,request,back_end_query,logger=None,ver
             #print('POST')
             try:
                 cat_file_path = back_end_query.upload_file('user_catalog_file', back_end_query.scratch_dir)
-                par_dic['user_catalog_file'] = cat_file_path
+                if cat_file_path is not None:
```
volodymyrss, line 554: could there be a test? Cat uploading caused quite some problems, it should be tested. If you do not want, let's make an issue and we do soon after.

volodymyrss (changes requested):

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: `calculate_job_id` is used when generating a `job_id` starting from the dictionary of the request, and during the vavlidation of the `job_id` (calculation from the `analysis_parameters.json` file)

cdci_data_analysis/analysis/instrument.py
dsavchenko: I do not fully know this catalog functionality, and not aware of other issues. So anyway will need some clarifications to write meaningful tests. Thus better to split into another PR.

cdci_data_analysis/flask_app/dispatcher_query.py
dsavchenko: Another copy of the make_hash is in the pytests_fixtures

volodymyrss: yes, very good.

cdci_data_analysis/analysis/instrument.py
volodymyrss: ok #173

cdci_data_analysis/flask_app/dispatcher_query.py
volodymyrss: yes, very good.

volodymyrss (approved):

PR #177, opened 2021-06-16
dump analysis_parameters only first time

burnout87:
To address the issue when clicking the Fit button.

Some test is needed

### DIFF
--- a/cdci_data_analysis/analysis/products.py
+++ b/cdci_data_analysis/analysis/products.py
@@ -85,13 +85,13 @@ class QueryOutput(object):
     def set_api_code(self, query_dict, url=None):
         self.prod_dictionary['api_code'] = DispatcherAPI.set_api_code(query_dict, url=url)
 
-    def dump_analysis_parameters(self,work_dir,query_dict):
+    def dump_analysis_parameters(self, work_dir, query_dict):
         file_path = FilePath(file_dir=work_dir, file_name='analysis_parameters.json')
         if not os.path.exists(file_path.path):
             with open(file_path.path, 'w')  as outfile:
                 _dict = {k: v for k, v in query_dict.items() if v is not None}
                 my_json_str = json.dumps(_dict, indent=4, sort_keys=True)
-                outfile.write(u'%s' % my_json_str) # why is this formatting necessary?
+                outfile.write(u'%s' % my_json_str)
 
     def set_products(self, keys, values):
         for k, v in zip(keys, values):
--- a/cdci_data_analysis/analysis/products.py
+++ b/cdci_data_analysis/analysis/products.py
@@ -91,7 +91,13 @@ class QueryOutput(object):
             with open(file_path.path, 'w')  as outfile:
                 _dict = {k: v for k, v in query_dict.items() if v is not None}
                 my_json_str = json.dumps(_dict, indent=4, sort_keys=True)
+                # TODO further test if formatting is really needed
                 outfile.write(u'%s' % my_json_str)
+        else:
+            with open(file_path.path) as outfile:
+                dict_analysis_parameters = json.load(outfile)
+            logger.info("analysis_parameters.json file already present, and originally used for the job_id calculation: %s",
+                         json.dumps(dict_analysis_parameters, sort_keys=True, indent=4))
 
     def set_products(self, keys, values):
         for k, v in zip(keys, values):
--- a/cdci_data_analysis/analysis/products.py
+++ b/cdci_data_analysis/analysis/products.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
+import os
 from builtins import (bytes, str, open, super, range,
                       zip, round, input, int, pow, object, map, zip)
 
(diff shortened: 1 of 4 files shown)


volodymyrss:
> To address the issue when clicking the Fit button.
> 
> Some test is needed

You mean, yet another  test is needed?

burnout87:
> > To address the issue when clicking the Fit button.
> > Some test is needed
> 
> You mean, yet another test is needed?

That was a comment from yesterday...now there is one specific for the case.

cdci_data_analysis/analysis/products.py
```diff
@@ -84,12 +85,13 @@ def set_analysis_parameters(self, query_dict):
...
-            _dict = {k: v for k, v in query_dict.items() if v is not None}
-            my_json_str = json.dumps(_dict, indent=4, sort_keys=True)
-            outfile.write(u'%s' % my_json_str) # why is this formatting necessary?
+    def dump_analysis_parameters(self, work_dir, query_dict):
+        file_path = FilePath(file_dir=work_dir, file_name='analysis_parameters.json')
+        if not os.path.exists(file_path.path):
+            with open(file_path.path, 'w')  as outfile:
+                _dict = {k: v for k, v in query_dict.items() if v is not None}
+                my_json_str = json.dumps(_dict, indent=4, sort_keys=True)
+                outfile.write(u'%s' % my_json_str)
```
volodymyrss, line 94: no but really, why do we need here `u'%s'`?

cdci_data_analysis/analysis/products.py
```diff
@@ -84,12 +85,13 @@ def set_analysis_parameters(self, query_dict):
...
 
-    def dump_analysis_parameters(self,work_dir,query_dict):
-        file_path=FilePath(file_dir=work_dir, file_name='analysis_parameters.json')
-        with open(file_path.path, 'w')  as outfile:
-            _dict = {k: v for k, v in query_dict.items() if v is not None}
-            my_json_str = json.dumps(_dict, indent=4, sort_keys=True)
-            outfile.write(u'%s' % my_json_str) # why is this formatting necessary?
+    def dump_analysis_parameters(self, work_dir, query_dict):
+        file_path = FilePath(file_dir=work_dir, file_name='analysis_parameters.json')
+        if not os.path.exists(file_path.path):
```
volodymyrss, line 90: and if exists we could at least print them to highlight special situation.

tests/test_server_basic.py
```diff
@@ -184,6 +184,81 @@ def test_same_request_different_users(dispatcher_live_fixture):
...
+        'query_type': "Dummy",
+        'instrument': 'empty',
+        'token': encoded_token,
+        'session_id': session_id,
+        'job_id': job_id,
+        'query_status': "ready"
+    }
+
+    jdata = ask(server,
+                params,
```
volodymyrss, line 249: but are you not also sending exactly the same parameters?

volodymyrss (changes requested):

tests/test_server_basic.py
burnout87: no, I am just changing the 'query_status'

cdci_data_analysis/analysis/products.py
burnout87: I did some test and if it is not formatted, then it has some issues during the extraction, but I didn't dig any further

tests/test_server_basic.py
volodymyrss: so the situation as with fit is not happening, and even if the parameters were different it would not fail. And the special case of different parameters for the same job id, like with fit, is not tested.

volodymyrss (changes requested):
> I did some test and if it is not formatted, then it has some issues during the extraction, but I didn't dig any further 

ok, even if this is not the point of the PR, but we could at least leave the comment which was there.

tests/test_server_basic.py
burnout87: The `query_status` is used for the calculation of the job_id, so, from one test to the other, the parameters are different while having the same `job_id`. Unless I am missing a point

volodymyrss: `query_status` is **not** used in calculation of `job_id`:

https://github.com/oda-hub/dispatcher-app/blob/0657d9241c3980a888cf577a6943666b96f0d9b4/cdci_data_analysis/flask_app/dispatcher_query.py#L238-L245

And it really should not be. If it was, job_id would change when request progresses (e.g. to ready), and would be denied due to mismatching job_id

burnout87: damn now I see my mistake, I forgot to make a last pull, so my `restricted_par_dic` was not up to date, and `query_status` was not there.

volodymyrss (approved):

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -1201,7 +1201,12 @@ def run_query(self, off_line=False, disp_conf=None):
         self.api = api # TODO: we should decide if it's memeber or not
 
         try:
-            self.validate_job_id()
+            # TODO still I am not entirely sure this is a good approach
+            query_status = self.par_dic['query_status']
+            # the idea is that, if the query is new then, there is no analysis_parameters.json file
+            # otherwise, like in the case of the click of the fit button, query_status will be 'ready'
+            # and so an analysis_parameters.json dump is already there
+            self.validate_job_id(query_status != 'new')
```
volodymyrss, line 1209: for readability, I would make explicit kw argument here `request_parameters_from_scratch_dir=...`.
but in principle, it does not matter, and also this logic is rather confused and to be revised later.

volodymyrss (approved):

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: exactly, probably it will work just fine now, but I am not 100% sure regarding its stability

PR #189, opened 2021-06-24
changed location of setting par_dic

burnout87:
Had to re-use this branch and open a new PR since I had missed part of how values of the `par_dic` are set.

Previously was breaking another test in the plugin

### DIFF
--- a/cdci_data_analysis/analysis/instrument.py
+++ b/cdci_data_analysis/analysis/instrument.py
@@ -186,6 +186,7 @@ class Instrument:
                 sentry_client.capture('raven.events.Message',
                                            message=f'Error while setting input scw_list file from the frontend {e}')
             raise RequestNotUnderstood("Error while setting input scw_list from the frontend")
+        self.set_pars_from_dic(par_dic, verbose=verbose)
 
     def run_query(self, product_type,
                   par_dic,
@@ -441,7 +442,6 @@ class Instrument:
                 pass
             else:
                 raise RuntimeError
-        self.set_pars_from_dic(par_dic, verbose=verbose)
 
     def set_input_products(self, par_dic, input_file_path,input_prod_list_name):
         if input_file_path is None:
--- a/cdci_data_analysis/analysis/instrument.py
+++ b/cdci_data_analysis/analysis/instrument.py
@@ -148,12 +148,12 @@ class Instrument:
         if self.data_server_query_class is not None:
             return self.data_server_query_class(config=config,instrument=self).test_has_input_products(instrument,logger=logger)
 
-    def read_frontend_files(self,
-                            par_dic,
-                            request,
-                            temp_dir,
-                            verbose,
-                            sentry_client=None):
+    def read_frontend_inputs(self,
+                             par_dic,
+                             request,
+                             temp_dir,
+                             verbose,
+                             sentry_client=None):
         # TODO probably exception handling can be further improved and/or optmized
         # set catalog
         try:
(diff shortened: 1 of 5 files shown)


cdci_data_analysis/analysis/instrument.py
```diff
@@ -148,12 +148,12 @@ def test_has_input_products(self, config, instrument,logger=None):
         if self.data_server_query_class is not None:
             return self.data_server_query_class(config=config,instrument=self).test_has_input_products(instrument,logger=logger)
 
-    def read_frontend_files(self,
-                            par_dic,
-                            request,
-                            temp_dir,
-                            verbose,
-                            sentry_client=None):
+    def read_frontend_inputs(self,
```
volodymyrss, line 151: if you do want to organize the naming a bit, I would even call it something like "parse_inputs_from_files" or along those lines. This does not have to be coming from frontend.

even if oda_api currently does not do it, it might do in the future.

the purpose of this function appears to be to interpret (not readly "upload") files passed in the request and set par_dict accordingly. 
right?

cdci_data_analysis/analysis/instrument.py
```diff
@@ -186,6 +186,7 @@ def read_frontend_files(self,
                 sentry_client.capture('raven.events.Message',
                                            message=f'Error while setting input scw_list file from the frontend {e}')
             raise RequestNotUnderstood("Error while setting input scw_list from the frontend")
+        self.set_pars_from_dic(par_dic, verbose=verbose)
```
volodymyrss, line 189: Is this somthing that appeared to be necessary after the other PR? I am perplexed why it passed tests?

volodymyrss (changes requested):

cdci_data_analysis/analysis/instrument.py
burnout87: yes, good point, files could in the future be sent also from other sources

cdci_data_analysis/analysis/instrument.py
burnout87: because we never tested the range of certain input values directly within the dispatcher-app, but in the dispatcher-plugin-integral

eg [https://github.com/oda-hub/dispatcher-plugin-integral/blob/2889b5d7d3fa0feba7740dbf628e25abb1d74c56/tests/test_isgri_plugin.py#L425-L459](https://github.com/oda-hub/dispatcher-plugin-integral/blob/2889b5d7d3fa0feba7740dbf628e25abb1d74c56/tests/test_isgri_plugin.py#L425-L459)

volodymyrss: ok, can it be also tested directly in dispatcher-app?

burnout87: yes it should, it was never done before

tests/test_server_basic.py
```diff
@@ -716,6 +716,50 @@ def test_list_file(dispatcher_live_fixture):
     assert job_id == calculated_job_id
 
 
+def test_value_range(dispatcher_long_living_fixture):
```
volodymyrss, line 719: this is useful, but I do not really understand why is it part of this PR?

burnout87: It came out from our discussion and then I decided to put in this PR, just to avoid to create another one only for a single test

volodymyrss: ok, I did not check what test in the plugin breaks, if this is analogous to it, it's nice

burnout87: A range check for E1 value, so I reproduced here a similar one, after the change I applied.

volodymyrss (approved):

Issue #146, opened 2021-05-31
make sure message on 403 as shown in the frontend is informative


volodymyrss:
what is shown on frontend is noted here:

https://github.com/oda-hub/dispatcher-app/blob/master/cdci_data_analysis/flask_app/app.py#L232

We need to be sure dispatcher formats adequate messages in way that is shown in all cases, it's not always the case, with:

* token expired - need an advice
* priviledges insufficient
 
to check also that frontend-astrooda shows messages as needed in 403 status

volodymyrss:
what is confusing is that frontend session lifetime is not the same as dispatcher-connected token.
We may 

This is part of bigger challenge, that frontend is not guaranteed to have connection to dispatcher, and will currently return somewhat incomprehensible error (data server connection error) when can not reach it. 
Error message should provide some idea of what component is responsible and give instruction for action:
reload page, reconnect.

PR #191, opened 2021-07-01
validate response schema

volodymyrss:
if response schema is not maintained, frontend will not show expected messages

what is currently expected by frontend is specified in the schema with explanations in comments:

https://github.com/oda-hub/dispatcher-app/blob/master/cdci_data_analysis/flask_app/app.py#L234

The frontend assumptions on response are deduced from frontend-astrooda code.

this would help https://github.com/oda-hub/dispatcher-app/issues/146

### DIFF
--- a/cdci_data_analysis/flask_app/app.py
+++ b/cdci_data_analysis/flask_app/app.py
@@ -284,6 +284,8 @@ def run_analysis():
         logger.info("towards log_run_query_result")
         log_run_query_result(request_summary, r[0])
 
+        QueryOutJSON().load(r[0].json)
+
         return r
     except APIerror as e:
         raise
--- a/cdci_data_analysis/flask_app/app.py
+++ b/cdci_data_analysis/flask_app/app.py
@@ -28,6 +28,7 @@ import time as _time
 
 from .logstash import logstash_message
 from .schemas import QueryOutJSON
+from marshmallow.exceptions import ValidationError
 
 from ..plugins import importer
 
@@ -255,7 +256,6 @@ def run_analysis():
         logger.info("towards log_run_query_result")
         log_run_query_result(request_summary, r[0])
 
-        QueryOutJSON().load(r[0].json)
 
         return r
     except APIerror as e:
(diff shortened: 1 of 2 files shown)


cdci_data_analysis/flask_app/app.py
```diff
@@ -284,6 +284,8 @@ def run_analysis():
         logger.info("towards log_run_query_result")
         log_run_query_result(request_summary, r[0])
 
+        QueryOutJSON().load(r[0].json)
```
burnout87, line 287: Perhaps this is not complete? Since certain fields are not recognized?

volodymyrss: In principle, it does not need to specify all fields.
In fact, some fields can be optional.

volodymyrss: We should be aware that as this schema is derived from frontend, anything that does not fit it will not be shown correctly by frontend (and oda_api, which tries to follow it)

burnout87: But several tests are failing. Is there a way to "ignore" those unrecognized fields?

volodymyrss: At some point defaults for validation changed in marshmallow.
I made the necessary adaptations, all passes now.

But then it's even strange that all passes now. In fact, some fields are allowed by not used: e.g. error_message outside of exit status should be breaking validation.

We should eventually adapt schema to prevent these unused messages. I will make a simple attempt now.

volodymyrss: I made it so `error_message`, if present, must be empty. We should change the code where needed.

volodymyrss: suspiciously large number of them is failing. maybe it is expected. 
Could you please take over in verifying? 
If ok, we can close this one, and proceed with fixing in another one.

burnout87: yes thats ok

burnout87 (approved):

PR #197, opened 2021-07-01
Output schema for download functionalities 

burnout87:


burnout87:
Some fields of the `QueryOutJSON` schema were set with `required = True` which was causing several tests not to pass. Now all the fields are optional, but we can probably have those to a default value when not provided.

burnout87:
The changes I suggest here collide a bit with those implemented in #199 .

What do you think of these @volodymyrss ?

cdci_data_analysis/flask_app/schemas.py
```diff
@@ -21,16 +21,17 @@ class Meta:
     query_status = fields.Str(
                         validate=OneOf(["done", "failed", "submitted"]),
                         description="",
-                        required=True
+                        required=False,
```
volodymyrss, line 24: it's used by the clients in most cases, it must be required in those cases

cdci_data_analysis/flask_app/schemas.py
```diff
@@ -21,16 +21,17 @@ class Meta:
...
+                        missing=""
```
volodymyrss, line 25: this is not a valid value for this field, see OneOf above. It never makes sense for this to be ""

cdci_data_analysis/flask_app/schemas.py
```diff
@@ -21,16 +21,17 @@ class Meta:
...
     session_id = fields.Str(required=False) # is it required?
     job_id = fields.Str(required=False) # is it required?
 
     error_message = fields.Str(
-                        validate=OneOf([""]),
                         description="",
-                        required=False # but if present, should be empty
+                        # TODO why should this empty and in which case ?
+                        required=False, # but if present, should be empty
+                        missing=""
```
volodymyrss, line 36: this missing substitute value is never going to be used.
Also you removed the validation which requires no message to be send in this unused field.

volodymyrss (changes requested):
I think validation should be as strict as needed by the clients (hence explanations about how the fields are consumed by the clients in descriptions), but controlled by one or more switches (e.g. variables) to allow more loose validation.
Switches may be kept off in production and default tests at first.

volodymyrss:
> Some fields of the `QueryOutJSON` schema were set with `required = True` which was causing several tests not to pass. Now all the fields are optional, but we can probably have those to a default value when not provided.

they were set on purpose, that's the point of this activity: to find out which dispatcher responses (produced in some tests) do follow consistent schema, as needed by the clients.

cdci_data_analysis/flask_app/app.py
```diff
@@ -274,7 +274,8 @@ def run_analysis():
 @app.after_request
 def validate_schema(response):
     try:
-        QueryOutJSON().load(response.json)
+        if response.is_json:
```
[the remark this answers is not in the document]
burnout87: Exactly

cdci_data_analysis/flask_app/schemas.py
burnout87: This was one of the main reasons why the majority of the tests, (those were an error was expected) were not passing. Why should not this field be used?

cdci_data_analysis/flask_app/schemas.py
burnout87: Sure, then one of the possible values should be used, so a `query_status` is always returned

cdci_data_analysis/flask_app/schemas.py
volodymyrss: it should not be used because **it is not** used by the clients. Each field in schema is motivated by the client, as explained in the descriptions where they are. This approach should persist and expand.

cdci_data_analysis/flask_app/schemas.py
volodymyrss: it makes sense to return it when it is used by the clients. Is there really a meaningful default when no information on actual value is provided?
Instead, I would specialize QueryOut to cases when some things are expected and other when it is needed not to be. Gradually.

But to implement this "gradually" we can add switches, and remove/make-on-by-default them when it is sorted out.
https://github.com/oda-hub/dispatcher-app/pull/199

Or at least leave TODO comments in less important cases, like this one.

cdci_data_analysis/flask_app/schemas.py
volodymyrss: controlling this was the principal purpose of the switch I added in https://github.com/oda-hub/dispatcher-app/pull/199 and this field being used sometimes is probably the main cause of errors not shown correctly in the frontend - assuming that setting  this field has some value for the user at the moment is not useful.

burnout87:
@volodymyrss to handle the case of download, as well as js9 requests, it outputs now a default response while still enforcing the validation schema

cdci_data_analysis/flask_app/app.py
```diff
@@ -275,7 +275,11 @@ def run_analysis():
 def validate_schema(response):
     try:
         if dispatcher_strict_validate:
-            QueryOutJSON().load(response.json)
+            if response.is_streamed:
+                # TODO this should output a "default" response in case of download/js9 request
```
volodymyrss, line 279: you mean which output?

burnout87: It should create a default empty object, with default values, since no query output is generate by the dispatcher-app in case of download of files

volodymyrss: create where?

burnout87: It is just a default behavior in case no query output is produced by the dispatcher. We can also skip the validation for the specific case of download of files

volodymyrss: Well, this validation step here could validate more than QueryOutput, since all requests are passed through this method.
But testing download response by checking if {} matches QueryOutput is a bit misleading.

Indeed, I was suspecting some misunderstanding, since as of now, these default values are simply not used, and no default output is generated anywhere.

So it is better to ignore schema validation when there is  no schema.

burnout87: Then, for the moment, this whole PR is not necessary. 

Perhaps to mark this as a future improvement

cdci_data_analysis/flask_app/app.py
```diff
@@ -275,7 +275,9 @@ def run_analysis():
 def validate_schema(response):
     try:
         if dispatcher_strict_validate:
-            QueryOutJSON().load(response.json)
+            # TODO in case of download/js9 request a dedicated validaiton schema should be defined
+            if not response.is_streamed:
```
volodymyrss, line 279: separating-out the case of download file is a step for more clear response schema, why not

volodymyrss (approved):

PR #199, opened 2021-07-02
no validation by default

volodymyrss:
I should have disabled it by default when introducing the validation. Else really the dispatcher breaks.
Let's gradually make sure the validation passes, and then make it enforced by default.

### DIFF
--- a/cdci_data_analysis/flask_app/app.py
+++ b/cdci_data_analysis/flask_app/app.py
@@ -27,7 +27,7 @@ import logging
 import time as _time
 
 from .logstash import logstash_message
-from .schemas import QueryOutJSON
+from .schemas import QueryOutJSON, dispatcher_strict_validate
 from marshmallow.exceptions import ValidationError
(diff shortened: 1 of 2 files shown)


cdci_data_analysis/flask_app/app.py
```diff
@@ -274,7 +274,8 @@ def run_analysis():
 @app.after_request
 def validate_schema(response):
     try:
-        QueryOutJSON().load(response.json)
+        if dispatcher_strict_validate:
+            QueryOutJSON().load(response.json)
```
burnout87, line 278: As I tried in #197 , checking if the response contains a json would make it work for the `download` functionalities

volodymyrss: it would let through various useless html too. Instead, we should check if response contains files, or, even better, if it is response to download.

burnout87 (approved):

PR #187, opened 2021-06-23
Anticipated load from frontend

burnout87:
Loading of list files and catalog files from front-end is now performed before `job_id` calculation, and also before the scratch folder creation.

The data have therefore to be saved temporarily in another folder.

PR #203, opened 2021-07-07
fixed catalog read

burnout87:


burnout87:
Catalog that were selected from the frontend were not used because of the changes applied with #187 , this should fix the issue.

tests/test_server_basic.py
```diff
@@ -718,6 +718,95 @@ def test_list_file(dispatcher_live_fixture):
...
+    logger.info("constructed server: %s", server)
+
+    # let's generate a valid token
+    token_payload = {
+        **default_token_payload,
+        "roles": "unige-hpc-full, general",
+    }
+    encoded_token = jwt.encode(token_payload, secret_key, algorithm='HS256')
+
+    selected_catalog = """{"cat_column_descr":[["meta_ID","<i8"],["src_names","<U20"],["significance",">f4"],["ra",">f4"],["dec",">f4"],["NEW_SOURCE",">i2"],["ISGRI_FLAG","<i8"],["FLAG","<i8"],["ERR_RAD","<f8"]],"cat_column_list":[[1],["Test A"],[6.233789443969727],[299.81768798828125],[40.72419357299805],[-32768],[2],[0],[0.000029999999242136255]],"cat_column_names":["meta_ID","src_names","significance","ra","dec","NEW_SOURCE","ISGRI_FLAG","FLAG","ERR_RAD"],"cat_coord_units":"deg","cat_frame":"fk5","cat_lat_name":"dec","cat_lon_name":"ra"}"""
```
volodymyrss, line 778: could this line be shortened?

volodymyrss: is it certain that's what oda_api sends? I do not remember. There should be a test with oda_api somewhere too.

volodymyrss (approved):
Thanks, I could anticipate improvements, but maybe it's better for now just restore functionality.

See couple of notes though.

tests/test_server_basic.py
burnout87: I will provide a test also with oda_api

burnout87: > could this line be shortened?

Perhaps, but I don't know much regarding the way catalogs are formatted. Is there a place I can look into to understand more?

volodymyrss: > > could this line be shortened?
> 
> Perhaps, but I don't know much regarding the way catalogs are formatted. Is there a place I can look into to understand more?

you can look into how oda_api sends it.

tests/test_server_basic.py
```diff
@@ -718,6 +718,165 @@ def test_list_file(dispatcher_live_fixture):
...
+    logger.info("constructed server: %s", server)
+
+    # let's generate a valid token
+    token_payload = {
+        **default_token_payload,
+        "roles": "unige-hpc-full, general",
+    }
+    encoded_token = jwt.encode(token_payload, secret_key, algorithm='HS256')
+
+    selected_catalog_dict = dict(
```
volodymyrss, line 778: just FYI, this is not exactly how we would create these catalogs, since it's a bit verbose this way. 
see examples in this [repository](https://github.com/oda-hub/oda_api_benchmark/blob/master/examples/V404_Cyg_analysis.ipynb)

however, directly constructing as you did has an advantage of expressing the format in here.

volodymyrss (approved):

tests/test_server_basic.py
burnout87: thanks

A good improvement for the future is to have a dedicated fixture that would build a catalog in the most appropriate way

volodymyrss: maybe at some point it would be even useful  for users to construct the catalog more intuitively.

PR #204, opened 2021-07-07
check scw_list size for email

burnout87:


cdci_data_analysis/analysis/email_helper.py
```diff
@@ -166,9 +177,14 @@ def send_email(
...
     #   compressed_request_url = compress_request_url_params(request_url)
-    
 
-    if len(request_url) > 600:
+    url_warning_message = ""
+
+    if len(request_url) > 2000:
+        possibly_compressed_request_url = ""
+        permanent_url = False
+        url_warning_message = "A URL to inspect the status of the job could not be generated."
```
volodymyrss, line 186: this is not particularily instructive for the user. What can use do with this information?
I suggest something like that:
"
You can retrieve the results by repeating the request.
Unfortunately, due to a known issue with very large requests, URL automatically  the selecting request parameters could not be generated. This will be fixed in a future release."

volodymyrss (changes requested):

cdci_data_analysis/analysis/email_helper.py
burnout87: Thanks, this was indeed a very first version of the information for the user, to be extended for sure

volodymyrss (approved):

PR #207, opened 2021-07-22
to suppress email submission completely with token config

volodymyrss:
I will generate the token elsewhere, but I need this suppression for larger analysises I run.

### DIFF
--- a/cdci_data_analysis/analysis/tokenHelper.py
+++ b/cdci_data_analysis/analysis/tokenHelper.py
@@ -56,23 +56,3 @@ def get_decoded_token(token, secret_key):
     # decode the encoded token
     if token is not None:
         return jwt.decode(token, secret_key, algorithms=[default_algorithm])
-
-def update_token(token, secret_key, payload_mutation: FunctionType):
-    token_payload = jwt.decode(token, secret_key, algorithms=[default_algorithm])    
-
-    # update here
-    # tem=300000
-
-    token_payload = payload_mutation(token_payload) 
-
-    out_token = jwt.encode(token_payload, secret_key, algorithm=default_algorithm)
-
-    return out_token
-
-def update_token_suppress_email(token, secret_key):
-    def payload_mutation(token_payload):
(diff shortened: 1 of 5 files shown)


tests/test_server_basic.py
```diff
@@ -473,6 +473,10 @@ def test_email_oda_api(dispatcher_live_fixture, dispatcher_local_mail_server):
         **default_token_payload
     }
     encoded_token = jwt.encode(token_payload, secret_key, algorithm='HS256')
+    
+    if isinstance(encoded_token, bytes):
```
burnout87, line 477: Was this case happening ?

cdci_data_analysis/analysis/tokenHelper.py
```diff
@@ -42,11 +44,35 @@ def get_token_user_sending_submitted_interval_email(decoded_token):
...
\ No newline at end of file
+        return jwt.decode(token, secret_key, algorithms=[default_algorithm])
+
+def update_token(token, secret_key, payload_mutation: FunctionType):
+    token_payload = jwt.decode(token, secret_key, algorithms=[default_algorithm])    
+
+    # update here
+    # tem=300000
+
+    token_payload = payload_mutation(token_payload)
```
burnout87, line 66: Is the goal of the `payload_mutation` to just change the content of the token payload? And sometimes also to return it?

burnout87 (approved):
Just a couple of quetions

tests/test_server_basic.py
volodymyrss: yes, that's why I added it. It depends on pyjwt version I guess.

cdci_data_analysis/analysis/tokenHelper.py
volodymyrss: actually, this part it's not finished. And I did also some token inspection/modification in oda_api:
https://github.com/oda-hub/oda_api/pull/56/files which starts to duplicate this. So I remove, and think later where to put it better.

PR #214, opened 2021-08-06
Test empty image

burnout87:


burnout87:
I tried to create a dummy Numpy product for the `empty` instrument, but I am not entirely sure I did in the correct way, hence the tag.

This will be helpful for dealing with [https://github.com/oda-hub/oda_api/issues/60](https://github.com/oda-hub/oda_api/issues/60)

tests/test_server_basic.py
```diff
@@ -1009,3 +1009,45 @@ def test_example_config(dispatcher_test_conf):
     print("\n\n\ntest_config_keys", test_config_keys)
 
     assert set(example_config_keys) == set(test_config_keys)
+
+
+def test_mosaic_image(dispatcher_live_fixture):
```
volodymyrss, line 1014: it's a bit strange that "mosaic" is mentioned in generic dispatcher. It has concept of Image but not details like mosaic. But it's minor.

burnout87: I am not too familiar with these concepts and the relative terminology.

So just using image would be enough? Whreas mosaic is a more instrument-specific terminology?

volodymyrss: It depends. But what is relevant here is what is already present in dispatcher. You can see that there is `ImageQuery` and `ImageProduct` but no mosaics. It's a reasonable choice to introduce image but not mosaic.

volodymyrss:
It's related to some issue, yes?

burnout87:
> It's related to some issue, yes?

In order to tackle what we discussed here [https://github.com/oda-hub/oda_api/issues/60](https://github.com/oda-hub/oda_api/issues/60), I wanted to have one of the empty/dummy instruments to provide as a product an image.

volodymyrss (approved):

PR #217, opened 2021-08-16
Performance

volodymyrss:


volodymyrss:
This is critical since this resolves delay in post-processing which otherwise makes some requests so slow that they almost never succeed, and usually time-out.

cdci_data_analysis/analysis/products.py
```diff
@@ -334,35 +313,42 @@ def write(self, file_name=None, overwrite=True, file_dir=None):
...
-
-    def add_url_to_fits_file(self,par_dict,url='',use_primary=True,add_query_dict=True):
+        # turns out the way astropy writes some structures really overloads some nfs mounts.
+        # this happened on UNIGE site after update. 
+        # workaround is made here and should be configurable
+        logger.debug("\033[31mbefore %s.write(file_name=%s) as %s\033[0m", self, file_name, file_path)
+
+        with tempfile.NamedTemporaryFile(delete=False) as f:
+            self.data.write_fits_file(f, overwrite=overwrite)
+            f.close()
```
burnout87, line 323: since it's inside a `with` block, this should not be necessary...

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -87,21 +87,41 @@ def instrument_name(self):
...
+        return self._query_progression
+
+    def log_query_progression(self, message):        
+        self.query_progression.append(dict(
+                t_s = time_.time(),
+                message = message,                
+            ))
+
+        t0 = self.query_progression[0]['t_s']
+        self.logger.warning("%s %s s", message, self.query_progression[-1]['t_s'] - t0)
```
burnout87, line 104: Is this intended just for being used inside the `InstrumentQueryBackend` class?

burnout87 (commented):
Just a couple of questions...

cdci_data_analysis/analysis/products.py
volodymyrss: > since it's inside a `with` block, this should not be necessary...

see next time, I need it to be closed before moving.

burnout87: oh yes, didn't properly see that

volodymyrss: maybe you are comparing with `open` context which works differently. 
Here really we do not really need context exit, but that's one of the nicer ways  for `tempfile` to make a named file.

cdci_data_analysis/flask_app/dispatcher_query.py
volodymyrss: yes, and it is used.

burnout87 (approved):

cdci_data_analysis/analysis/products.py
burnout87: indeed, I thought it would basically work in the same way

PR #222, opened 2021-08-20
Long scw list email

volodymyrss:


burnout87:
While inspecting the code and running some other tests, I encountered the following case for the api_code after being wrapped by the `black` library.

Before being wrapped:

```
...
    "token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJtdG1AbXRtY28ubmV0IiwibmFtZSI6Im1tZWhhcmdhIiwicm9sZXMiOiJnZW5lcmFsIiwiZXhwIjoxNjI5NzE4OTAwLCJ0ZW0iOjAsIm1zdG91dCI6dHJ1ZSwibXNzdWIiOnRydWUsImludHN1YiI6NX0.Q14ZbVOIf8H9vpqxRYwjMHUA0tS-aiF5iw2gjrhadxI",
...
```

after applying the wrapping:

```
...
    "token": (
        "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJtdG1AbXRtY28ubmV0IiwibmFtZSI6Im1tZWhhcmdhIiwicm9sZX"
        "MiOiJnZW5lcmFsIiwiZXhwIjoxNjI5NzE4OTAwLCJ0ZW0iOjAsIm1zdG91dCI6dHJ1ZSwibXNzdWIiOnRydWUsImludHN1YiI6N"
        "X0.Q14ZbVOIf8H9vpqxRYwjMHUA0tS-aiF5iw2gjrhadxI"
    ),
...
```

Is it ok to have those parentheses in the token?

The consequence of this is that the token itself is not then replaced with the replacement string: 
[https://github.com/oda-hub/dispatcher-app/blob/8a4acb2d9b0620edff814b52a78d0d51cf0a806a/cdci_data_analysis/analysis/email_helper.py#L180](https://github.com/oda-hub/dispatcher-app/blob/8a4acb2d9b0620edff814b52a78d0d51cf0a806a/cdci_data_analysis/analysis/email_helper.py#L180)


Is this an expected behavior?

volodymyrss:
> While inspecting the code and running some other tests, I encountered the following case for the api_code after being wrapped by the `black` library.
> 
> Before being wrapped:
> 
> ```
> ...
>     "token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJtdG1AbXRtY28ubmV0IiwibmFtZSI6Im1tZWhhcmdhIiwicm9sZXMiOiJnZW5lcmFsIiwiZXhwIjoxNjI5NzE4OTAwLCJ0ZW0iOjAsIm1zdG91dCI6dHJ1ZSwibXNzdWIiOnRydWUsImludHN1YiI6NX0.Q14ZbVOIf8H9vpqxRYwjMHUA0tS-aiF5iw2gjrhadxI",
> ...
> ```
> 
> after applying the wrapping:
> 
> ```
> ...
>     "token": (
>         "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJtdG1AbXRtY28ubmV0IiwibmFtZSI6Im1tZWhhcmdhIiwicm9sZX"
>         "MiOiJnZW5lcmFsIiwiZXhwIjoxNjI5NzE4OTAwLCJ0ZW0iOjAsIm1zdG91dCI6dHJ1ZSwibXNzdWIiOnRydWUsImludHN1YiI6N"
>         "X0.Q14ZbVOIf8H9vpqxRYwjMHUA0tS-aiF5iw2gjrhadxI"
>     ),
> ...
> ```
> 
> Is it ok to have those parentheses in the token?

there are no parentheses in the token, that's how long string can be written in python. 
Arguably, that's [the best way](https://stackoverflow.com/questions/1874592/how-to-write-very-long-string-that-conforms-with-pep8-and-prevent-e501) to do it. Black is following this approach for a reason.

> 
> The consequence of this is that the token itself is not then replaced with the replacement string, thus having the token inside the email:
> 
> https://github.com/oda-hub/dispatcher-app/blob/8a4acb2d9b0620edff814b52a78d0d51cf0a806a/cdci_data_analysis/analysis/email_helper.py#L180

There is *always* token inside the email.    As you an see, `api_code_no_token` is not currently used in the template. Maybe we wanted to keep it for the future, it's enough to move it before wrapping.

> 
> Is this an expected behavior?

Absolutely. Token has to be broken in multiple lines, and the best way to do it is to use brackets.

burnout87:
> 
> there are no parentheses in the token, that's how long string can be written in python.
> Arguably, that's [the best way](https://stackoverflow.com/questions/1874592/how-to-write-very-long-string-that-conforms-with-pep8-and-prevent-e501) to do it. Black is following this approach for a reason.
> 

The parentheses I am referring to are not inside the token, just _around_ the token itself.

> > The consequence of this is that the token itself is not then replaced with the replacement string, thus having the token inside the email:
> > https://github.com/oda-hub/dispatcher-app/blob/8a4acb2d9b0620edff814b52a78d0d51cf0a806a/cdci_data_analysis/analysis/email_helper.py#L180
> 
> There is _always_ token inside the email. As you an see, `api_code_no_token` is not currently used in the template. Maybe we wanted to keep it for the future, it's enough to move it before wrapping.
> 
Yes I noticed that. So it will be used at some point in the future, and adjusted if needed.

> > Is this an expected behavior?
> 
> Absolutely. Token has to be broken in multiple lines, and the best way to do it is to use brackets.

burnout87 (approved):

volodymyrss:
> > there are no parentheses in the token, that's how long string can be written in python.
> > Arguably, that's [the best way](https://stackoverflow.com/questions/1874592/how-to-write-very-long-string-that-conforms-with-pep8-and-prevent-e501) to do it. Black is following this approach for a reason.
> 
> The parentheses I am referring to are not inside the token, just _around_ the token itself.

ah ok, because you said `parentheses in the token`.

Around the token they are not only ok, they are necessary to break the line, which is of course the whole point here.

> 
> > > The consequence of this is that the token itself is not then replaced with the replacement string, thus having the token inside the email:
> > > https://github.com/oda-hub/dispatcher-app/blob/8a4acb2d9b0620edff814b52a78d0d51cf0a806a/cdci_data_analysis/analysis/email_helper.py#L180
> > 
> > 
> > There is _always_ token inside the email. As you an see, `api_code_no_token` is not currently used in the template. Maybe we wanted to keep it for the future, it's enough to move it before wrapping.
> 
> Yes I noticed that. So it will be used at some point in the future, and adjusted if needed.
> 
It might be used. Or might be not. We did not enable it since it's quite useful to have the token in the email directly. It makes the email more sensitive, but that's perhaps the right amount of sensitivity for this email.

PR #223, opened 2021-08-20
`use_scws` parameter reinserted for frontend

burnout87:
Addresses [https://github.com/oda-hub/doc-review-guidelines/issues/57#issuecomment-902516130](https://github.com/oda-hub/doc-review-guidelines/issues/57#issuecomment-902516130)

### DIFF
--- a/cdci_data_analysis/flask_app/dispatcher_query.py
+++ b/cdci_data_analysis/flask_app/dispatcher_query.py
@@ -404,12 +404,15 @@ class InstrumentQueryBackEnd:
         self.use_scws = self.par_dic.pop('use_scws', None)
         #
         if 'scw_list' in self.par_dic.keys():
+            if self.use_scws == 'no':
+                raise RequestNotUnderstood("scw_list parameter was provided "
+                                           "despite use_scws was indicating this was not provided, "
+                                           "please check the inputs you provided")
             _p = request.values.getlist('scw_list')
             if len(_p) > 1:
                 self.par_dic['scw_list'] = _p
             # use_scws should be set for, if any, url link within the email
             if self.use_scws is None:
-                # TODO ideally should be configurable, since the frontend might change behavior (though unlikely)
                 self.use_scws = 'form_list'
             print('=======> scw_list',  self.par_dic['scw_list'], _p, len(_p))
         else:
--- a/cdci_data_analysis/flask_app/dispatcher_query.py
+++ b/cdci_data_analysis/flask_app/dispatcher_query.py
@@ -104,7 +104,6 @@ class InstrumentQueryBackEnd:
         t0 = self.query_progression[0]['t_s']
         self.logger.warning("%s %s s", message, self.query_progression[-1]['t_s'] - t0)
 
-
     def __init__(self, app, instrument_name=None, par_dic=None, config=None, data_server_call_back=False, verbose=False, get_meta_data=False, download_products=False, resolve_job_url=False, query_id=None, update_token=False):
         self.logger = logging.getLogger(f"{repr(self)} [{query_id}]")
         self.logger = logging.getLogger(repr(self))
@@ -113,7 +112,6 @@ class InstrumentQueryBackEnd:
             self.logger.setLevel(logging.DEBUG)
         else:
             self.logger.setLevel(logging.INFO)
-        
         
         self.app = app
         try:
@@ -404,8 +402,9 @@ class InstrumentQueryBackEnd:
 
         self.par_dic = args.to_dict()
 
-        # if passed, disregard it, since it is nowhere necessary within the dispatcher-app
-        self.par_dic.pop('use_scws', None)
+        # if passed, disregard it, since it is nowhere necessary within the dispatcher-app,
+        # but re-attach it to the url within the email when sending it
+        self.use_scws = self.par_dic.pop('use_scws', None)
 
         if 'scw_list' in self.par_dic.keys():
             _p = request.values.getlist('scw_list')
@@ -728,7 +727,10 @@ class InstrumentQueryBackEnd:
                 except KeyError as e:
                     raise MissingRequestParameter(repr(e))
                 # build the products URL and get also the original requested product
-                products_url = self.generate_products_url_from_file(self.config.products_url, request_par_dict=original_request_par_dic)
+                products_url = self.generate_products_url_from_file(self.config.products_url,
+                                                                    request_par_dict=original_request_par_dic,
+                                                                    include_use_scws=True)
+
                 email_helper.send_email(
                     config=self.app.config['conf'],
                     logger=self.logger,
@@ -809,10 +811,13 @@ class InstrumentQueryBackEnd:
         else:
             return json.load(open(scratch_dir_parameters[0]))
 
-
     # TODO make sure that the list of parameters to ignore in the frontend is synchronized
-    def generate_products_url_from_par_dict(self, products_url, par_dict) -> str:
+    def generate_products_url_from_par_dict(self, products_url, par_dict, include_use_scws=False) -> str:
         par_dict = par_dict.copy()
+        if include_use_scws:
+            # for the frontend
+            par_dict['use_scws'] = self.use_scws
+
         _skip_list_ = ['token', 'session_id', 'job_id']
 
         for key, value in dict(par_dict).items():
@@ -822,9 +827,8 @@ class InstrumentQueryBackEnd:
         request_url = '%s?%s' % (products_url, urlencode(par_dict))
         return request_url
 
-    def generate_products_url_from_file(self, products_url, request_par_dict) -> str:
-        return self.generate_products_url_from_par_dict(products_url, request_par_dict)
-
+    def generate_products_url_from_file(self, products_url, request_par_dict, include_use_scws=False) -> str:
+        return self.generate_products_url_from_par_dict(products_url, request_par_dict, include_use_scws=include_use_scws)
 
     def run_query_mock(self, off_line=False):
 
@@ -1488,7 +1492,9 @@ class InstrumentQueryBackEnd:
                                                                self.app.config['conf'],
                                                                decoded_token=self.decoded_token):
                         try:
-                            products_url = self.generate_products_url_from_par_dict(self.app.config.get('conf').products_url, self.par_dic)
+                            products_url = self.generate_products_url_from_par_dict(self.app.config.get('conf').products_url,
+                                                                                    self.par_dic,
+                                                                                    include_use_scws=True)
 
                             email_helper.send_email(
                                 config=self.app.config['conf'],
--- a/cdci_data_analysis/flask_app/dispatcher_query.py
+++ b/cdci_data_analysis/flask_app/dispatcher_query.py
@@ -411,6 +411,11 @@ class InstrumentQueryBackEnd:
             _p = request.values.getlist('scw_list')
             if len(_p) > 1:
                 self.par_dic['scw_list'] = _p
+            # it could be a comma-separated string, so let's convert to a list
+            elif len(_p) == 1:
+                _p = str(_p)[1:-1].replace('\'','').split(",")
+                if len(_p) > 1:
+                    self.par_dic['scw_list'] = _p
             # use_scws should be set for, if any, url link within the email
             if self.use_scws is None:
                 self.use_scws = 'form_list'
@@ -840,6 +845,10 @@ class InstrumentQueryBackEnd:
                 self.use_scws = 'form_list'
             # for the frontend
             par_dict['use_scws'] = self.use_scws
+            # setting proper scw_list formatting
+            # str_scw_list = str(par_dict['scw_list'])
+            # str_scw_list = par_dict['scw_list'].split(",")
+            # str_scw_list = [item.strip() for item in str_scw_list]
 
         _skip_list_ = ['token', 'session_id', 'job_id']
 
--- a/cdci_data_analysis/flask_app/dispatcher_query.py
+++ b/cdci_data_analysis/flask_app/dispatcher_query.py
@@ -412,6 +412,8 @@ class InstrumentQueryBackEnd:
                 self.use_scws = 'form_list'
             print('=======> scw_list',  self.par_dic['scw_list'], _p, len(_p))
         else:
+            # not necessary to check the case of scw_list passed via file,
+            # since it is verified at a later stage
             if self.use_scws is not None and self.use_scws == 'form_list':
                 raise RequestNotUnderstood("scw_list parameter was expected to be passed, but it has not been found")
 
--- a/cdci_data_analysis/flask_app/dispatcher_query.py
+++ b/cdci_data_analysis/flask_app/dispatcher_query.py
@@ -829,7 +829,7 @@ class InstrumentQueryBackEnd:
     def generate_products_url_from_par_dict(self, products_url, par_dict) -> str:
         par_dict = par_dict.copy()
 
-        if 'scw_list' in par_dict and self.use_scws is not None:
+        if ('scw_list' in par_dict and self.use_scws is not None) or self.use_scws == 'no':
             # for the frontend
             par_dict['use_scws'] = self.use_scws
 
--- a/cdci_data_analysis/flask_app/dispatcher_query.py
+++ b/cdci_data_analysis/flask_app/dispatcher_query.py
@@ -403,7 +403,7 @@ class InstrumentQueryBackEnd:
         self.par_dic = args.to_dict()
 
         # if passed, disregard it, since it is nowhere necessary within the dispatcher-app,
-        # but re-attach it to the url within the email when sending it
+        # but re-attach it to the url within the email when sending it since it is used by the frontend
         self.use_scws = self.par_dic.pop('use_scws', None)
 
         if 'scw_list' in self.par_dic.keys():
@@ -728,8 +728,7 @@ class InstrumentQueryBackEnd:
                     raise MissingRequestParameter(repr(e))
                 # build the products URL and get also the original requested product
                 products_url = self.generate_products_url_from_file(self.config.products_url,
-                                                                    request_par_dict=original_request_par_dic,
-                                                                    include_use_scws=True)
+                                                                    request_par_dict=original_request_par_dic)
 
                 email_helper.send_email(
                     config=self.app.config['conf'],
@@ -812,9 +811,9 @@ class InstrumentQueryBackEnd:
             return json.load(open(scratch_dir_parameters[0]))
 
     # TODO make sure that the list of parameters to ignore in the frontend is synchronized
-    def generate_products_url_from_par_dict(self, products_url, par_dict, include_use_scws=False) -> str:
+    def generate_products_url_from_par_dict(self, products_url, par_dict) -> str:
         par_dict = par_dict.copy()
-        if include_use_scws:
+        if 'scw_list' in par_dict and self.use_scws is not None:
             # for the frontend
             par_dict['use_scws'] = self.use_scws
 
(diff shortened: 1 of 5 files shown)


cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -404,8 +402,9 @@ def set_args(self, request, verbose=False):
 
         self.par_dic = args.to_dict()
 
-        # if passed, disregard it, since it is nowhere necessary within the dispatcher-app
-        self.par_dic.pop('use_scws', None)
+        # if passed, disregard it, since it is nowhere necessary within the dispatcher-app,
+        # but re-attach it to the url within the email when sending it
```
volodymyrss, line 406: *since it is used by frontend

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -809,10 +811,13 @@ def find_job_id_parameters(self, job_id):
         else:
             return json.load(open(scratch_dir_parameters[0]))
 
-
     # TODO make sure that the list of parameters to ignore in the frontend is synchronized
-    def generate_products_url_from_par_dict(self, products_url, par_dict) -> str:
+    def generate_products_url_from_par_dict(self, products_url, par_dict, include_use_scws=False) -> str:
         par_dict = par_dict.copy()
+        if include_use_scws:
```
volodymyrss, line 817: is it possible to determine if `include_use_scws` within this function, from the parameter list. If it is needed or not can be derived unambigiously, by convention: if scw list is provided, it is used. I think. That's what is provided to the API.

tests/test_job_management.py
```diff
@@ -280,6 +279,10 @@ def get_expected_products_url(dict_param,
     dict_param_complete.pop("session_id", None)
     dict_param_complete.pop("job_id", None)
 
+    # to adapt to the dispatcher-app, and have it at the end of the url
```
volodymyrss, line 282: it's important somewhere that it is at the end?

volodymyrss (changes requested):

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -404,8 +402,9 @@ def set_args(self, request, verbose=False):
 
         self.par_dic = args.to_dict()
 
-        # if passed, disregard it, since it is nowhere necessary within the dispatcher-app
-        self.par_dic.pop('use_scws', None)
+        # if passed, disregard it, since it is nowhere necessary within the dispatcher-app,
+        # but re-attach it to the url within the email when sending it
+        self.use_scws = self.par_dic.pop('use_scws', None)
```
volodymyrss, line 407: maybe entire logic of determining if we `use_scws` can be kept in a separate method, called before this pop.
It would use combination of frontend-way and api-way to derive use_scws.

Then, product URLs are always made for frontend and should include `use_scws`.

API codes do not.

tests/test_job_management.py
burnout87: No, but the test would fail because the parameters are ordered differently from the way the dispatcher handles the `use_scws` parameter, ie added at the end of the dictionary, during the production of the url for the email.

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -396,6 +396,18 @@ def set_sentry_client(self, sentry_url=None):
     def get_current_ip(self):
         return socket.gethostbyname(socket.gethostname())
 
+    def set_scws_realted_params(self):
+        # it is nowhere necessary within the dispatcher-app,
+        # but it is re-attached to the url within the email
+        # when sending it since it is used by the frontend
+        self.use_scws = self.par_dic.pop('use_scws', None)
+        #
+        if 'scw_list' in self.par_dic.keys():
```
volodymyrss, line 405: I think if `scw_list` is in params, and the query is from api, the eventual URL to be parsed by the frontend also has to contain `use_scws`. Else url will not work in the frontend.

Could you verify this?

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -396,6 +396,18 @@ def set_sentry_client(self, sentry_url=None):
     def get_current_ip(self):
         return socket.gethostbyname(socket.gethostname())
 
+    def set_scws_realted_params(self):
+        # it is nowhere necessary within the dispatcher-app,
+        # but it is re-attached to the url within the email
+        # when sending it since it is used by the frontend
+        self.use_scws = self.par_dic.pop('use_scws', None)
```
volodymyrss, line 403: also if `use_scws` is set in parameters, but no actual scw list is given, issue should be raised.

tests/test_job_management.py
volodymyrss: in general, python dictionaries were not really ordered until ~3.6. Even now they are ordered a bit unexpectedly (for performance). query parameter ordering is also sometimesambigious. Maybe it's better to avoid relying on the order and code..

volodymyrss: so maybe we should always order parameters in the url. To reduce ambiguity.

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: Ah true! It makes sense. 

So it should be added anyway when creating the url for the email.

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -809,10 +814,13 @@ def find_job_id_parameters(self, job_id):
         else:
             return json.load(open(scratch_dir_parameters[0]))
 
-
     # TODO make sure that the list of parameters to ignore in the frontend is synchronized
     def generate_products_url_from_par_dict(self, products_url, par_dict) -> str:
         par_dict = par_dict.copy()
+        if 'scw_list' in par_dict and self.use_scws is not None:
```
volodymyrss, line 820: actually, if `self.scw_list` is already set by the function above, maybe it can be used here?

burnout87: `self.scw_list` does not actually exist, or am I wrong?

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: And the value that it should have is `form_list`, so that the right radio button will be selected.

tests/test_job_management.py
burnout87: I had to sort the keys of the dictionary that is then used to create the url

volodymyrss: > I had to sort the keys of the dictionary that is then used to create the url

It's good in any case, then URL is more deterministic. Same request will be expressed by the same URL.

tests/test_job_management.py
```diff
@@ -1123,6 +1134,55 @@ def test_email_very_long_request_url(dispatcher_long_living_fixture, dispatcher_
 This will be fixed in a future release.""" in email_data
 
 
+def test_email_scws_list(dispatcher_live_fixture,
```
volodymyrss, line 1137: oh, could we use this test to make clear our (argumably, somewhat unnecessarily convoluted) logic, in all of the following cases:

* `use_scws` NOT set and `scw_list` set (API and scw list result)
* `use_scws` NOT set and `scw_list` NOT set (API and time range result)
* `use_scws` in `form_list`, `user_file` and `scw_list` set (frontend, scw list result) Not sure about this one.
* `use_scws`==`no` (frontend, time range result)
* `use_scws`!=`no` and `scw_list` NOT set (frontend, exception)

Sorry to stress on this one, but it's a thing which unfortunately caused confusion in the past as well as at this late development stage.

Net result in all cases should be that products URL contains `use_scws` set correspondingly to the product kind, and API code never contains `use_scws` but contains `scw_list` ONLY if scw_list is used.

ALTERNATIVELY, better, would to make API behave exactly like frontend. This would introduce a breaking change in oda_api, meaning that new dispatcher would be only allowed with new oda_api. We could do this, if we make dispatcher refuse older API versions.
Maybe it's too much organization to do this. We can discuss tomorrow too.

I would suggest, for now, to complete the test.

volodymyrss (changes requested):

burnout87:
I think I have now covered all the cases we discussed, with the various cases for `use_scws`. There are some tricky cases to handle, hopefully I manage to cover them all.

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -809,23 +825,30 @@ def find_job_id_parameters(self, job_id):
...
             return json.load(open(scratch_dir_parameters[0]))
 
-
     # TODO make sure that the list of parameters to ignore in the frontend is synchronized
     def generate_products_url_from_par_dict(self, products_url, par_dict) -> str:
         par_dict = par_dict.copy()
+
+        if ('scw_list' in par_dict and self.use_scws is not None) or self.use_scws == 'no':
+            # for the frontend
+            par_dict['use_scws'] = self.use_scws
```
volodymyrss, line 834: will this not create cases when `use_scws` is not set in the url?
I guess it does not matter sometimes, but it does not hurt to have it always set for the frontend.

cdci_data_analysis/pytest_fixtures.py
```diff
@@ -702,6 +721,13 @@ def from_run_analysis_response(cls, r):
             session_id = r.json()['session_id'],
             job_id = r.json()['job_monitor']['job_id']
         )
+
+    @classmethod
```
volodymyrss, line 725: this largely duplicates previous method.
if you add this, the previous one may be hust `from_run_analysis_response(r.json())`

tests/test_job_management.py
```diff
@@ -1162,6 +1168,127 @@ def test_email_very_long_request_url(dispatcher_long_living_fixture, dispatcher_
...
+        "roles": "unige-hpc-full, general",
+    }
+    encoded_token = jwt.encode(token_payload, secret_key, algorithm='HS256')
+
+    params = {
+        'query_status': "new",
+        'product_type': "dummy",
+        'query_type': "Real",
+        'instrument': 'empty-async',
+        'use_scws': use_scws_value,
```
volodymyrss, line 1195: is there also a case of not passing use_scws? Not really the same as None.

tests/test_job_management.py
```diff
@@ -1162,6 +1168,127 @@ def test_email_very_long_request_url(dispatcher_long_living_fixture, dispatcher_
...
+        # no list should be passed
+        jdata = ask(server,
+                    params,
+                    max_time_s=150,
+                    expected_query_status=None,
+                    expected_status_code=None
+                    )
+    else:
+        if passing_scw_list:
+            params['scw_list'] = [f"0665{i:04d}0010.001" for i in range(5)]
```
volodymyrss, line 1224: oh, by the way, we experienced another variation: sometimes it's a list, and sometimes it's coma-separated string. Good to test this here!

tests/test_job_management.py
```diff
@@ -1162,6 +1168,127 @@ def test_email_very_long_request_url(dispatcher_long_living_fixture, dispatcher_
...
+                    )
+    elif use_scws_value == 'no':
+        # no list should be passed
+        jdata = ask(server,
+                    params,
+                    max_time_s=150,
+                    expected_query_status=None,
+                    expected_status_code=None
+                    )
+    else:
```
volodymyrss, line 1222: and `None` case?

tests/test_job_management.py
```diff
@@ -1162,6 +1168,127 @@ def test_email_very_long_request_url(dispatcher_long_living_fixture, dispatcher_
...
+        'product_type': "dummy",
+        'query_type': "Real",
+        'instrument': 'empty-async',
+        'use_scws': use_scws_value,
+        'token': encoded_token
+    }
+
+    if use_scws_value == 'user_file':
+        scw_list_file_obj = None
+        if passing_scw_list:
```
volodymyrss, line 1201: this condition is repeated 3 times. Can it not be layer above the other `if`?

tests/test_job_management.py
```diff
@@ -1162,6 +1168,127 @@ def test_email_very_long_request_url(dispatcher_long_living_fixture, dispatcher_
...
+                    params,
+                    method='post',
+                    max_time_s=150,
+                    expected_query_status=None,
+                    expected_status_code=None,
+                    files=scw_list_file_obj
+                    )
+    elif use_scws_value == 'no':
+        # no list should be passed
+        jdata = ask(server,
```
volodymyrss, line 1216: this does not test the case when `use_scws_value == "no"` but something is passed. Similarly in other cases

volodymyrss (changes requested):

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: In case `use_scws` is not set:

- If `scw_list` is set, then `use_scws` is set to automatically `form_list`
- If `scw_list` is not set, then `use_scws` remains not set, since it won't be needed

Or am I miss some case?

tests/test_job_management.py
burnout87: Ok! I will include

burnout87: Is that a case for the api?

tests/test_job_management.py
burnout87: What is missing in the parameters for the request?

volodymyrss:
> Is that a case only for the api? 

Please verify. Normally frontend should forbid this.

tests/test_job_management.py
volodymyrss: I mean when `use_scws_value == "no"` but scw_list is passed.

burnout87: ok, and in that case what behavior do we expect? consider the `scw_list` passed and change the value of `use_scws` to `form_list` so to use the passed list in the frontend?

volodymyrss: Normally, such a request can not reasonably emerge in frontend or API, and if it is sent, it has to be considered as self-contradictory, ambiguous, and hence return some 400 error.

It might be that it is sent by frontend at some point. But let's make it 400 and leave a comment note

burnout87:
> > Is that a case only for the api?
> 
> Please verify. Normally frontend should forbid this.

frontend does not indeed allow this, but I would include also a case where the list is specified as string in a file. Does it make sense to consider also this case?

volodymyrss:
> > > Is that a case only for the api?
> > 
> > 
> > Please verify. Normally frontend should forbid this.
> 
> frontend does not indeed allow this, but I would include also a case where the list is specified as string in a file. Does it make sense to consider also this case?

List as string? In what form?
Currently, in the file, it is specified as new-line-separated string with comas.
Please verify, deduce from`set_input_products` what it accepts and test that.

burnout87:
> > > > Is that a case only for the api?
> > > 
> > > 
> > > Please verify. Normally frontend should forbid this.
> > 
> > 
> > frontend does not indeed allow this, but I would include also a case where the list is specified as string in a file. Does it make sense to consider also this case?
> 
> List as string? In what form?
> Currently, in the file, it is specified as new-line-separated string with comas.
> Please verify, deduce from`set_input_products` what it accepts and test that.

Like described here:
[https://github.com/oda-hub/dispatcher-app/pull/223#discussion_r694568588](https://github.com/oda-hub/dispatcher-app/pull/223#discussion_r694568588)

volodymyrss:
> > > > > Is that a case only for the api?
> > > > 
> > > > 
> > > > Please verify. Normally frontend should forbid this.
> > > 
> > > 
> > > frontend does not indeed allow this, but I would include also a case where the list is specified as string in a file. Does it make sense to consider also this case?
> > 
> > 
> > List as string? In what form?
> > Currently, in the file, it is specified as new-line-separated string with comas.
> > Please verify, deduce from`set_input_products` what it accepts and test that.
> 
> As a comma-separated string, like described here:
> [#223 (comment)](https://github.com/oda-hub/dispatcher-app/pull/223#discussion_r694568588)

It does not really matter if `str(list(...))` appears in the file. It's not likely to happen. It's good to test it though, why not. 
What's best to make sure, however, is that it is not inserted in the products url, since frontend will not allow it.

burnout87:
> > > > > > Is that a case only for the api?
> > > > > 
> > > > > 
> > > > > Please verify. Normally frontend should forbid this.
> > > > 
> > > > 
> > > > frontend does not indeed allow this, but I would include also a case where the list is specified as string in a file. Does it make sense to consider also this case?
> > > 
> > > 
> > > List as string? In what form?
> > > Currently, in the file, it is specified as new-line-separated string with comas.
> > > Please verify, deduce from`set_input_products` what it accepts and test that.
> > 
> > 
> > As a comma-separated string, like described here:
> > [#223 (comment)](https://github.com/oda-hub/dispatcher-app/pull/223#discussion_r694568588)
> 
> It does not really matter if `str(list(...))` appears in the file. It's not likely to happen. It's good to test it though, why not.
> What's best to make sure, however, is that it is not inserted in the products url, since frontend will not allow it.

Actually the dispatcher has always set the `scw_list` passed in the file as part of the `par_dic`, thus having the list in product url. Though the value of `use_scws` is properly set to a value that frontend will accept

volodymyrss:
> > > > > > > Is that a case only for the api?
> > > > > > 
> > > > > > 
> > > > > > Please verify. Normally frontend should forbid this.
> > > > > 
> > > > > 
> > > > > frontend does not indeed allow this, but I would include also a case where the list is specified as string in a file. Does it make sense to consider also this case?
> > > > 
> > > > 
> > > > List as string? In what form?
> > > > Currently, in the file, it is specified as new-line-separated string with comas.
> > > > Please verify, deduce from`set_input_products` what it accepts and test that.
> > > 
> > > 
> > > As a comma-separated string, like described here:
> > > [#223 (comment)](https://github.com/oda-hub/dispatcher-app/pull/223#discussion_r694568588)
> > 
> > 
> > It does not really matter if `str(list(...))` appears in the file. It's not likely to happen. It's good to test it though, why not.
> > What's best to make sure, however, is that it is not inserted in the products url, since frontend will not allow it.
> 
> Actually the dispatcher has always set the `scw_list` passed in the file as part of the `par_dic`, thus having the list in product url. Though the value of `use_scws` is properly set to a value that frontend will accept

evidently, no https://github.com/oda-hub/doc-review-guidelines/issues/57

Let's make an exhaustive test and see.

burnout87:
> 
> Let's make an exhaustive test and see.

Ok I just quickly verified locally and the list is wrongly formatted. I will adjust it and adjust the test also.

volodymyrss:
> > Let's make an exhaustive test and see.
> 
> Ok I just quickly verified locally and the list is wrongly formatted. I will adjust it and adjust the test also.

Ok! I would say very exhaustive test here is even more important than refining. Several things are unfortunately designed with input product passing, what concerns to `use_scws` option as well as file and string  content.

What's important is that input is parsed unambiguously when possible and conventional and returining error when not.

After be parsed, it should lead to be:

* products url always with use_scws
* as long as scw_list is provided in products url, it should be understandable by the
* no unnecessary parameters, including use_scws, in the API code.
* scw_list formatted in consistent way (let's make it coma separated string) in the API code

cdci_data_analysis/flask_app/dispatcher_query.py
volodymyrss: if `scw_list` is not set, frontend can still benefit from `use_scws`. Else it takes the default which is fine sometimes, but why not set always?

And for file, yes, another case.

burnout87:
> 
>     * products url always with use_scws

yes this is checked

> 
>     * as long as scw_list is provided in products url, it should be understandable by the frontend

For this, `scw_list` as a comma-separated string should be provided. For this purpose, there is now an additional an additional check in the dispatcher that will, if needed, re-format the list before generating the products url, in order to be accepted for the frontend

> 
>     * no unnecessary parameters, including use_scws, in the API code.

`use_scws` in the API code is checked, for the others, a separate dedicated test will be developed

> 
>     * scw_list formatted in consistent way (let's make it coma separated string) in the API code

comma-separated string is now in use

Issue #229, opened 2021-08-31
Avoid token validation for url resolution?


burnout87:

burnout87:
After a discussion with @ferrigno @andriineronov and @motame , we realized it is not strictly necessary to include the token when building the `resolve-job-url` link to include within the email.

The token will be in any case validated by the dispatcher once the request is submitted bu the frontend (which will include the token).

burnout87:
Here it is a possible way to visualize an error message coming from a non valid email link.

It is based on using a dedicated error_message, generated from the dispatcher, and visualized over the frontend.

What do you think?

![image](https://user-images.githubusercontent.com/3757833/131868062-9e9e1625-8af8-46cf-bc32-93f173a8322a.png)

volodymyrss:
> After a discussion with @ferrigno @andriineronov and @motame , we realized it is not necessary to include the token when building the `resolve-job-url` link to include within the email.
> 
> The token will be in any case validated by the dispatcher once the request is submitted bu the frontend (which will include the token).

the problem with removing check is that it has to be revised very carefully, since it is a sensitive point.
but we discussed it in the PR.

token prevents from learning request content based on the session id/job id. This is not too dangerous.

The question is however why is it necessary to remove a safety check which changes nothing?
Token pretty likely to live comparable time to the session id / job id.

burnout87:
Also error status code is displayed

![image](https://user-images.githubusercontent.com/3757833/132491544-2969e4b3-44bc-4992-945c-4b1ec77e2456.png)

PR #230, opened 2021-08-31
handling errors during resolve job url

burnout87:


burnout87:
The token is not included in the url for the job resolution, then also, when validating the `job_id`, this is not considered.

cdci_data_analysis/analysis/tokenHelper.py
```diff
@@ -59,10 +61,13 @@ def get_token_user_fail_email(decoded_token):
...
 
-def get_decoded_token(token, secret_key):
+def get_decoded_token(token, secret_key, validate_token=True):
     # decode the encoded token
     if token is not None:
-        return jwt.decode(token, secret_key, algorithms=[default_algorithm])
+        if validate_token:
+            return jwt.decode(token, secret_key, algorithms=[default_algorithm])
+        else:
+            return json.loads(base64.b64decode(token.split(".")[1]+"=").decode())
```
volodymyrss, line 70: I am not sure this is guaranteed to always work. It was a simple trick useful in client when jwt is not installed. But in dispatcher, one should use proper jwt decode without validation.

volodymyrss:
This could potentially become dangerous. It could be useful in some cases, but it depends. Did it come from a use case?

volodymyrss (changes requested):
It is actually a new feature. It might be useful but it is to discuss. It is a sensitive location and review should be careful.

ferrigno:
Yes. I clicked on an URL with expired token and it returned a json, very user unfriendly.

burnout87:
> This could potentially become dangerous. It could be useful in some cases, but it depends. Did it come from a use case?

Mainly if the token is expired, and the user clicks on the link received by the email.

volodymyrss:
> > This could potentially become dangerous. It could be useful in some cases, but it depends. Did it come from a use case?
> 
> Mainly if the token is expired, and the user clicks on the link received by the email.

These URLs themselves can not live very long, since they rely on internal store which we can not promise to maintain long-term. It's even said in the email.
We should make token live longer though, as agreed.

There could be another use case - to she the URL. With the same challenges.
So, we could do this, carefully, but it's a bit unfortunate anyway.

Maybe better next step would be to gzip the url. E.g. scwlist is quite compressible. I made a placeholder there but did not complete.

burnout87:
But the idea here, is to avoid the visualization of the json (with the related error eg token expired) over the browser, and have the user being directed to the frontend where he would have still the error message more clearly visualized.

volodymyrss:
> But the idea here, is to avoid the visualization of the json (with the related error eg token expired) over the browser, and have the user being directed to the frontend where he would have still the error message more clearly visualized.

ok, so resolve URL would do the same valiadation but return redirect instead of just error. That's ok.

burnout87:
> > But the idea here, is to avoid the visualization of the json (with the related error eg token expired) over the browser, and have the user being directed to the frontend where he would have still the error message more clearly visualized.
> 
> ok, so resolve URL would do the same valiadation but return redirect instead of just error. That's ok.

Is there already a way to redirect to a location (which in our case, is the frontend) including an error message (eg token expired) ?

burnout87:
> > But the idea here, is to avoid the visualization of the json (with the related error eg token expired) over the browser, and have the user being directed to the frontend where he would have still the error message more clearly visualized.
> 
> ok, so resolve URL would do the same valiadation but return redirect instead of just error. That's ok.

[https://github.com/oda-hub/dispatcher-app/issues/229#issuecomment-911784726](https://github.com/oda-hub/dispatcher-app/issues/229#issuecomment-911784726)

burnout87:
After discussion we decided to re-introduce the token in the job resolution url. In case of error during the resolution, a redirect still takes place but with an error message explaining the reasons, that will be displayed over the frontend in a more user friendly manner..

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -547,20 +545,25 @@ def resolve_job_url(self):
...
-    def validate_job_id(self, request_parameters_from_scratch_dir=False):
+    def validate_job_id(self, request_parameters_from_scratch_dir=False, job_resolution=False):
         """    
         makes sure self.job_id is compatible, for given user token, with:
         if request_parameters_from_scratch_dir:
             parameters found in directory matching job_id
         else: 
             parameters provided in this request
+
+        if job_resolution:
```
volodymyrss, line 563: this is no more, yes?

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -547,20 +545,25 @@ def resolve_job_url(self):
...
-        self.validate_job_id(request_parameters_from_scratch_dir=True)
+        self.validate_job_id(request_parameters_from_scratch_dir=True, job_resolution=True)
 
         return self.generate_products_url_from_par_dict(
             self.app.config['conf'].products_url,
             self.par_dic
             )
 
-    def validate_job_id(self, request_parameters_from_scratch_dir=False):
+    def validate_job_id(self, request_parameters_from_scratch_dir=False, job_resolution=False):
```
volodymyrss, line 555: job_resolution is no more used; maybe it can be kept, but with TODO or a message

cdci_data_analysis/flask_app/app.py
```diff
@@ -298,13 +289,22 @@ def test_mock():
...
 
+        return redirect(location, 302)
+    except APIerror as e:
+        message_dict = {
+            "error_message": e.message,
+            "status_code": e.status_code
+        }
+        location = '%s?%s' % (app.config['conf'].products_url, urlencode(message_dict))
+        return redirect(location, 302)
+    except Exception as e:
```
volodymyrss, line 306: when does this happen? in unexpected error? And then it does what, just redirects to bare site?

cdci_data_analysis/flask_app/templates/email.html
```diff
@@ -27,6 +27,8 @@
 {% else %}
 Note that this url is NOT permanent, and will expire within about a week.<br>
```
volodymyrss, line 28: maybe let's make this "NOT" bold too?

volodymyrss (changes requested):

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: I will just remove it. It is no longer needed

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: indeed, remained there from a previous commit

cdci_data_analysis/flask_app/app.py
burnout87: Mainly is there to always redirect to the frontend page, but of course an addition more informative message is necessary

volodymyrss: Indeed! Would be quite unfortunate to click on the link and and up on the front page with no obvious error. It's bad to show incomprehensive errors, but it's even worse to pretend there is no error and do nothing! As I user I could be really annoyed.

burnout87: Now more error information are included in the redirect url

volodymyrss (approved):

PR #232, opened 2021-09-06
implement special case of use_scws in resolve_url

volodymyrss:


volodymyrss:
unclear why coverage is shown reduced while all added lines but two are covered. may need to look into it

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -401,6 +401,8 @@ def set_scws_related_params(self, request):
         # it is nowhere necessary within the dispatcher-app,
         # but it is re-attached to the url within the email
         # when sending it since it is used by the frontend
+
+        # why use both `request` and `self.par_dict`? par_dict should be derived from request while preserving list
```
burnout87, line 405: Most likely, something that escaped while  refactoring...good point indeed

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -530,13 +532,17 @@ def prepare_download(self, file_list, file_name, scratch_dir):
         return tmp_dir, file_name
 
     def resolve_job_url(self):
-        request_par_dic = self.get_request_par_dic()
+        self.par_dic.update(self.get_request_par_dic())
+
+        # what if scw list from request overwrites that in self.par_dict?
+        self.set_scws_related_params(request)
```
burnout87, line 538: If that happens, then it is only supposed to overwrite with a different formatting, it's not going to overwrite it with a different one.

burnout87 (approved):
Do you plan to implement also the change in the `set_scws_related_params` method?

Otherwise I can take care of it in a dedicated PR.

cdci_data_analysis/flask_app/dispatcher_query.py
volodymyrss: oh but it's going to change the parameters.
here I added a test to show how it happens:

https://github.com/oda-hub/dispatcher-app/blob/54ed837d68c4e61dd707a7252c8732cdf8bc962b/tests/test_job_management.py#L1318-L1322

I am not even sure it is a bug. A bit of an unintended behavior. Easy to prevent

volodymyrss:
> Do you plan to implement also the change in the `set_scws_related_params` method?
> 
> Otherwise I can take care of it in a dedicated PR.

please, though I do not think this is most urgent. So as long as it does not take too long.

tests/test_job_management.py
```diff
@@ -1310,7 +1310,17 @@ def test_email_scws_list(dispatcher_live_fixture,
...
+
                     r = requests.get(extracted_product_url.replace('PRODUCTS_URL/dispatch-data', server))
+
+                    # this is to show how parameters can be overwritten in resolve; this never happens intentionally and is not dangerous
+                    # could be prevented
+                    alt_scw_list = ['066400220010.001', '066400230010.001']
+                    r_alt = requests.get(extracted_product_url.replace('PRODUCTS_URL/dispatch-data', server), params={'scw_list': alt_scw_list})
+                    alt_extracted_scw_list = parse_qs(parse.urlparse(r_alt.url).query)['scw_list'][0].split(",")
+                    assert alt_extracted_scw_list == alt_scw_list
+                    assert alt_extracted_scw_list != params['scw_list']
```
burnout87, line 1322: Ok I see. 

Can we implement a check on the dispatcher-side to prevent this? Can we assume that when `resolve-job-url` is called, then only two (or three) params are expected? Those would be `job_id`, `session_id` and `token`?

volodymyrss: sure, but I am not even sure it is a bug. A bit of an unintended behavior. Easy to prevent.

burnout87:
> > Do you plan to implement also the change in the `set_scws_related_params` method?
> > Otherwise I can take care of it in a dedicated PR.
> 
> please, though I do not think this is most urgent. So as long as it does not take too long.

I checked again the code and re-understood why the `request` parameter is needed (or at least, with the current design) and was also used before this mall refactoring. Here: 

```request.values.getlist('scw_list')```

allows to obtain the `scw_list` parameter when this is passed in the request as list. In such a case, by simply using:

```self.par_dic['scw_list']```

only the first element of the list would be returned.

Surely this can be handled in the same function where the `par_dic` is filled, but as it is now, all the scw-related parameters are handled together when building the dispatcher object. So, I think is ok to leave it like that.

volodymyrss:
of course!
it's quite unfortunate to re-read `request`. The treatment does not need to be scw-related in par_dict, just one of the two options:
* some parameters will be loaded as lists when `par_dic` is  filled. noted which ones elsewhere.
or
* parameter read as string and parsed in single scw-treating function

PR #233, opened 2021-09-07
neutralized frontend defaults

volodymyrss:


volodymyrss:
another option is to add all of the valid parameter defaults in `oda_api` request. It knows the list from validation.

volodymyrss:
if the approach with setting defaults prior to job id works, this is not needed. let's see

volodymyrss:
now instead of neutralizing we actively set them with #243

PR #235, opened 2021-09-07
new possible job status

volodymyrss:


volodymyrss:
This is one of the other possible statuses.
The problem only occurs when client (e.g. frontend) requests with this status in the request.

If this happens is a bit random since the last status is sent.

burnout87 (commented):
I don't understand in which case this status is found. 

Will it need to be handled also during the execution of `run_query` ?

volodymyrss:
> I don't understand in which case this status is found.
> 
> Will it need to be handled also during the execution of `run_query` ?

It's one of the possible callback statuses. I guess at some point the treatment became more strict and this started to fail.
There was no test.
I added a test.

cdci_data_analysis/analysis/job_manager.py
```diff
@@ -95,7 +95,7 @@ def __init__(self,
 
     @staticmethod
     def get_allowed_job_status_values():
-        return ['done', 'main_done', 'failed', 'progress', 'submitted', 'ready', 'unknown', 'unaccessible','aliased', 'post-processing']
```
burnout87: Why removed back? Will this keep triggering the 410 error?

volodymyrss: Please consider the rest of the code. All unexpected callbacks are now treated as progress updates, since it's better to send it as so for the user.

cdci_data_analysis/analysis/job_manager.py
```diff
@@ -186,8 +186,11 @@ def write_dataserver_status(self, status_dictionary_value=None,
         if status_dictionary_value is None:
             pass
         else:
-            self.monitor['status'] = status_dictionary_value
-
+            if status_dictionary_value in self._allowed_job_status_values_:                
+                self.monitor['status'] = status_dictionary_value
+            else:
+                self.monitor['status'] = "progress" # any unknown message is progress
```
burnout87, line 192: This will divert any unknown status to `progress`, right? Would not be useful to issue a warning that an unknown status has been returned?

burnout87 (commented):
Just one question

cdci_data_analysis/analysis/job_manager.py
volodymyrss: I thought about it but it would produce too many messages. The callbacks were made to provide a lot of auxiliary information in anticipation of several future features, this includes other statuses. So regulating their content from dispatcher too much is not ideal.
I added debug.

burnout87 (approved):

PR #241, opened 2021-09-10
content temp_dir setting inputs

burnout87:


burnout87:
In case of error when uploading th inputs from the frontend (eg scw_list file) the content of the temp_dir is printed and sent to sentry.

volodymyrss:
ok, I guess it's a bit trickly, although possible to test?

volodymyrss (approved):

burnout87:
> ok, I guess it's a bit trickly, although possible to test?

It is indeed, but there is one test now that covers it, just partially though.

Once the same behavior will be reproduced then we'll be able to build a more tailored test.

volodymyrss:
> > ok, I guess it's a bit trickly, although possible to test?
> 
> It is indeed, but there is one test now that covers it, just partially though.
> 

Well it's non-critical functionality, as long as all existing tests pass it's sufficient.

> Once the same behavior will be reproduced then we'll be able to build a more tailored test.

Is it not possible to artificially reproduce it by creating a file in the directory from test? Maybe it's tricky if it was actually a sort of a race condition.

burnout87:
> > Once the same behavior will be reproduced then we'll be able to build a more tailored test.
> 
> Is it not possible to artificially reproduce it by creating a file in the directory from test? Maybe it's tricky if it was actually a sort of a race condition.

Maybe it was like that, reason why is not easy to reproduce.

burnout87:
> > 
> > Is it not possible to artificially reproduce it by creating a file in the directory from test? Maybe it's tricky if it was actually a sort of a race condition.

Sure it is possible, but I am not sure how I can _simulate_ a race condition for this specific case

volodymyrss:
> > > Is it not possible to artificially reproduce it by creating a file in the directory from test? Maybe it's tricky if it was actually a sort of a race condition.
> 
> Sure it is possible, but I am not sure how I can _simulate_ a race condition for this specific case

in the test you might be able to insert the file in a more controlled way. Anyway it does not matter.

Oh, and if it's race, maybe we could make anticipate it? Maybe retry deleting?

burnout87:
> 
> Maybe retry deleting?

That's more or less what I did now, forcing the deletion of the file in the temp folder, directly from the test.

cdci_data_analysis/analysis/instrument.py
```diff
@@ -155,45 +156,76 @@ def parse_inputs_files(self,
                              verbose,
                              use_scws,
                              sentry_client=None):
+        # if any, for exceptions additional imformation
```
volodymyrss, line 159:
```suggestion
        # if any, for exceptions additional information
```

tests/test_job_management.py
```diff
@@ -1395,7 +1395,7 @@ def test_email_scws_list(dispatcher_long_living_fixture,
         'please check the inputs')
     error_message_scw_list_missing_file = (
         'scw_list file was expected to be passed, but it has not been found, '
-        'please check the inputs')
+        'please check the inputs, content of the temporary directory is []')
```
volodymyrss, line 1398: I do not suppose we content of temp directory in this case, when error is something else.

cdci_data_analysis/analysis/instrument.py
```diff
@@ -155,45 +156,76 @@ def parse_inputs_files(self,
...
         # TODO probably exception handling can be further improved and/or optmized
         # set catalog
         try:
             self.upload_catalog_from_fronted(par_dic=par_dic, request=request, temp_dir=temp_dir)
         except Exception as e:
+            if temp_dir is not None and os.path.exists(temp_dir):
+                content_temp_dir = os.listdir(temp_dir)
+            error_message = 'Error while uploading catalog file from the frontend'
+            if content_temp_dir is not None:
+                error_message += f', content of the temporary directory is {content_temp_dir}'
```
volodymyrss, line 170: this block is almost exactly repeated 5 times. It can be made in context or a decorator. Or even just single try-except block (in this case it to keep the same message we'd need to store the couple of changing words in a variable)

volodymyrss (changes requested):

tests/test_job_management.py
burnout87: You mean we're not interested? You're right, but the exception happens exactly when the inputs are parsed, and the content of the temp_dir is reported.

volodymyrss: > You mean we're not interested in it? You're right, but the exception happens exactly when the inputs are parsed, and the content of the temp_dir is reported.

there is no need to report content of `temp_dir` unless the problem was in deleting the `temp_dir`. We could also separately catch just that exact exception.

cdci_data_analysis/analysis/instrument.py
burnout87: Removed now quite some code repetition.

tests/test_job_management.py
burnout87: Ok, didn't consider that was referred to a case with erroneous inputs

cdci_data_analysis/analysis/instrument.py
burnout87: But of course I'm always open to more improvements

volodymyrss (approved):

PR #248, opened 2021-09-29
No global variable

volodymyrss:


cdci_data_analysis/analysis/instrument.py
```diff
@@ -54,6 +54,8 @@
 # relative import eg: from .mod import f
 
 # list of parameters not to be included in the par_dic object
+
+# TODO: this is not preserved between requests, and is not thread safe. Why not pass it in class instances?
 params_not_to_be_included = ['user_catalog',]
```
burnout87, line 59: Very valid point(s). It makes more sense to have in the class.

cdci_data_analysis/analysis/instrument.py
```diff
@@ -261,6 +269,7 @@ def run_query(self, product_type,
                     # assess the permissions for the query execution
                     self.check_instrument_query_role(query_obj, product_type, roles, par_dic)
 
+                    # why called again?
```
burnout87, line 272: honestly...no idea, we can definitely get rid of one call

cdci_data_analysis/analysis/instrument.py
```diff
@@ -295,6 +304,7 @@ def run_query(self, product_type,
                     query_out.set_failed(product_type, logger=logger, sentry_client=sentry_client, excep=e)
 
         # adding query parameters to final products
+        # TODO: this can be misleading since it's the parameters actually used
         query_out.set_analysis_parameters(par_dic)
```
burnout87, line 308: Yes, true. An alternative approach could be to call this before the rest of the default values is set. But, I would not change it now.

cdci_data_analysis/analysis/instrument.py
```diff
@@ -339,6 +349,7 @@ def get_par_by_name(self,par_name):
 
         for _query in self._queries_list:
             if par_name in _query.par_names:
+                # TODO: this picks the last one if there are many?..
                 p  =  _query.get_par_by_name(par_name)
```
burnout87, line 353: Yes. But can we have actually two parameters with the same fir a given query?

tests/test_job_management.py
```diff
@@ -1404,14 +1408,30 @@ def test_email_scws_list(dispatcher_long_living_fixture,
...
-    jdata = ask(server,
-                params,
-                method=ask_method,
-                max_time_s=150,
-                expected_query_status=None,
-                expected_status_code=None,
-                files=scw_list_file_obj
-                )
+    # this sets global variable
+    requests.get(server + '/api/par-names')
```
burnout87, line 1412: But by clearing up and re-initializing the `params_not_to_be_included` as it is done now, this should no longer be necessary...?

cdci_data_analysis/analysis/instrument.py
volodymyrss: Nothing prevents that. Also, many `query` objects are scanned here. I made a test [here](https://github.com/oda-hub/dispatcher-app/blob/d2808b2d9bd1948b48be28963a23cb0732f3473f/tests/test_unit.py#L48-L55) for curiosity.
No point to fix it here. It could be addressed and fixed separately. It's not important, it's just an unfortunate design.
Parameter name does not need to be globally unique.

tests/test_job_management.py
volodymyrss: This is for test, and this triggers the problem when `params_not_to_be_included` is not cleared up. So by keeping it in the test, we make sure that this particular problem is actually fixed.

burnout87 (approved):

PR #254, opened 2021-10-07
default conversion for time, default t_format, some test adaption

burnout87:


volodymyrss:
> @volodymyrss what do you think of this? still time to include in the release?

Oh, but I set the tag for 21.10.0000. So that marks the release, and we will apply the patches as necessary.
I even applied patch following the hotfix  - 21.10.0001. 

See, I set 21.10.0002  milestone for this. Let's review it carefully, look on it in staging, and then release without much delay.

A problem is also that I am very busy with other stuff for the conference, so I see if I manage to look into this. 

But I will try, thanks anyway!

tests/test_job_management.py
```diff
@@ -543,6 +542,7 @@ def test_validation_job_id(dispatcher_live_fixture):
             'DEC': -29.74516667,
             'T1': '2017-03-06T13:26:48.000',
             'T2': '2017-03-06T15:32:27.000',
+            'T_format': 'isot'
```
volodymyrss, line 545: what happens if this is not passed?

burnout87: Now the `T_format` will be included for the calculation of the `job_id`, and it will be set to the default value of `isot`.

If not specified, or if it has a different value (eg `mjd`), it will be set to that default value, after having converted the value of the Time (eg `T1`, `T2`) to `isot` format.

volodymyrss: Ah ok, it the default functionality will take care of it.

burnout87: Precisely

cdci_data_analysis/analysis/parameters.py
```diff
@@ -440,9 +450,25 @@ def __init__(self,value=None,T_format=None,name=None,Time_format_name=None):
...
+    def set_group_par(par_group, par_dic, params_not_to_be_included=None, verbose=False):
+        if params_not_to_be_included is None:
+            params_not_to_be_included = []
+        num_time_params = 0
+        for par in par_group:
+            if par.name is not None and par.name not in params_not_to_be_included:
+                par.set_from_form(par_dic, verbose=verbose)
+                num_time_params += 1
+        if num_time_params > 0:
+            par_dic['T_format'] = 'isot'
```
volodymyrss, line 465: this could be more generic, apply to all parameters. `'T_format'` is in `units_name` actually. And the default format should be stored somewhere too.

Then the decision to add a request parameter with default units name can be made more generally. Also preferrably by a function returning complementary parameter dict, not by modifying passed dict.

tests/reference_emails/submitted-dummy.html
```diff
@@ -42,19 +42,20 @@
 disp = DispatcherAPI(url=&#34;PRODUCTS_URL/dispatch-data&#34;, instrument=&#34;mock&#34;)<br>
 <br>
 par_dict = {<br>
-    &#34;product_type&#34;: &#34;Real&#34;,<br>
+    &#34;DEC&#34;: -29.74516667,<br>
```
volodymyrss, line 45: I must say alphabetic order is less readable, many sequences are not in regular order (e.g. RA-DEC, product is important enough to be first, etc). But it does not matter so much. Of course it's important to have defined ordering. Maybe later we can add different sorting. Not now.

tests/test_job_management.py
```diff
@@ -1955,6 +1953,80 @@ def x(arg):
...
+        "tem": 0
+            }
+    encoded_token = jwt.encode(token_payload, secret_key, algorithm='HS256')
+
+    dict_param = dict(
+        query_status="new",
+        query_type="Real",
+        instrument="empty-async",
+        product_type="dummy",
+        T1="57818.560277777775",
```
volodymyrss, line 1975: What will happen with invalid format? 
Also what if `T1=57818.560277777775` and  `T2="2011-11-11T11:11:11"`?

volodymyrss (changes requested):

cdci_data_analysis/analysis/instrument.py
```diff
@@ -128,24 +129,21 @@ def set_pars_from_dic(self, par_dic, verbose=False):
...
-            for par in (query_obj.parameters +
-                        self.instrumet_query.parameters +
-                        self.src_query.parameters):
-                # this is required because in some cases a parameter is set without a name (eg UserCatalog),
-                # or they don't have to set (eg scw_list)
-                # 
-                if par.name is not None and par.name not in params_not_to_be_included:
-                    par.set_from_form(par_dic, verbose=verbose)
-
-                self.logger.info("set_pars_from_dic>> par: %s par.name: %s par.value: %s par_dic[par.name]: %s", par, par.name, par.value, par_dic.get(par.name, None))
```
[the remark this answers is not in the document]
burnout87: because I moved directly inside the `set_group_par` method

cdci_data_analysis/analysis/parameters.py
```diff
@@ -246,19 +258,20 @@ def set_from_form(self, form, verbose=False):
...
+            self.set_par(value=v, form=form, units=u)
         else:
             # set the default value
             form[par_name] = self.value
+
             if verbose is True:
                 logger.debug('setting par: ', par_name, ' not in dictionary, setting to the default value')
 
-    def set_par(self, value, units=None):
+    def set_par(self, value, form, units=None):
```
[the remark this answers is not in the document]
burnout87: ok, let's also keep in mind that in also several other locations this approach has been used

tests/reference_emails/submitted-dummy.html
burnout87: I thought the same, I will add a dedicated TODO

tests/test_job_management.py
burnout87: It really depends by what is specified (if it is!) inside `T_format`.
>
>  Also what if `T1=57818.560277777775` and `T2="2011-11-11T11:11:11"`?

it will throw an excepotion when it will attempt a conversion, depending on the value of `T_format`

cdci_data_analysis/analysis/parameters.py
burnout87: the problem in this case, for a `Time` parameter more specifically, is that there is one single units name (`T_format`) that will be applied to all the time parameters (`T1` and `T2`), so it becomes tricky to generalize

tests/test_job_management.py
volodymyrss: I see now it is not exactly the normal defaults mechanism since `T_format` does not correspond to a parameter name, but to it's units. It's a different point but related to the rest of the PR.

cdci_data_analysis/analysis/parameters.py
volodymyrss: should it be really a problem? Each parameter could take care of verifying and possibly setting it's units field. If many parameters do it for the same units field, the could behave in a way which safely stacks these validations/sets.

* if unit field is set (either by input or by another parameter before): check that the value fits it
* if unit field is not set (so not in input and no other parameters of the same type considered before): set the default, validate

tests/test_job_management.py
volodymyrss: > It really depends by what is specified (if it is!) inside `T_format`.
> 
> > Also what if `T1=57818.560277777775` and `T2="2011-11-11T11:11:11"`?
> 
> it will throw an exception when it will attempt a conversion, depending on the value of `T_format`

I mean we'd need to test that. I think before format was detected automatically. There are many options.

burnout87: Yes of course, I will include some tests

cdci_data_analysis/analysis/parameters.py
burnout87: what we want is that the unit field is always set to `isot` (or any other default), no matter what, that is why I implemented this approach only for Time parameters

after setting the value (in the form) to the `isot` format for all the Time parameters, set also the units name once for all

burnout87: does it make sense?

volodymyrss: all parameters with unit should be converted to the same, default, unit. Else the same problem as here will occur. Essentially job_id and the URL should not depend on particular representation of the parameter.

cdci_data_analysis/analysis/instrument.py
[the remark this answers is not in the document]
volodymyrss: Ok. I think it might be possible to avoid introducing `set_group_par`, which is an additional structure complexity. But it's a different thread

cdci_data_analysis/analysis/parameters.py
burnout87: > all parameters with unit should be converted to the same, default, unit. Else the same problem as here will occur. Essentially job_id and the URL should not depend on particular representation of the parameter.

ok, this could also help in avoid using the `set_group_par` function

cdci_data_analysis/analysis/instrument.py
```diff
@@ -128,24 +129,37 @@ def set_pars_from_dic(self, par_dic, verbose=False):
...
+                if par.name is not None and par.name not in params_not_to_be_included and par.default_units is not None:
+                    par_dic[par.units_name] = par.default_units
 
         else:
             for _query in self._queries_list:
-                for par in _query.parameters:
-                    if par.name is not None and par.name not in params_not_to_be_included:
-                        par.set_from_form(par_dic, verbose=verbose)
+                for par_type, group_par_type in itertools.groupby(_query.parameters, lambda x: type(x)):
+                    par_type.set_group_par(par_group=group_par_type,
```
volodymyrss, line 159: if it's removed above, is it still needed here?
by the way, this part and the one in the other branch seem partially duplicated. And it's because they are, changes need to put propagated to both manually. Would be better to de-duplicate: derive `param_list` depending on `product_type` , but loop over it only once.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -421,28 +460,48 @@ def check_int_value(value, units=None,name=None):
...
+        list_par_group = list(par_group)
+        for par in list_par_group:
+            if par.name is not None and par.name not in params_not_to_be_included:
+                par.set_from_form(par_dic, verbose=verbose)
+            logger.info("set_pars_from_dic>> par: %s par.name: %s par.value: %s par_dic[par.name]: %s",
+                        par, par.name, par.value, par_dic.get(par.name, None))
+
+        # set the default, isot
+        for par in list_par_group:
+            par_dic[par.units_name] = 'isot'
```
volodymyrss, line 495: this `"isot"` value is stored in `default_units`. And if that's used, the method doing this can be in Parameter (except maybe if possible to avoid new `set_group_par` entirely, I do not see why not, since it does not count anything anymore).

volodymyrss (changes requested):

cdci_data_analysis/analysis/instrument.py
burnout87: removed! forgot to remove some code

cdci_data_analysis/analysis/parameters.py
burnout87: It is now indeed like that, and there is no longer `set_group_par`, since it is not needed.

cdci_data_analysis/analysis/parameters.py
volodymyrss: > > all parameters with unit should be converted to the same, default, unit. Else the same problem as here will occur. Essentially job_id and the URL should not depend on particular representation of the parameter.
> 
> ok, this could also help in avoid using the `set_group_par` function

Just wanted to clarify that it applies to all parameters which have meaningful units. But it also can be things like different number representations.

burnout87: Till now I have seen only applied to `Time` parameters. But in the future there could be also other parameters to be treated in the same way

cdci_data_analysis/analysis/parameters.py
```diff
@@ -188,6 +187,7 @@ def __init__(self,
         self._allowed_values = allowed_values
         self.name = name
         self.units = units
+        self.default_units = default_units
```
volodymyrss, line 190: could be probably private (start with `_`) like `_allowed_units`. And also it should be checked that default units are in allowed units. For sanity.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -246,31 +269,33 @@ def set_from_form(self, form, verbose=False):
...
 
         if in_dictionary is True:
-            self.set_par(value=v, units=u)
+            form[par_name] = self.set_par(value=v, units=u)
         else:
             # set the default value
             form[par_name] = self.value
+
             if verbose is True:
                 logger.debug('setting par: ', par_name, ' not in dictionary, setting to the default value')
```
volodymyrss, line 283: this is not how logging is to be used. Actually this will raise an exception in logging. I realize it was not added just now, but still.

cdci_data_analysis/analysis/instrument.py
```diff
@@ -128,24 +129,26 @@ def set_pars_from_dic(self, par_dic, verbose=False):
...
+            # or they don't have to set (eg scw_list)
+            if par.name is not None and par.name not in params_not_to_be_included:
+                # set the value for par in to a default format,
+                # or to a default value if this is not included within the request
+                par.set_value_from_form(par_dic, verbose=verbose)
+            self.logger.info("set_pars_from_dic>> par: %s par.name: %s par.value: %s par_dic[par.name]: %s",
+                             par, par.name, par.value, par_dic.get(par.name, None))
+            if par.name == "scw_list":
+                self.logger.info("set_pars_from_dic>> scw_list is %s", par.value)
+        for par in param_list:
```
volodymyrss, line 149: why identical second loop?

cdci_data_analysis/analysis/instrument.py
```diff
@@ -128,24 +129,26 @@ def set_pars_from_dic(self, par_dic, verbose=False):
...
-                        par.set_from_form(par_dic, verbose=verbose)
+            param_list = [par for _query in self._queries_list for par in _query.parameters]
+
+        for par in param_list:
+            # this is required because in some cases a parameter is set without a name (eg UserCatalog),
+            # or they don't have to set (eg scw_list)
+            if par.name is not None and par.name not in params_not_to_be_included:
+                # set the value for par in to a default format,
+                # or to a default value if this is not included within the request
+                par.set_value_from_form(par_dic, verbose=verbose)
```
volodymyrss, line 144: I'd have this function return value, and maybe even units (as tuple). This way also `set_value_from_form` will not manipulate its arguments.

tests/test_job_management.py
```diff
@@ -1955,6 +1953,106 @@ def x(arg):
     assert len(my_globals['scwl_dict']['scw_list']) > max_length
 
 
+@pytest.mark.parametrize('time_combinations', [[57818.560277777775, 57818.64753472222],
+                                               ['2017-03-06T13:26:48.000', '2017-03-06T15:32:27.000'],
+                                               ['2017-03-06T13:26:48.000', 57818.64753472222],
+                                               [57818.560277777775, '2017-03-06T15:32:27.000']])
+@pytest.mark.parametrize('time_format', ['isot', 'mjd'])
+def test_email_t1_t2(dispatcher_live_fixture,
```
volodymyrss, line 1961: make this long-living fixture? To test faster. I doubt multi-threading counts here.

volodymyrss (changes requested):
I think it's much shorter and more structured already! Couple of questions still.

cdci_data_analysis/analysis/parameters.py
volodymyrss: we should try if the same already happens for, e.g., E1_keV, with values like 25 or 25.0 . Does it get normalized somewhere?

cdci_data_analysis/analysis/parameters.py
burnout87: do we still enable somewhere the `verbose` option?

cdci_data_analysis/analysis/parameters.py
burnout87: same approach used for checking `units`

cdci_data_analysis/analysis/parameters.py
volodymyrss: yes, well, `logging` is controlled without explict verbose, depends how the app is started. One way in test, another in gunicorn, third in ther https://github.com/oda-hub/dispatcher-app/blob/master/bin/run_osa_cdci_server.py#L66

What's important is to set message levels and logger names.

cdci_data_analysis/analysis/instrument.py
burnout87: It's not identical, they serve two different purposes, that should be treated in separately:

- the first one takes care of setting the various values in the `form`, (for the parameters not present in the `form`, those are set to their default value), as well as in their default format (eg `isot` for `T1` and `T2`)
- the second sets the `units_name` in the `form` for those parameters where this is specified (eg `T1` and `T2`)

The reason for having two distinct loops is related to ho the assigning of the value to the `_astropy_time` ([here](https://github.com/oda-hub/dispatcher-app/blob/64e032b28fe80fbcf8ea2cda532e592befb6ecce/cdci_data_analysis/analysis/parameters.py#L487-L496)): by combining the two loops together we would have the situation where starting from one time parameter, with a non default units (eg `mjd`), an `astropyTime` object is attempted to be instantiated with the default format: this causes an exception.

Probably there is a different and better way to instantiate `astropyTime` objects, so I am open to suggestions.

cdci_data_analysis/analysis/parameters.py
burnout87: For `E1_keV` (or `E2_keV`), the value that is passed (which will always be a string) is checked if it is in a valid format (`int` of `float` are accepted), so both the options you mentioned are accepted.

Regarding the setting of the `units_name` entry in the form, this is now applied for each parameter where this is specified, and to the defined `default_units`.

cdci_data_analysis/analysis/instrument.py
burnout87: does it make sense?

cdci_data_analysis/analysis/parameters.py
volodymyrss: > For `E1_keV` (or `E2_keV`), the value that is passed (which will always be a string) is checked if it is in a valid format (`int` of `float` are accepted), so both the options you mentioned are accepted.
> 
> Regarding the setting of the `units_name` entry in the form, this is now applied for each parameter where this is specified, and to the defined `default_units`.

checked - yes, but will it be transformed to the same form before generation of `job_id` and the URL? 25 and 25.0 mean exactly the same here but might look different.

cdci_data_analysis/analysis/instrument.py
volodymyrss: Not particularily, sorry. When I test actually find another issue, with `None` key being inserted in par_dic, and can not be jsonified. I am looking into it.

cdci_data_analysis/analysis/parameters.py
burnout87: passing 20 is different from passing 20.0 (or even 20.): the first will be interpreted as an integer whereas the second as a float, and will also yield different `job_id`

cdci_data_analysis/analysis/instrument.py
volodymyrss: Ok I see. It's because the true input unit is irreversibly overwritten when the first parameter of a given type is is processed. 

This is consequence of modifying arguments, which I think is often misleading (even if sometimes very beneficial).

Your approach is a way, I suggest to keep both input and output par_dic and substitute at the end. See #278

burnout87: Yes, that's the problem, it was not easy to explain. It comes out and it's clear when debugging.

What you suggest makes sense so we also avoid to modify the `par_dic`.

cdci_data_analysis/analysis/parameters.py
burnout87: To address this, we should apply the approach used for the `Time` parameter, also to the `SpectralBoundary`, and perhaps also to other parameters if necessary. But in a separate dedicated PR.

volodymyrss: Maybe separate PR, yes, but what are the consequences of what is done in this one? Is it not so that at the general level (not specific to time), parameter value is taken from the instance, after it's set. In this case, no matter what is the type/unit, it will be valid. 

Maybe this general approach is suitable for time too? If so, it would be in this PR. And it will spread on other cases automatically.

burnout87: What I meant is that for each type of parameter, a default format of the value should be adopted, and then used (ie set inside `par_dic`, and used for later `job_id`(s) generation). Therefore for each parameter a different discussion (and related PR) should be opened.

As for this PR, now the `job_id` calculation will always be performed with the time values in `isot` format, and the `units_name` set to `isot`.

volodymyrss: > What I meant is that for each type of parameter, a default format of the value should be agreed, and then adopted (ie set inside `par_dic`, and used for later `job_id`(s) generation). Therefore for each parameter a different discussion (and related PR) should be opened.
> 
> As for this PR, now the `job_id` calculation will always be performed with the time values in `isot` format, and the `units_name` set to `isot`.

That's a lot of nearly-identical PR's for each parameter. There can be a generic adjustment of Parameter class with reasonable defaults. Since some generic adjustments will be necessary, there is no sense to separate in many.

But if you want to be done with this block, we can merge this, and develop a suitable treatment in one another PR.

volodymyrss (approved):

volodymyrss:
Except let's just wait for it to pass tests.

cdci_data_analysis/analysis/parameters.py
burnout87: Yes, in a dedicated PR. This was specific to `Time` parameter.

volodymyrss: > Yes, in a dedicated PR. This was specific to `Time` parameter.

For something specitifc to `Time` parameter this PR made lot's of changes in generic part, not directly related to `Time` parameter.

PR #263, opened 2021-10-19
Hotfix 262 - `inspect_state` functionality

burnout87:


burnout87:
> we should add a separate role for this

I called this new role `jobs manager`, but we can discuss on a more appropriate one.

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -318,25 +318,35 @@ def __init__(self, app,
...
             secret_key = app.config.get('conf').secret_key
             decoded_token = tokenHelper.get_decoded_token(token, secret_key)
             logger.info("==> token %s", decoded_token)
         except jwt.exceptions.ExpiredSignatureError:
-            raise RequestNotAuthorized("The token provided is expired.")
+            # raise RequestNotAuthorized("The token provided is expired.")
+            return make_response('The token provided is expired.'), 403
         except jwt.exceptions.InvalidTokenError:
-            raise RequestNotAuthorized("The token provided is not valid.")
+            # raise RequestNotAuthorized("The token provided is not valid.")
```
volodymyrss, line 329: why, it did not give the right message? It's treated by error handled of flask. But fine.

volodymyrss (approved):

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: It is more about the usage and intended scope we want to make of this functionality. But ok for me to use the `flask` one, in the other way we can have an output more consistent with the output we have from the other endpoints.

volodymyrss: whenever the request is not authorized it could just be RequestNotAuthorized captured in the same place. Output can be adjusted. If there is a way to approach to more to a single way, let's do it here. Else it can be refactored later.

volodymyrss:
Oh and the test breaks.

volodymyrss (changes requested):
Test breaks?

burnout87:
> Test breaks?

Seen it. Now it should go

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: Ok, I'll leave as it is now. There are already some `TODO`s for the output adaptation to the `QueryOutJSON` JSONSchema, and this can be taken care when this adaptation will be done.

burnout87:
> Test breaks?

And now they passed!

volodymyrss (approved):

Issue #261, opened 2021-10-15
give more information with an exception


volodymyrss:
to produce not-so-comprehensive error:

* open mmoda front page, with all default parameters
* put `Crab` in source name. Click `resolve`.
* click `submit` - to get isgri image, by default.

this will appear:

![image](https://user-images.githubusercontent.com/3909535/137542789-6c21131b-aaea-49df-abec-ac1c421598f9.png)

there is no useful result there. But we could make the message more informative. E.g., add:

```
Unfortunately, the analysis product you requested is empty, since there is no usable data for the parameter combination you requested: time period, software version, etc.
We did not find any unexplained exceptions.
This additional message may be helpful:

HERE_PUT_BACKEND_MESSAGE
```

burnout87:
Something like that would already be better I guess. I will add the additional information

![image](https://user-images.githubusercontent.com/3757833/138255699-5fa12ce3-9a55-436b-b93f-0c00d242ad02.png)

Issue #227, opened 2021-08-25
convert T1 and T2 to ISOT **before** computing job ID


volodymyrss:

burnout87:
I can take care of this since it is needed for the release, is there  an example code I can inspect to understand the conversion to apply?

volodymyrss:
if, for example, `T1=51000 ` in API, a job_id will be calculated with this value.
But, once parameter is processed, it will be converted in the same format, ISOT. And this will appear in the url. Which means that UTL will lead to another job_id.
This can be verified with a simple test.

burnout87:
some small investigation is needed on when and how the parameters are set

volodymyrss:
this seems like an appropriate test but inappropriate software behavior.

ok, but, if you do not set `T1`, it's set to what? To the default, right? Taken from the `Parameter` instance?

burnout87:
> this seems like an appropriate test but inappropriate software behavior.
> 

inappropriate because the frontend would not know what to do with those values received from the URL?

burnout87:
> 
> I guess it would, given also the parameter on time format. But let's not do this!
> 

I don't think there's a conversion applied at frontend level, that uses also the time format information yet, but I might be wrong

> Instead, we need it so that the parameter from `par_dic` is set in instance of `class Time(Parameter)`. And then the value is extracted from this instance. And it needs to be extracted in standardized format. And only this value should be used in the job_id, URL, api code. This applies to many parameters, in principle. But it's most obvious for `Time`.

Ok, I think I already have an idea on how to proceed and where to put mu hands on, "should" be not too much work. 

So, we want that the time format used for job_id, URL, api code remains consistent across them, and this is the following: 

`2017-03-06T13:26:48.000`

PR #264, opened 2021-10-21
additional message in case of connection issues 

burnout87:
during test comunication

### DIFF
--- a/cdci_data_analysis/analysis/instrument.py
+++ b/cdci_data_analysis/analysis/instrument.py
@@ -299,9 +299,23 @@ class Instrument:
                     logger.warning("bad request from user, passing through: %s", e)
                     raise
                 except Exception as e: # we shall not do that
+                    message_prepend_str = ('Unfortunately, the analysis product you requested is empty, '
+                                           'since there is no usable data for the parameter combination you requested: '
+                                           'time period, software version, etc. We did not find any meaningful exceptions.\n'
+                                           'This additional message may be helpful:\n\n')
+
+                    e_message = None
+                    if hasattr(e, 'message') and e.message is not None:
+                        e_message = e.message
+
                     logger.error("run_query failed: %s", e)
                     # logger.error("run_query failed: %s", traceback.format_exc())
-                    query_out.set_failed(product_type, logger=logger, sentry_client=sentry_client, excep=e)
+                    query_out.set_failed(product_type,
+                                         e_message=e_message,
+                                         message_prepend_str=message_prepend_str,
+                                         logger=logger,
+                                         sentry_client=sentry_client,
+                                         excep=e)
 
         # adding query parameters to final products
         # TODO: this can be misleading since it's the parameters actually used
--- a/cdci_data_analysis/analysis/instrument.py
+++ b/cdci_data_analysis/analysis/instrument.py
@@ -535,9 +535,15 @@ class Instrument:
             else:
                 catalog_dic = json.loads(par_dic['selected_catalog'])
                 self.set_par('user_catalog', build_catalog(catalog_dic))
-
+        # setting user_catalog in the par_dic, either loading it from the file or aas an object
         if user_catalog_file is not None:
-            self.set_par('user_catalog', load_user_catalog(user_catalog_file))
+            catalog_object = load_user_catalog(user_catalog_file)
+            self.set_par('user_catalog', catalog_object)
+            # TODO set selected_catalog ? in this way it will show up in the email url link
+            # and used consistently for the job_id generation
+            self.set_par('selected_catalog', json.dumps(catalog_object.get_dictionary()))
+            # TODO verify that the file path is not needed in the frontend (should not)
+            par_dic.pop('user_catalog_file', None)
         else:
             if 'catalog_selected_objects' in par_dic.keys():
                 catalog_selected_objects = np.array(par_dic['catalog_selected_objects'].split(','), dtype=np.int)
--- a/cdci_data_analysis/analysis/instrument.py
+++ b/cdci_data_analysis/analysis/instrument.py
@@ -162,7 +162,7 @@ class Instrument:
 
         return p
 
-    def test_communication(self,config,logger=None):
+    def test_communication(self, config, logger=None):
         if self.data_server_query_class is not None:
             return self.data_server_query_class(config=config, instrument=self).test_communication(logger=logger)
         else:
--- a/cdci_data_analysis/analysis/instrument.py
+++ b/cdci_data_analysis/analysis/instrument.py
@@ -38,7 +38,7 @@ from .catalog import BasicCatalog
 from .products import QueryOutput
 from .queries import ProductQuery, SourceQuery, InstrumentQuery
 from .io_helper import upload_file
-from .exceptions import RequestNotUnderstood, RequestNotAuthorized
+from .exceptions import RequestNotUnderstood, RequestNotAuthorized, InternalError
 
 __author__ = "Andrea Tramacere"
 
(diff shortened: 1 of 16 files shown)


cdci_data_analysis/analysis/queries.py
```diff
@@ -388,13 +386,17 @@ def test_communication(self, instrument, query_type='Real', logger=None, config=
...
                                  logger=logger,
                                  sentry_client=sentry_client,
                                  excep=e,
                                  e_message=e_message,
                                  debug_message=debug_message)
 
+        except Exception as e:
+            e_message = f'test of communication with backend (instrument: {instrument.name}, product: {self.name}) failed!\n' + repr(e)
+
+            raise BadRequest(e_message)
```
volodymyrss, line 398: this is probably not a bad request, but a bad internal server error.

cdci_data_analysis/analysis/instrument.py
```diff
@@ -299,9 +299,23 @@ def run_query(self, product_type,
                     logger.warning("bad request from user, passing through: %s", e)
                     raise
                 except Exception as e: # we shall not do that
+                    message_prepend_str = ('Unfortunately, the analysis product you requested is empty, '
```
volodymyrss, line 302: this should be returned for a subset of exceptions, deriving from `UnfortunateRequestResults`: this means request lead to something apparently unfortunate. It's not bad, just empty, as it should be. Nothing wrong with that! Let's not shame users if their requests lead to nothing! It should be advised in the message that the user may want to consider another rrequest, jus tas written below.

In addition, we should also catch `BadRequest`. This means something is inconsistent, not interprettable by us in the request: e.g. parameter does not comply with format, out of range, etc.

Finally, other exceptions should be like `InternalError`. This should raise sentry issue and advise user that we are looking into it, it is worth retrying , and it might get better.

cdci_data_analysis/analysis/queries.py
```diff
@@ -375,11 +375,9 @@ def test_communication(self, instrument, query_type='Real', logger=None, config=
             query_out.set_done(message=message, debug_message=str(debug_message),status=status)
 
         except ConnectionError as e:
```
volodymyrss, line 377: I am most uncertain about this `ConnectionError`. I guess it never raises. Here, only `InternalError` may happen: the backend is either reachable or not. As of now, it may be caught as `Exception`, since they all are to be treated similarly: sentry and advise to the user to retry and we are working.

cdci_data_analysis/analysis/queries.py
volodymyrss, line 398: it's not user's fault! Bad server, with Internal Error.

cdci_data_analysis/analysis/queries.py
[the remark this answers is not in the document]
burnout87: Perhaps internally in some other plugin? Or perhaps those used to raise this exception before started using `DDAException`.

`UnfortunateRequestResults` is indeed thew one I had started intercepting, but was unsure was the right approach

cdci_data_analysis/analysis/queries.py
burnout87: Which exception is the most appropriate?

volodymyrss (changes requested):

cdci_data_analysis/analysis/queries.py
```diff
@@ -374,12 +374,15 @@ def test_communication(self, instrument, query_type='Real', logger=None, config=
 
             query_out.set_done(message=message, debug_message=str(debug_message),status=status)
 
-        except ConnectionError as e:
-            e_message = f'test of communication with backend (instrument: {instrument}) failed!'
-            e_message += "\n" + repr(e)
+        except (ConnectionError, UnfortunateRequestResults) as e:
```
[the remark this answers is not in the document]
volodymyrss: what's unclear to me is how cat `ConnectionError` ever arise from [cdci_osa_plugin/osa_dataserve_dispatcher.py#L243-L264](https://github.com/oda-hub/dispatcher-plugin-integral/blob/5c65f89fd0f2bab0ff3bdcf6beff2eb67ce5ac75/cdci_osa_plugin/osa_dataserve_dispatcher.py#L243-L264) which catches `Exception` and raises `DDAException`. No `ConnectionError` may come from this, it seems. Maybe a test would be helpful?

cdci_data_analysis/analysis/queries.py
volodymyrss: Maybe we do not have one. Actually I added one here https://github.com/oda-hub/dispatcher-app/blob/5c69e0f02a30bc2ab10a9206a6f1e68f1454f5ce/cdci_data_analysis/analysis/exceptions.py#L25-L27)

cdci_data_analysis/analysis/queries.py
volodymyrss: A bit dangerous to get rid without a test! But yeah, this only happens in exceptional cases. So a change here would not break too much. But still, would be nice to have a test.

cdci_data_analysis/analysis/queries.py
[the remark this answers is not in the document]
burnout87: Hence also my confusion. My interpretation was indeed that this is some legacy code.

cdci_data_analysis/analysis/queries.py
volodymyrss: It may be a lot to do right now. But maybe not. Could you try? So it will fit in the next patch, it should not be more than couple of days. 
If too long, a more minimal approach would suffice.

cdci_data_analysis/analysis/queries.py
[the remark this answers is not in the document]
volodymyrss: Yeah, and there are two ways to propagate exception:
* with pythonic  exceptions from [there](https://github.com/oda-hub/dispatcher-app/blob/master/cdci_data_analysis/analysis/exceptions.py)
* with dispatcher-y way in `query_out` set with `set_failed`.

Ideally, I would think it's better to stick to pythonic way and `set_failed` only once near the end.

cdci_data_analysis/analysis/queries.py
burnout87: So, the goal of the test is to verify that `ConnectionError` is never raised? Is every plugin implementing the `test_communication` differently?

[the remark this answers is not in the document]
volodymyrss: > Perhaps internally in some other plugin? Or perhaps those used to raise this exception before started using `DDAException`.
> 
> `UnfortunateRequestResults` is indeed thew one I had started intercepting, but was unsure was the right approach

It's right to intercept it.
But it's important to also intercept other ones.

cdci_data_analysis/analysis/queries.py
burnout87: Would be good to actually use this one.

cdci_data_analysis/analysis/queries.py
volodymyrss: There is no guarantee that what is raised is `ConnectionError`. Certainly it never happens with INTEGRAL plugin. Anyway, no matter what is raised there it is an internal error. Is there really any other possibility in testing communication? There can not be any bad request, and the result can not be unusual but acceptable.

volodymyrss:
Ok, so on [the last meeting](https://github.com/oda-hub/meetings/blob/main/2021-11-02/minutes.md) we agreed to postpone refactoring, and just make a very generic message like:

```
Your request produced an unusual result. It might not be what you expected.
It is possible that particular parameter selection should indeed lead to this outcome (e.g. there is no usable data). Please look carefully on your request.

It is also possible that the platform experienced a temporary issue. We aim distinguish all of such issues and report them clearly, but for now, we unfortunately can not be certain all cases like this are detected.
We try to discover on our own and directly address any temporary issue. But some issues might slip past us.
If you are willing to help us, please use "provide feedback" button below. We would greatly appreciate it!
```

volodymyrss:
So we finish this or postpone?

burnout87:
> So we finish this or postpone?

There is also #254 to take care of. But I just saw that this is a more "urgent" milestone

volodymyrss:
Yeah, we also agreed  on the meeting to wrap this up for now to make some sort of sense with list longer message and minimal (or None) functional changes, and then do refactoring in the future.

burnout87:
> Yeah, we also agreed on the meeting to wrap this up for now to make some sort of sense with list longer message and minimal (or None) functional changes, and then do refactoring in the future.

Yes, just got confused on which hotfix should be handled first.

volodymyrss:
> > Yeah, we also agreed on the meeting to wrap this up for now to make some sort of sense with list longer message and minimal (or None) functional changes, and then do refactoring in the future.
> 
> Yes, just got confused on which hotfix should be handled first.

but do we have [any more hotfixes really](https://github.com/issues?q=is%3Aopen+is%3Aissue+label%3Ahotfix+org%3Aoda-hub)?  all gone!

burnout87:
> > > Yeah, we also agreed on the meeting to wrap this up for now to make some sort of sense with list longer message and minimal (or None) functional changes, and then do refactoring in the future.
> > 
> > 
> > Yes, just got confused on which hotfix should be handled first.
> 
> but do we have [any more hotfixes really](https://github.com/issues?q=is%3Aopen+is%3Aissue+label%3Ahotfix+org%3Aoda-hub)? all gone!

Was #227 ever a hotfix? Perhaps we missed it, or I just got confused...

volodymyrss:
> > > > Yeah, we also agreed on the meeting to wrap this up for now to make some sort of sense with list longer message and minimal (or None) functional changes, and then do refactoring in the future.
> > > 
> > > 
> > > Yes, just got confused on which hotfix should be handled first.
> > 
> > 
> > but do we have [any more hotfixes really](https://github.com/issues?q=is%3Aopen+is%3Aissue+label%3Ahotfix+org%3Aoda-hub)? all gone!
> 
> Was #227 ever a hotfix? Perhaps we missed it, or I just got confused...

It was not. If it was, you could see the label changes in the issue. Also it was created before the release: there were no hotfixes at that time. It's a bug, but it's not the most critical one. We had to quickly address features for the release. Then hotfixes took priority. After this last issue here, we can make next patch release. And proceed with #227 for 21.10.0003

volodymyrss:
> With the new more general message we agreed:
> 
> image

I like this message. Too bad it is too generic. But that's life.

Oh, and the extra message, which is now on top, could it be at the end, after:

`This additional information might help:`

volodymyrss (approved):

burnout87:
> Oh, and the extra message, which is now on top, could it be at the end, after:
> 
> `This additional information might help:`

Like it was here?

[https://github.com/oda-hub/dispatcher-app/issues/261#issuecomment-948589009](https://github.com/oda-hub/dispatcher-app/issues/261#issuecomment-948589009)

volodymyrss:
> > Oh, and the extra message, which is now on top, could it be at the end, after:
> > `This additional information might help:`
> 
> Like it was here?
> 
> [#261 (comment)](https://github.com/oda-hub/dispatcher-app/issues/261#issuecomment-948589009)

Yes, but with the message in this issue. Only the order like there.

burnout87:
Temporary final version

image

volodymyrss:
> Temporary final version
> 
> image

Looks good (for now)! I also looked on other fixes in [staging](https://frontend-staging.obsuks1.unige.ch/mmoda) seems ok, as expected.

So if this is done we can set 21.10.0002 tonight. I will make very small patch release notes and add a banner. Well maybe tomorrow.

burnout87:
> 
> So if this is done we can set 21.10.0002 tonight. I will make very small patch release notes and add a banner. Well maybe tomorrow.

I will merge now, so I guess you can go for it.

Issue #259, opened 2021-10-09
creating of `temp_sid_XXXXX` is not thread-safe


volodymyrss:
I get this message in RequestNotUnderstood.

this is log

```python
ERROR:[ InstrumentQueryBackEnd : instrument-not-set ]:traceback: Traceback (most recent call last):
  File "/pyenv/versions/3.8.5/lib/python3.8/site-packages/cdci_data_analysis/analysis/io_helper.py", line 168, in mkdir
    self._file_path.mkdir()
  File "/pyenv/versions/3.8.5/lib/python3.8/pathlib.py", line 1284, in mkdir
    self._accessor.mkdir(self, mode)
FileExistsError: [Errno 17] File exists: 'temp_sid_Y0CQ65ZT4WN6ISIG'

During handling of the above exception, another exception occurred:
APIerror RequestNotUnderstood() constructor: [ InstrumentQueryBackEnd : isgri ] constructor failed: ('!Error error ', FileExistsError(17, 'File exists'), 'in creating dir', 'temp_sid_Y0CQ65Z
T4WN6ISIG')

Traceback (most recent call last):
  File "/pyenv/versions/3.8.5/lib/python3.8/site-packages/cdci_data_analysis/flask_app/dispatcher_query.py", line 234, in __init__
    self.set_temp_dir(self.par_dic['session_id'], verbose=verbose)
  File "/pyenv/versions/3.8.5/lib/python3.8/site-packages/cdci_data_analysis/flask_app/dispatcher_query.py", line 530, in set_temp_dir
    td.mkdir()
  File "/pyenv/versions/3.8.5/lib/python3.8/site-packages/cdci_data_analysis/analysis/io_helper.py", line 170, in mkdir
    raise RuntimeError('!Error error ',e,'in creating dir', self.path)
RuntimeError: ('!Error error ', FileExistsError(17, 'File exists'), 'in creating dir', 'temp_sid_Y0CQ65ZT4WN6ISIG')
```

it fails when there are multiple dispatcher processes, preventing scaling. I think it also fails with one process but many threads, but less often.

I reduce number of processes for now, and make a hotfix

volodymyrss:
I did not fully trace why the directory creation is conflicting. I am mildly concerned simple change will introduce issues. Right now, I will just reduce parallelism. Unfortunately, at high load (as I am doing now), with many catalogs, it seems to happens.

[...]

PR #269, opened 2021-10-29
hotfix-259 - safe creation of temp dir

burnout87:
some test adaption is needed

### DIFF
--- a/cdci_data_analysis/flask_app/dispatcher_query.py
+++ b/cdci_data_analysis/flask_app/dispatcher_query.py
@@ -526,11 +526,8 @@ class InstrumentQueryBackEnd:
         if job_id is not None:
             suffix += '_jid_'+job_id
 
-        td = tempfile.mkdtemp(suffix = suffix)
-
-        td = FilePath(file_dir=td)
-        td.mkdir()
-        self.temp_dir = td.path
+        td = tempfile.mkdtemp(suffix=suffix)
+        self.temp_dir = td
 
     def move_temp_content(self):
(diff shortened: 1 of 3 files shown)


cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -518,17 +518,16 @@ def set_temp_dir(self, session_id, job_id=None, verbose=False):
...
+            suffix += '_sid_' + session_id
 
         if job_id is not None:
-            td += '_jid_'+job_id
+            suffix += '_jid_'+job_id
 
-        td = FilePath(file_dir=td)
-        td.mkdir()
-        self.temp_dir = td.path
+        td = tempfile.mkdtemp(suffix=suffix)
```
volodymyrss, line 529: yeah, since this creates unique temp dir, it will not be possible to find it in the test without special effort.

volodymyrss (approved):

volodymyrss:
> > With the new approach for the creation of the `temp_dir` a test was no longer working, and it's actually good.
> > What would be interesting is to put the dispatcher under "stress" in order to simulate the situation that led to the issue described in #259, and verify that this no longer occurs.
> > Is there already a similar test for doing so?
> 
> Any ideas on how to tackle this?

it's not even just the stress: temp directory was created just with `job_id` and `session_id`, no randomness. So if two requests for `job_id` and `session_id` are at about the same time, one of the threads/processes will try to delete the directory where another one has written. It can be emulated but it's tricky.

PR #270, opened 2021-10-29
Long api code attached email

burnout87:


volodymyrss:
I will look on it when possible. To be clear, we can not really deploy it before [hotfixes](https://github.com/issues?q=is%3Aopen+archived%3Afalse++org%3Aoda-hub+label%3Ahotfix) are implemented. So while it is possible to merge it before, it may delay hotfix.

burnout87:
> I will look on it when possible. To be clear, we can not really deploy it before [hotfixes](https://github.com/issues?q=is%3Aopen+archived%3Afalse++org%3Aoda-hub+label%3Ahotfix) are implemented. So while it is possible to merge it before, it may delay hotfix.

I will convert it back to draft. Since changes that we apply with the hotfixes might have an impact here.

cdci_data_analysis/analysis/email_helper.py
```diff
@@ -173,9 +175,18 @@ def send_email(
...
-        pass
+        attachment_file_path = store_email_api_code_attachment(api_code, status, scratch_dir, sending_time=sending_time)
+        with open(attachment_file_path, "r") as fil:
+            api_code_email_attachment = MIMEApplication(
+                fil.read(),
+                Name=os.path.basename(attachment_file_path)
+            )
+        api_code_email_attachment.add_header('Content-Disposition',
+                                             'attachment',
+                                             filename="api code")
```
volodymyrss, line 189:
```suggestion
filename="api_code.py")
```

cdci_data_analysis/flask_app/templates/email.html
```diff
@@ -57,8 +57,9 @@
 To learn how to get a new token, please refer to <a href="{{ oda_site.manual_reference }}">these instructions</a>.<br>
 <br>
 {% else %}
-Please note that we were not able to embed API code in this email, since exceptionally, it was too large. 
+Please note that we were not able to embed API code in this email, since exceptionally, it was too large.
 Our support team has been alerted about this incident.
+However, a python script with the api code has been attached to this email.
```
volodymyrss, line 62: It's not really that problematic anymore. Just less favorable. Also I do not think we get an alert?

```suggestion
Please note the API code for this query was too large to embed it in the email text. Instead, we attach it as a python script.
```

tests/test_job_management.py
```diff
@@ -121,6 +121,15 @@ def test_public_async_request(dispatcher_live_fixture, dispatcher_local_mail_ser
     r'expire in .*? .*?\.'
 ]
 
+ignore_api_code_patterns = [
```
volodymyrss, line 124: maybe `api_attachment_attachment_patterns` or something like that

volodymyrss (changes requested):
Small comments

volodymyrss (approved):

PR #277, opened 2021-11-12
posting result execution to the product gallery

burnout87:


burnout87:
Now also the image upload is supported

volodymyrss:
are you uploading some dummy results with this into [here](http://cdciweb02.isdc.unige.ch/mmoda-pg/data-products-gallery) ? very nice!

burnout87:
> are you uploading some dummy results with this into [here](http://cdciweb02.isdc.unige.ch/mmoda-pg/data-products-gallery) ? very nice!

Exactly! And they come directly from the `oda_api`

burnout87:
> are you uploading some dummy results with this into [here](http://cdciweb02.isdc.unige.ch/mmoda-pg/data-products-gallery) ? very nice!

BTW I saw your view for the data-products is blocked, can I remove the block and apply some css class? they are very simple adjustments

volodymyrss:
> > are you uploading some dummy results with this into [here](http://cdciweb02.isdc.unige.ch/mmoda-pg/data-products-gallery) ? very nice!
> 
> Exactly! And they come directly from the `oda_api`

I notice the change there too, cool! 
By the way, titles probably could be part supplied by user, but also in part autogenerated but with considerable attention. Probably even both titles are useful: user-provided and auto-generated.

For the view - please change absolutely anything there. I did it very quickly just to see what is possible in drupal. It's just better if we do not both edit the same view. So if you want - take over this one, I will not touch it anymore. If I want to do further experiments - I make another one.

burnout87:
> 



> > > are you uploading some dummy results with this into [here](http://cdciweb02.isdc.unige.ch/mmoda-pg/data-products-gallery) ? very nice!
> > 
> > 
> 
> I notice the change there too, cool! By the way, titles probably could be part supplied by user, but also in part autogenerated but with considerable attention. Probably even both titles are useful: user-provided and auto-generated.

Ok, I will combine both.

> 
> For the view - please change absolutely anything there. I did it very quickly just to see what is possible in drupal. It's just better if we do not both edit the same view. So if you want - take over this one, I will not touch it anymore. If I want to do further experiments - I make another one.

Thanks! Check now after having applied those classes I had i mind!

burnout87:
Ok, just verified that it is possible to specify the author of the content posted, this confirms  that what we discussed can be done!

So, my idea is to extract the email address from the token (mmoda/oda-api-token) and use it to retrieve the related user-id in the product gallery: if there is a match, then use it to specify the content (eg data-product) author.

cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+class ContentType(Enum):
+    ARTICLE = auto()
+    DATA_PRODUCT = auto()
+    OBSERVATION = auto()
+    ASTROPHYSICAL_ENTITY = auto()
+
+
+def discover_mmoda_pg_token():
```
volodymyrss, line 22: this should come from dispatcher config

cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -0,0 +1,193 @@
...
+    bytes_img = img.read()
+    b_64_img = base64.b64encode(bytes_img).decode("utf8")
+    img_name = img.filename
+    img_extension = os.path.splitext(img_name)[1][1:]
+
+    body_post_img["data"][0]["value"] = b_64_img
+    body_post_img["uri"][0]["value"] = "public://" + img_name
+    body_post_img["filename"][0]["value"] = img_name
+    body_post_img["filemime"]["value"] = "image/" + img_extension
+    body_post_img["_links"]["type"]["href"] = "http://cdciweb02.isdc.unige.ch/mmoda-pg/rest/type/file/image"
```
volodymyrss, line 61: the URL base should from dispatcher config too

cdci_data_analysis/analysis/drupal_helper.py
burnout87: yes, given what we agreed with the modification of the author of the content

volodymyrss: Not even because of that. 
Even before modification: this file you'd find locally has to belong to dispatcher.
Maybe there is some misunderstanding of how this is to be used: it will run in the dispatcher production, right?

burnout87: > Maybe there is some misunderstanding of how this is to be used: it will run in the dispatcher production, right?

no misunderstanding, it's clear

cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -0,0 +1,367 @@
...
+        }
+        # post the article
+        log_res = requests.get(f"{product_gallery_url}/observations/{observation_id}?_format=hal_json",
+                               headers=headers
+                               )
+        output_get = log_res.json()
+        if log_res.status_code < 200 or log_res.status_code >= 300:
+            raise RequestNotUnderstood(output_get['message'],
+                                       status_code=log_res.status_code,
+                                       payload={'error_message': 'error while retrieving the user id'})
```
volodymyrss, line 220: message not right.

by the way, I suspect the reason why the message and the comments are occasionally not right is because it is copied and modified :)
So if the repetition was avoided, by putting the authorization and the error detection in a function, it would prevent this cases. But really as you wish!

cdci_data_analysis/flask_app/templates/body_article_product_gallery.py
```diff
@@ -0,0 +1,43 @@
+body_article = {
+    "_links": {
+        "type": {
+            "href": "http://cdciweb02.isdc.unige.ch/mmoda-pg/rest/type/node/"
```
volodymyrss, line 4: maybe we could put some generic url here to avoid showing with our internal stuff

cdci_data_analysis/config_dir/conf_env.yml.example
```diff
@@ -54,5 +54,11 @@ dispatcher:
         # for when a request has been submitted
         email_sending_job_submitted_default_interval: 5
 
+    # product gallery related set of options
+    product_gallery_options:
+        # url of the product gallery home
+        product_gallery_url: "http://cdciweb02.isdc.unige.ch/mmoda-pg/"
```
volodymyrss, line 60: maybe let's replace internal url with something

cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -0,0 +1,367 @@
...
+    # set the type of content to post
+    link_content_type = body_gallery_article_node["_links"]["type"]["href"] + 'data_product'
+    body_gallery_article_node["_links"]["type"]["href"] = link_content_type
+
+    # set the initial body content
+    body_value = ''
+    product_type = ''
+    t1 = t2 = None
+    # get products
+    scratch_dir_json_fn = f'scratch_sid_{session_id}_jid_{job_id}'
```
volodymyrss, line 261: that's just `scratch_dir`

cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -0,0 +1,367 @@
...
+        t2 = analysis_parameters_json_content_original.pop('T2')
+
+        # TODO no need to set all the parameters by default
+        # for k, v in analysis_parameters_json_content_original.items():
+        #     # assuming the name of the field in drupal starts always with field_
+        #     field_name = str.lower('field_' + k)
+        #     body_gallery_article_node[field_name] = [{
+        #         "value": v
+        #     }]
+        body_value = (f'''''')
```
volodymyrss, line 287: `body_value = ''` ?

cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -0,0 +1,367 @@
...
+    else:
+        raise RequestNotUnderstood(message="Request data ont found",
```
volodymyrss, line 289: `not found`

volodymyrss: could it be something more descriptive, like "Results of the ODA product request could not be found. Perhaps wrong job_id was passed?". Else we have plenty of very different `requests` around: request as sent now, request to drupal, etc.

cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -0,0 +1,367 @@
...
+                                   payload={'error_message': 'error while posting article'})
+
+    # set observation
+    if 'T1' in kwargs:
+        t1 = kwargs.pop('T1')
+    if 'T2' in kwargs:
+        t2 = kwargs.pop('T2')
+
+    observation_drupal_id = get_observation_drupal_id(product_gallery_url, jwt_token,
```
volodymyrss, line 298: it would be quite useful for the user to know that the observation was derived automatically. Could it be returned as a comment?

cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -0,0 +1,367 @@
...
+                                                      t1=t1, t2=t2, observation_id=observation_id)
+    body_gallery_article_node["field_derived_from_observation"] = [{
+        "target_id": observation_drupal_id
+    }]
+
+    # let's go through the kwargs and if any overwrite some values for the product to post
```
volodymyrss, line 304: nice!

tests/test_server_basic.py
```diff
@@ -1393,3 +1394,60 @@ def test_get_query_products_exception(dispatcher_live_fixture):
...
+def test_product_gallery_post_article(dispatcher_live_fixture):
+    server = dispatcher_live_fixture
+
+    logger.info("constructed server: %s", server)
+
+    # send simple request
+    # let's generate a valid token
+    token_payload = {
+        **default_token_payload,
+        "roles": "general, unige-hpc-full",
```
volodymyrss, line 1409: we should add a role to regulate gallery posting. something like `post-product-gallery`

tests/test_server_basic.py
```diff
@@ -1393,3 +1394,60 @@ def test_get_query_products_exception(dispatcher_live_fixture):
     print("jdata : ", jdata)
 
     assert jdata['exit_status']['message'] == 'InternalError()\nfailing query\n'
+
+
+@pytest.mark.test_drupal
+def test_product_gallery_post_article(dispatcher_live_fixture):
```
volodymyrss, line 1400: there are, of course, plenty of other cases, but maybe it's not worth effort to test them, as you wish.

cdci_data_analysis/flask_app/app.py
```diff
@@ -392,6 +393,42 @@ def get(self, path):
...
+        decoded_token = tokenHelper.get_decoded_token(token, secret_key)
+        logger.info("==> token %s", decoded_token)
+    except jwt.exceptions.ExpiredSignatureError:
+        # raise RequestNotAuthorized("The token provided is expired.")
+        return make_response('The token provided is expired.'), 403
+    except jwt.exceptions.InvalidTokenError:
+        # raise RequestNotAuthorized("The token provided is not valid.")
+        return make_response('The token provided is not valid.'), 403
+
+    jwt_pg_token = drupal_helper.get_mmoda_pg_token(app_config.product_gallery_jwt_token_location)
```
volodymyrss, line 416: so this token is limited lifetime, right? How do we update it in production? Did you manage to make it live longer? 
If not, we could also decode it with pyjwt, change, and encode back. If we can get the secret key from drupal (I assume it's possible).

In any case, would be nice to have a very simple endpoint to write this token. Since we will have to update it sooner or later.

volodymyrss (changes requested):
Thanks! Just a couple of small requests (e.g. on token lifetime) and plenty of small suggestions which you may as well ignore.

volodymyrss:
Oh, and we should not forget to account for visibility of the products. We might want to start adding private products very soon. And we also want to control what is ready to be shown (can be addressed with visibility too)

tests/test_server_basic.py
burnout87: As agreed in [https://github.com/oda-hub/doc-multi-user/pull/9](https://github.com/oda-hub/doc-multi-user/pull/9), this will be `gallery poster`

cdci_data_analysis/flask_app/templates/body_article_product_gallery.py
burnout87: this will actually be extracted from the configuration, so it will go

cdci_data_analysis/flask_app/app.py
burnout87: Investigation with @motame will perhaps help to better understand how to tackle this issue.

As also agreed, an  endpoint will also help.

volodymyrss: > Investigation with @motame will perhaps help to better understand how to tackle this issue.

Do try to decode the token, it's base64 json (part of it).

If it changes with new VPN, it might have host name or ip in it. Then it's no problem for production dispatcher, since it never changes of course.

> As also agreed, an endpoint will also help.

It will likely expire anyway. So an endpoint is critical. 
It will also help in short term no matter what are the other issues.

burnout87: Just verified that the validity of the token issued is 1 hour

cdci_data_analysis/analysis/drupal_helper.py
burnout87: good idea, will put a `TODO` in case I don't manage to implement it now

cdci_data_analysis/flask_app/app.py
volodymyrss: > Just verified: the validity of the token issued is 1 hour

if you have the secret key and the protocol, you might be able to produce (maybe easiest modify starting from existing) the token locally with `pyjwt`. And set whatever duration.

tests/test_server_basic.py
volodymyrss: I am not too happy with `poster` since it's also a noun but why not, not a big deal.

cdci_data_analysis/flask_app/app.py
burnout87: > 
> if you have the secret key and the protocol, you might be able to produce (maybe easiest modify starting from existing) the token locally with `pyjwt`. And set whatever duration.

indeed what I am doing

tests/test_server_basic.py
burnout87: we still have time to change to something else [https://github.com/oda-hub/doc-multi-user/pull/9](https://github.com/oda-hub/doc-multi-user/pull/9) is not closed yet

tests/test_server_basic.py
burnout87: indeed I was not planning to test them all at this stage

cdci_data_analysis/flask_app/app.py
volodymyrss: hm, it's really what I meant. I thought the token would be updated outside of dispatcher (with drupal or otherwise) and uploaded to it?

If we trust the dispatcher with the secret key, why not generate pg-token with every request? No need to even store it in a file?

burnout87: ok then in that case, there's no  need to take care of extending the validity of the token, am I right?

At every request a fresh new token will be generated, and the only secret information stored will be the key.

volodymyrss: yeah, if secret key is available, and tokens generated like make sense to drupal, this is sufficient. No need for an additional endpoint. 
We'll update the secret infrequently, and hence possible to do it on deployment.

volodymyrss (changes requested):
as in the comments for the token/secret situation

burnout87:
> as in the comments for the token/secret situation

a new token is now generated at each request, and the secret_key is stored in the configuration

cdci_data_analysis/flask_app/app.py
burnout87: a new token is now generated at each request, and the secret_key is stored in the configuration

volodymyrss (dismissed):

volodymyrss:
I imagine you might have trouble to test it given that there is no gallery fixture. I would make something like monkeypatch requests, but it's of limited used in such a case at best.

burnout87:
> I imagine you might have trouble to test it given that there is no gallery fixture. I would make something like monkeypatch requests, but it's of limited used in such a case at best.
> 
> So, unless some solution can be found quickly, I suggest to merge it. Would be cool to have it this week. I'll need the secret too to insert.

Partially related to that, testing it was not easy and so straightforward. But I am just waiting for all the checks to complete before merging

volodymyrss (approved):

volodymyrss:
> > I imagine you might have trouble to test it given that there is no gallery fixture. I would make something like monkeypatch requests, but it's of limited used in such a case at best.
> > So, unless some solution can be found quickly, I suggest to merge it. Would be cool to have it this week. I'll need the secret too to insert.
> 
> Partially related to that, testing it was not easy and so straightforward. But I am just waiting for all the checks to complete before merging

Check on coverage might fail. Please merge anyway if possible.

Actually I am reasonably certain it would be possible to setup a side container as [here](https://github.com/oda-hub/dispatcher-plugin-antares/blob/a5fd5e9d885c8b1500af46f2a6a080fdf4b37d9a/.github/workflows/python-app.yml#L18-L22) with just drupal, preset with some default stuff. Maybe later?

On the bright side, if we only use dispatcher as relay between oda_api and drupal pg, we do not need very many changes.

Once you merge, it should appear (in few min) in staging https://frontend.staging.odahub.io/mmoda/dispatch-data/.

burnout87:
> 
> Check on coverage might fail. Please merge anyway if possible.

Waiting for the last check to complete, then will be able to do so.

> 
> Actually I am reasonably certain it would be possible to setup a side container as [here](https://github.com/oda-hub/dispatcher-plugin-antares/blob/a5fd5e9d885c8b1500af46f2a6a080fdf4b37d9a/.github/workflows/python-app.yml#L18-L22) with just drupal, preset with some default stuff. Maybe later?

Yes, I would say later, but I am really curious in having a more id-depth view on the container configuration and run, and this could be a really good exercise I could try to do.

> 
> On the bright side, if we only use dispatcher as relay between oda_api and drupal pg, we do not need very many changes.

Indeed probably not, only very minor ones will be required from time to time.

PR #280, opened 2021-11-12
Parameters default format

burnout87:


burnout87:
For now, just `Time` and `SpectralBoundary` are dealt with, for the other parameters default format needs to be listed.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -297,6 +274,10 @@ def set_par(self, value, units=None):
         if units is not None:
             self.units = units
         self.value = value
+        # this should return the value in a default format
+        return self.get_value_default_format(value)
+
+    def get_value_default_format(self, value):
```
volodymyrss, line 280: maybe let's call it something like `get_value_in_default_format`?
Else it may sound like "get default format of value" or something.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -476,14 +457,11 @@ def __init__(self, value=None, T_format='isot', name=None, Time_format_name=None
...
 
-    def set_par(self, value, units=None):
-        # sets the value to the format initially specified in the form
-        if units is not None:
-            self.units = units
-        self.value = value
-        # return the relative isot format value
-        # afterwards also the units_name will be set to isot
-        return self._astropy_time.isot
+    def get_value_default_format(self, value):
```
volodymyrss, line 460: this is the kind of case I would use type hint, at least to str. 
Not even for just static analysis, for for explaining to ourselves our intent

```suggestion
    def get_value_default_format(self, value) -> str:
```

cdci_data_analysis/analysis/parameters.py
```diff
@@ -683,14 +661,14 @@ def __init__(self,value=None,E_units='keV',name=None):
         #wtform_dict = {'keV': FloatField}
 
         super(SpectralBoundary, self).__init__(value=value,
-                                   units=E_units,
-                                   check_value=self.check_energy_value,
-                                   name=name,
-                                   allowed_units=_allowed_units)
+                                               units=E_units,
+                                               check_value=self.check_energy_value,
+                                               name=name,allowed_units=_allowed_units)
```
volodymyrss, line 666: in two lines?

cdci_data_analysis/analysis/parameters.py
```diff
@@ -683,14 +661,14 @@ def __init__(self,value=None,E_units='keV',name=None):
...
                                    #wtform_dict=wtform_dict)
 
-
-
+    def get_value_default_format(self, value):
+        # might return int or float
+        return ast.literal_eval(value)
```
volodymyrss, line 671: this should be always float. But actually, this could be deduced from units in `Parameter`. Anything which has physical units is float.

Also there is `Float` class. This can inherit from `Float`. And almost all others, except for `Integer` - that one is int.

this sort of generalization will apply to most if not all cases.

volodymyrss (changes requested):

cdci_data_analysis/analysis/parameters.py
burnout87: I don't know why they're like this here on guthub...

volodymyrss: hm. They are not like this elsewhere? I did not comment on it on github: I review directly in vscode.
Of course, both are MS products, and MS has been recently in code synthesis, who knows!

burnout87: but I use pyCharm which is not evil-MS...

volodymyrss: I used it before but it's not eco-friendly, it burns  lot's of cpu power! For some reason.

burnout87:
I am having troubles also in #277 when installing the dependencies, specifically, it fails when installing `pathlib`.

Have you seen this happening before?

cdci_data_analysis/analysis/parameters.py
burnout87: Same indeed applies to `Energy` and `DetectionThreshold` classes.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -297,6 +274,10 @@ def set_par(self, value, units=None):
         if units is not None:
             self.units = units
         self.value = value
+        # this should return the value in a default format
```
volodymyrss, line 277: to me, this is the kind of comment which is make entirely redundant by carefully crafted method name :) It almost literally repeats the next line :)
No complains though, just wonder if this is how it also looks for you!

cdci_data_analysis/analysis/parameters.py
```diff
@@ -389,6 +371,17 @@ def value(self, v):
...
             self._v=None
 
+    def get_value_in_default_format(self, value):
+        # might return int or float
+        if type(value) == str:
+            value = ast.literal_eval(value)
+
+        if self.default_units == 'float':
+            value = float(value)
+        elif self.default_units == 'int':
```
volodymyrss, line 381: can it really be `int`, in `Float` class?
and what happens in `else` branch?

cdci_data_analysis/analysis/parameters.py
```diff
@@ -389,6 +371,17 @@ def value(self, v):
...
+            value = int(value)
+        return float(value)
```
volodymyrss, line 383: it seems as if this entire method can be reduced to just this return.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -476,14 +469,11 @@ def __init__(self, value=None, T_format='isot', name=None, Time_format_name=None
...
 
-    def set_par(self, value, units=None):
-        # sets the value to the format initially specified in the form
-        if units is not None:
-            self.units = units
-        self.value = value
-        # return the relative isot format value
-        # afterwards also the units_name will be set to isot
-        return self._astropy_time.isot
+    def get_value_in_default_format(self, value) -> str:
```
volodymyrss, line 472: it does not always return str. Currently, it would be `Union[str, float, None]`. But the method should be modified to never return `None`, else unclear what happens to it.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -476,14 +469,11 @@ def __init__(self, value=None, T_format='isot', name=None, Time_format_name=None
...
+        if self._default_units == 'isot':
+            return self._astropy_time.isot
+        elif self._default_units == 'mjd':
+            return self._astropy_time.mjd
```
volodymyrss, line 476: the whole method:

```suggestion
return getattr(self._astropy_time, self._default_units)
```

cdci_data_analysis/analysis/parameters.py
```diff
@@ -674,29 +664,24 @@ def _set_angle(self, value, units):
...
 
 
+class SpectralBoundary(Float):
+    def __init__(self, value=None, E_units='keV' ,name=None):
 
-class SpectralBoundary(Parameter):
-    def __init__(self,value=None,E_units='keV',name=None):
-
-        _allowed_units = ['keV','eV','MeV','GeV','TeV','Hz','MHz','GHz']
+        _allowed_units = ['keV', 'eV', 'MeV', 'GeV', 'TeV', 'Hz', 'MHz', 'GHz']
```
volodymyrss, line 670: there is some confusion between `type` and `units`. `SpectralBoundary` has type `float` but units e.g. `keV`.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -710,10 +695,10 @@ def check_energy_value(value, units=None,name=None):
             raise RuntimeError('type of ',name,'not valid',type(value))
 
 
-class Energy(Parameter):
+class Energy(Float):
```
volodymyrss, line 698: this is (almost? Or completely?) identical to `SpectralBoundary`

cdci_data_analysis/analysis/parameters.py
```diff
@@ -761,10 +739,8 @@ def __init__(self,value=None,units='sigma',name=None):
                                    name=name,
                                    allowed_units=_allowed_units)
                                    #wtform_dict=wtform_dict)
-
-
-
-
+    # TODO this method, behaves exaclty the same as the one for the Energy and SpectralBoundary classes
```
volodymyrss, line 742: this should be almost the same. It behaves like `Float`. The only difference is units (`sigma` in this case).

tests/test_job_management.py
```diff
@@ -1953,6 +1953,28 @@ def x(arg):
     assert len(my_globals['scwl_dict']['scw_list']) > max_length
 
 
+@pytest.mark.parametrize('sb_value', [25, 25., 25.64547871216879451687311211245117852145229614585985498212321])
+def test_spectral_parameter(dispatcher_live_fixture, sb_value):
```
volodymyrss, line 1957: it's vital to have a complete test like this!

but I'd also add a unit test simply constructing parameters and getting default values from them. Just to capture how we handle conversions.

volodymyrss (changes requested):

cdci_data_analysis/analysis/parameters.py
burnout87: Comments add (or at least they should add) clarity and explanations to the code, that's why I add them even though they might effectively be useless or not vital, like in this case. In this specific case, it probably is a leftover comment from a previous commit, just forgot to remove it.

cdci_data_analysis/analysis/parameters.py
burnout87: yes, I copied from another class, and forgot to reduce it, return value is sufficient

volodymyrss:
> I am having troubles also in #277 when installing the dependencies, specifically, it fails when installing `pathlib`.
> 
> Have you seen this happening before?

Have I seen troubles in python dependencies? A lot! In this package? Not really. But, feel free to avoid pathlib. It's just that it's used in renku so it's nice to align.

cdci_data_analysis/analysis/parameters.py
volodymyrss: comments are critical when the intent is not obvious!
But they are not an excuse for unclear code, when making code clear is a clear option, like it's already done here.

cdci_data_analysis/analysis/parameters.py
burnout87: when do you see this returning `None`?

cdci_data_analysis/analysis/parameters.py
burnout87: so perhaps we should rename it `default_format` from `default_units`?

cdci_data_analysis/analysis/parameters.py
burnout87: the only difference is the `_allowed_units` list, perhaps that's what really differentiate them

burnout87: also, Energy does not have a `units` set, should it have one?

burnout87:
> > I am having troubles also in #277 when installing the dependencies, specifically, it fails when installing `pathlib`.
> > Have you seen this happening before?
> 
> Have I seen troubles in python dependencies? A lot! In this package? Not really. But, feel free to avoid pathlib. It's just that it's used in renku so it's nice to align.

It's really weird why this happens!

cdci_data_analysis/analysis/parameters.py
volodymyrss: https://github.com/oda-hub/dispatcher-app/blob/f0ee9e2bdb08084a754e2f6d7297b066b4727da2/cdci_data_analysis/analysis/parameters.py#L472-L477

there is no `else`. If neither of the two conditions is satisfied it returns `None`. This might happen if the validation elsewhere does not succeed, or allowed formats are redefined in inherited class, or something like that.

In any case, the method, as it is, may end up there, which will cause an unclear error.
If `else` can only happen in the case of some usage error, it should raise and exception.

cdci_data_analysis/analysis/parameters.py
volodymyrss: yeah. Well I'd call it `type`: it's not exactly the same as format.
As it's inherited from `Float`, float should take care of the entire business of having parameter type. Float, in principle, may have no unit. But anything with unit would also be a float.

cdci_data_analysis/analysis/parameters.py
burnout87: yes, indeed I reduced the method as you suggest in the other comment

cdci_data_analysis/analysis/parameters.py
burnout87: Same, as for the `Float` Parameter, applies to `Integer`.

burnout87: > so perhaps we should rename it `default_format` from `default_units`?
>
Even better, having them both allows to have a unit as well as being a float (or integer).

volodymyrss: both `default_type` and `default_units` ? yes.

cdci_data_analysis/analysis/parameters.py
volodymyrss: it should!

burnout87: ok, which one should it be?

volodymyrss: allowed units should be the same. default units should be `keV`

tests/test_job_management.py
burnout87: test included!

cdci_data_analysis/analysis/instrument.py
```diff
@@ -138,19 +138,17 @@ def set_pars_from_dic(self, par_dic, verbose=False):
...
+            # or they don't have to set (eg scw_list)
             if par.name is not None and par.name not in params_not_to_be_included:
-                # set the value for par in to a default format,
+                # set the value for par to a default format,
                 # or to a default value if this is not included within the request
-                
                 updated_par_dic[par.name] = par.set_value_from_form(par_dic, verbose=verbose)
 
-                if par.default_units is not None:
+                if par.units_name is not None and par.default_units is not None:
```
volodymyrss, line 151: what if only one of these two conditions is satisfied? This would be a problem in constructing, e.g. in plugin. Will there be an exception when constructing?

cdci_data_analysis/analysis/parameters.py
```diff
@@ -437,6 +446,8 @@ def value(self, v):
         else:
             self._v=None
 
+    def get_value_in_default_format(self, value):
+        return int(value)
 
     @staticmethod
     def check_int_value(value, units=None,name=None):
```
volodymyrss, line 453: yeah, this function is not really checking for int. It accepts float, at least. And it does it in a peculiar way...

cdci_data_analysis/analysis/parameters.py
```diff
@@ -476,14 +487,8 @@ def __init__(self, value=None, T_format='isot', name=None, Time_format_name=None
...
-    def set_par(self, value, units=None):
-        # sets the value to the format initially specified in the form
-        if units is not None:
-            self.units = units
-        self.value = value
-        # return the relative isot format value
-        # afterwards also the units_name will be set to isot
-        return self._astropy_time.isot
+    def get_value_in_default_format(self, value) -> Union[str, float, None]:
+        return getattr(self._astropy_time, self.default_units)
```
volodymyrss, line 491: before we freese it, any change to rename it to `default_unit`? It's a single unit. As opposed to `allowed_units` where it's many.
It's a bit inconsistent throughout. But I guess at least `default_units` is not yet used in plugins?

cdci_data_analysis/analysis/parameters.py
```diff
@@ -674,29 +679,27 @@ def _set_angle(self, value, units):
 #
 
 
+class SpectralBoundary(Float):
```
volodymyrss, line 682: this should just inherit from `Energy`, changing nothing.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -674,29 +679,27 @@ def _set_angle(self, value, units):
...
+    def __init__(self, value=None, E_units='keV', name=None):
 
-class SpectralBoundary(Parameter):
-    def __init__(self,value=None,E_units='keV',name=None):
-
-        _allowed_units = ['keV','eV','MeV','GeV','TeV','Hz','MHz','GHz']
-
+        _allowed_units = ['keV', 'eV', 'MeV', 'GeV', 'TeV', 'Hz', 'MHz', 'GHz']
```
volodymyrss, line 685: this should be set in `Energy`

cdci_data_analysis/analysis/parameters.py
```diff
@@ -674,29 +679,27 @@ def _set_angle(self, value, units):
...
+                         units=E_units,
+                         default_units='keV',
+                         check_value=self.check_energy_value,
+                         name=name,
+                         allowed_units=_allowed_units)
+                           #wtform_dict=wtform_dict)
 
     @staticmethod
     def check_energy_value(value, units=None,name=None):
         #print('check type of ',name,'value', value, 'type',type(value))
```
volodymyrss, line 701: this method body is repeated multiple times. Can it be just kept in `Float` (with optimizations)?

cdci_data_analysis/analysis/instrument.py
burnout87: are you referring to a case of a `Parameter` that has no `default_units`? but it has a `units_name`?

well, yes, in that case yes probably an exception is necessary

cdci_data_analysis/analysis/parameters.py
```diff
@@ -745,14 +746,12 @@ def check_energy_value(value, units=None,name=None):
...
-
-
-class DetectionThreshold(Parameter):
+class DetectionThreshold(Float):
     def __init__(self,value=None,units='sigma',name=None):
 
         _allowed_units = ['sigma']
-
+        if units is None or units == '':
+            units = 'keV'
```
volodymyrss, line 754: `'sigma'`. Also let's set it to default units.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -761,10 +760,8 @@ def __init__(self,value=None,units='sigma',name=None):
                                    name=name,
                                    allowed_units=_allowed_units)
                                    #wtform_dict=wtform_dict)
-
-
-
-
+    # TODO this method, behaves exaclty the same as the one for the Energy and SpectralBoundary classes
```
volodymyrss, line 763: this method belongs  to float.

volodymyrss (changes requested):

cdci_data_analysis/analysis/instrument.py
volodymyrss: yes. And also dor the opposite (`default_units` but no `units_name`) - maybe a warning.

cdci_data_analysis/analysis/parameters.py
burnout87: > before we freese it, any change to rename it to `default_unit`? It's a single unit.
yes I guess it's not gonna be a problem

> But I guess at least `default_units` is not yet used in plugins?
It has been introduced with this PR so, no, I will check anyway

cdci_data_analysis/analysis/parameters.py
burnout87: so, extend the `_allowed_units` of the `Energy class?

cdci_data_analysis/analysis/parameters.py
burnout87: actually we could use `check_int_value` (and `check_float_value`) for `Float`

cdci_data_analysis/analysis/parameters.py
volodymyrss: yes please!

cdci_data_analysis/analysis/parameters.py
burnout87: so, actually I realized that I named it as `default_units` to be consistent with `units`

cdci_data_analysis/analysis/parameters.py
burnout87: indeed how I do now, inside Parameter directly

cdci_data_analysis/analysis/instrument.py
burnout87: is it really a problem if `units_name` is not specified for a `Parameter` ?

cdci_data_analysis/analysis/parameters.py
volodymyrss: why use `check_int_value` for `float`? We really just care that value can be cast to `float`. `int` can be safely cast. Not the opposite though: from float to int.

cdci_data_analysis/analysis/parameters.py
volodymyrss: > so, actually I realized that I named it as default_units to be consistent with units

Yeah that's what I meant by `inconsistent throughout`. Meh. Maybe let's add a TODO instead? Else need to preserve interface carefully.

burnout87: indeed, I would stick with this (wrong) approach, I will add the TODO

cdci_data_analysis/analysis/parameters.py
burnout87: I meant that we could use use `check_int_value` in this case (`Integer`), and `check_float_value` for `Float`.

Or I misunderstood you comment?

cdci_data_analysis/analysis/instrument.py
volodymyrss: it's  not a big problem, but it's probably unintended. It's good to add it. So a warning - not an exception.

cdci_data_analysis/analysis/parameters.py
volodymyrss: we could, but we need to adapt them.

right now, they are identical:

https://github.com/oda-hub/dispatcher-app/blob/d8b313db28c2de51fd90e4640064f2183a9437a1/cdci_data_analysis/analysis/parameters.py#L459-L474

https://github.com/oda-hub/dispatcher-app/blob/d8b313db28c2de51fd90e4640064f2183a9437a1/cdci_data_analysis/analysis/parameters.py#L405-L421

burnout87: the only difference is the conversion to a `np.float` / `np.int` object, hence my question: is that really necessary?

volodymyrss: it is really necessary to rewrite them according to our expectations!

burnout87: ok some differentiation from the methods can be made

but my question is specific to the creation of a numpy object (`np.float(value)`/`np.int(value)`) are those necessary? why they were there in first place?

burnout87: and BTW I also discovered that the usage of `np.float` and `np.int` is deprecated

[https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations)

and that `float` and `int` respectively can be used

cdci_data_analysis/analysis/parameters.py
burnout87: it will be there, it makes sense

cdci_data_analysis/analysis/parameters.py
```diff
@@ -382,30 +393,34 @@ def value(self):
...
     @staticmethod
     def check_float_value(value, units=None, name=None):
         #print('check type of ',name,'value', value, 'type',type(value))
-        if value is None or value=='':
+        if value is None or value == '':
             pass
         else:
             try:
-              value=ast.literal_eval(value)
+                value = ast.literal_eval(value)
```
volodymyrss, line 413: I do not understand at all what is the purpose of this. It's a very peculiar way to convert time type, effectively discarded by the rest of the method.

volodymyrss, line 413: I do not know why is this except needed?

cdci_data_analysis/analysis/parameters.py
```diff
@@ -382,30 +393,34 @@ def value(self):
...
-            value=np.float(value)
-            if type(value) == int or type(value) == np.int:
-                pass
-            elif type(value) == float or type(value) == np.float:
-                pass
-            else:
-                raise RuntimeError('type of ', name, 'not valid', type(value))
+            try:
+                value = float(value)
+                if type(value) == float or type(value) == np.float:
```
volodymyrss, line 418: `value` was just cast to `float`. This check is useless.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -382,30 +393,34 @@ def value(self):
...
+                    pass
+                else:
+                    raise
```
volodymyrss, line 421: raise what?

cdci_data_analysis/analysis/parameters.py
burnout87: I think it was there because, in case values in a string format (eg '1234') would be correctly converted ( 1234).

That's the explanation I give

cdci_data_analysis/analysis/parameters.py
burnout87: indeed, I realized it and removed it

cdci_data_analysis/analysis/parameters.py
burnout87: intermediate commit, should be gone by now

cdci_data_analysis/analysis/parameters.py
burnout87: in case an actual float value is passed

burnout87: in any case, both here as well as for the `Integer` class, it can be removed I think

volodymyrss: I mean the whole block. It's either going to be converted with `float(value)` or it is not float-able. What could be the case which is string literal of float but not converted with `float`?

cdci_data_analysis/analysis/parameters.py
volodymyrss: No idea. It's an old part and not written by me. Some workaround for something.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -448,13 +467,13 @@ def check_int_value(value, units=None,name=None):
...
-            value=np.int(value)
-            if type(value) == int or type(value) == np.int:
-                pass
-            elif type(value) == float or type(value) == np.float:
-                pass
-            else:
-                raise RuntimeError('type of ', name, 'not valid', type(value))
+            try:
+                # TODO a different and bigger precision (eg float64) should be considered?
+                if type(value) == float:
```
volodymyrss, line 460: `isinstance(value, float)` is a more conventional way, and covers cases which we might have.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -448,13 +467,13 @@ def check_int_value(value, units=None,name=None):
...
+                    logger.warning("the value %s will be casted to int" % value)
```
volodymyrss, line 461: maybe warning is not sufficient here. This potentially changes the number. I would raise something leading to invalid usage.
If someone passes "10.0" to max_poinings = it's their fault.

cdci_data_analysis/analysis/parameters.py
burnout87: I was indeed wondering the same, but in any case, it is gone now because non-necessary

cdci_data_analysis/analysis/parameters.py
burnout87: It will no longer use `np.float` or `np.int`, rather `float` or `int`

tests/test_unit.py
```diff
@@ -54,3 +55,64 @@ def test_repeating_parameters(add_duplicate):
...
+                value=value,
+                name="p_float"
+            )
+    else:
+        p_float = Float(
+            value=value,
+            name="p_float"
+        )
+
+        assert p_float.get_value_in_default_format(p_float.value) == p_float.value
```
volodymyrss, line 74: `value` instead of `p_float.value`? The latter possibly already transformed.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -674,85 +673,42 @@ def _set_angle(self, value, units):
...
+                                     units=E_units,
+                                     default_units='keV',
+                                     check_value=check_value,
+                                     name=name,
+                                     allowed_units=_allowed_units)
                                    #wtform_dict=wtform_dict)
 
 
+class SpectralBoundary(Energy):
+    def __init__(self, value=None, E_units='keV', name=None):
```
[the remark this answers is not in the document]
burnout87: After the last changes, there is very little difference between this class and `Energy, but this class is used a lot within the plugin.

Otherwise, I would just simply suggest to remove it

cdci_data_analysis/analysis/instrument.py
```diff
@@ -138,31 +138,32 @@ def set_pars_from_dic(self, par_dic, verbose=False):
...
                     updated_par_dic[par.units_name] = par.default_units
 
+                if par.units_name is not None and par.default_units is None:
+                    raise RequestNotUnderstood("Error when setting the parameter %s: "
+                                               "default unit not specified" % par.name)
+                if par.units_name is None and par.default_units is not None:
```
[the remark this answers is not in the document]
burnout87: would you consider an action also in case both fields were not available?

cdci_data_analysis/analysis/parameters.py
```diff
@@ -761,10 +717,8 @@ def __init__(self,value=None,units='sigma',name=None):
...
                                    allowed_units=_allowed_units)
                                    #wtform_dict=wtform_dict)
-
-
-
-
+    # TODO this method, behaves exaclty the same as the one for the Energy and SpectralBoundary classes
+    # is this ever expected to be more specific?
     @staticmethod
     def check_value(value, units=None,name=None):
```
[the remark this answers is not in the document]
burnout87: yes, exactly the same

I guess it can be removed since it extends `Float`

tests/test_unit.py
burnout87: I dont understand, what would you change?

tests/test_unit.py
```diff
@@ -54,3 +55,64 @@ def test_repeating_parameters(add_duplicate):
...
+            Integer(
+                value=value,
+                name="p_integer"
+            )
+    else:
+        p_integer = Integer(
+            value=value,
+            name="p_integer"
+        )
+        assert p_integer.value == int(value)
```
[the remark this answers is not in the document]
burnout87: So what do you suggest?

As of now, any non-int value is not accepted

tests/test_unit.py
volodymyrss: like you did in another case:

```suggestion
        assert p_float.get_value_in_default_format(p_float.value) == value
```

cdci_data_analysis/analysis/parameters.py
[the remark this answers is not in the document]
volodymyrss: Since it's already used, removing is challenging.

But it can be just 
```suggestion
class SpectralBoundary(Energy):
    pass
```

cdci_data_analysis/analysis/parameters.py
```diff
@@ -167,24 +167,40 @@ def __init__(self,
...
+
         self.check_value = check_value
 
         if allowed_units is None:
             allowed_units = []
         else:
             allowed_units = allowed_units.copy()
 
+        if allowed_types is None:
+            allowed_types = []
```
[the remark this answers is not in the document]
burnout87: ok, then we can follow the same approach with `allowed_units` ?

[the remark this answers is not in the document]
burnout87: and further reduce some code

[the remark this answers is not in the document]
burnout87: also, the code was designed with the approach of performing a unit(s) check only in case the list of allowed units was not None and not empty:

[https://github.com/oda-hub/dispatcher-app/blob/b1773a866332301b0be602bbf16238e3f0aa7e78/cdci_data_analysis/analysis/parameters.py#L237-L239](url)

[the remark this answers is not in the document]
volodymyrss: > also, the code was designed with the approach of performing a unit(s) check only in case the list of allowed units was not None and not empty:
> 
> [https://github.com/oda-hub/dispatcher-app/blob/b1773a866332301b0be602bbf16238e3f0aa7e78/cdci_data_analysis/analysis/parameters.py#L237-L239](url)

I saw that it ignores `[]` as well as `None`, but why make two indicators for disabled check? 


> ok, then we can follow the same approach with `allowed_units` ?

we can!

cdci_data_analysis/analysis/instrument.py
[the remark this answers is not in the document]
volodymyrss: > would you consider an action also in case both fields were not available?

A warning.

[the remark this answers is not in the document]
volodymyrss: By the way, it's not `RequestNotUnderstood` it's `InternalError`: this can only happen if constructor was called incorrectly, and this is done in plugins, in  the code. It's not user error.

tests/test_unit.py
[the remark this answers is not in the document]
burnout87: yes, since the discussion could be very long

tests/test_unit.py
```diff
@@ -54,3 +55,64 @@ def test_repeating_parameters(add_duplicate):
...
+        assert p_integer.get_value_in_default_format(p_integer.value) == int(value)
```
volodymyrss, line 117: casting to `int` here is mostly unnecessary since in this branch it's already int

cdci_data_analysis/analysis/parameters.py
```diff
@@ -476,14 +484,8 @@ def __init__(self, value=None, T_format='isot', name=None, Time_format_name=None
...
 
-    def set_par(self, value, units=None):
-        # sets the value to the format initially specified in the form
-        if units is not None:
-            self.units = units
-        self.value = value
-        # return the relative isot format value
-        # afterwards also the units_name will be set to isot
-        return self._astropy_time.isot
+    def get_value_in_default_format(self, value) -> Union[str, float, None]:
```
volodymyrss, line 487: `value` unused!

cdci_data_analysis/analysis/parameters.py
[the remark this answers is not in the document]
burnout87: > but why make two indicators for disabled check?

nice question, one of those parts of the code that I would definitely modify if it wasn't for the fact that I would be concerned of breaking something else. somewhere else

cdci_data_analysis/analysis/parameters.py
```diff
@@ -297,6 +299,9 @@ def set_par(self, value, units=None):
         if units is not None:
             self.units = units
         self.value = value
+        return self.get_value_in_default_format(value)
+
+    def get_value_in_default_format(self, value):
```
volodymyrss, line 304: oh, why does it need the argument? it already has it as instance attribute

cdci_data_analysis/analysis/parameters.py
```diff
@@ -674,85 +669,31 @@ def _set_angle(self, value, units):
...
-                                   units=E_units,
-                                   check_value=self.check_energy_value,
-                                   name=name,
-                                   allowed_units=_allowed_units)
+                                     units=E_units,
+                                     default_units='keV',
+                                     check_value=check_value,
+                                     name=name,
+                                     allowed_units=_allowed_units)
                                    #wtform_dict=wtform_dict)
```
volodymyrss, line 688: just

```suggestion
        super().__init__(value=value,
```

cdci_data_analysis/analysis/parameters.py
```diff
@@ -232,35 +254,11 @@ def units(self):
     @units.setter
     def units(self,units):
 
-        if self._allowed_units !=[] and self._allowed_units is not None:
-
+        if self._allowed_units is not None and self._allowed_units != []:
```
volodymyrss, line 257:
```suggestion
if self._allowed_units is not None:
```

cdci_data_analysis/analysis/parameters.py
```diff
@@ -218,12 +230,22 @@ def default_units(self):
         return self._default_units
 
     @default_units.setter
-    def default_units(self, units):
+    def default_units(self, par_unit):
+        if self._allowed_types is not None:
+            self.chekc_units(par_unit, self._allowed_units, self.name)
```
volodymyrss, line 235: Oh, maybe since we are deep into this, can we renamed `chekc_units`? With compatibility, by adding this:

```
def chekc_units(self, *args, **kwargs):
    warning('please update to new interface! -- ....')
    return self.check_units(*args, **kwargs)
```

and only using `check_units` in new things.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -382,30 +389,26 @@ def value(self):
 
     @value.setter
     def value(self, v):
-        if v is not None and v!='':
+        if v is not None and v != '':
             self.check_float_value(v, name=self.name, units=self.units)
             self._v = np.float(v)
```
volodymyrss, line 394: exterminate `np.float`!

cdci_data_analysis/analysis/parameters.py
```diff
@@ -382,30 +389,26 @@ def value(self):
...
     @staticmethod
     def check_float_value(value, units=None, name=None):
         #print('check type of ',name,'value', value, 'type',type(value))
-        if value is None or value=='':
+        if value is None or value == '':
             pass
         else:
             try:
-              value=ast.literal_eval(value)
+                value = float(value)
```
volodymyrss, line 409: it does the same as `float(value)`, without assignment. Let's either use new `value` or not assign it.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -431,30 +435,30 @@ def value(self):
...
+        return int(self.value)
 
     @staticmethod
     def check_int_value(value, units=None,name=None):
         #print('check type of ',name,'value', value, 'type',type(value))
         if value is None or value=='':
             pass
         else:
+            if isinstance(value, float):
+                message = '%s is an invalid value for %s since it cannot be used as an Integer' % (value, name)
```
volodymyrss, line 455: hm, just curious, why did you moved away from f-strings? If anything, this '%' form is the oldest. f-string and `.format` are at least equivalent. 
See [note](https://realpython.com/python-string-formatting/) [pic](https://files.realpython.com/media/python-string-formatting-flowchart.4ecf0148fd87.png)

cdci_data_analysis/analysis/parameters.py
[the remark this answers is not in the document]
volodymyrss: you can modify it by preserving interface. E.g. when units are set - `allowed_units` can be checked for `[]`, and if so - assigned `None`. And a warning given.

This is a relatively small part of code, all contained in one 700-line file. But it's old, written with little attention, and affects lot's of things. 
Doing it right will improve on various current and future issues with parameter types.

cdci_data_analysis/analysis/parameters.py
burnout87: I remember asking you the same question on another topic and you replied to me that, given the amount of logs produced, this more old-style formatting style would be less heavy, in comparison with the f-string one.

burnout87: this comment:

[https://github.com/oda-hub/dispatcher-plugin-integral/pull/65#discussion_r723526817](url)

volodymyrss: ha, well I guess I was not clear enough. Maybe the confusion was that I said logging (generic logging) and not `logging` (python module). That case applied to python module logging, used in the commented line. It is also what is referenced in the article linked in that comment.

If you re-read the comment with the `logging` module case, you realize why it matters.
What `logging` does is peculiar. In fact, it does not really use old-style formatting. It just uses very similar string,

For example, compare the next two lines:
```
logger.warning("unknown dev OSA version %s, allowed %s", bla, blabla)
logger.warning("unknown dev OSA version %s, allowed %s" % bla, blabla)
```

First line passes arguments to a logging.warning function, and function does something internally. What it does is somewhat intelligent and optimized.

Second line does the python string formatting, independently of the output levels, and then passes the whole string to `logging.warning`.

burnout87: ok is a bit more clear now, but yeah there are several cases to take into account, and that got me a bit confused especially with the various terminology

whereas this specif case is related to a simple string formatting where for us the `f-{}` is the suggested option

volodymyrss: It's hard not to get confused with python string formatting. There are like 5 different ways, some look similar (for "convenience") but are in fact quite different.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -341,241 +350,203 @@ def reprJSON(self):
...
             except:
-                pass
-            value=np.float(value)
-            if type(value) == int or type(value) == np.int:
-                pass
-            elif type(value) == float or type(value) == np.float:
-                pass
-            else:
-                raise RuntimeError('type of ', name, 'not valid', type(value))
+                raise RuntimeError(f'type {type(value)} not valid for {name}')
```
volodymyrss, line 412: If `value = 'bla'` the message will say `str not valid`, but `str` can be valid if `value = '1.1'`. It's not just about the  type, value also matters.
So could we make this longer, to help debug, adding the actual value in the message along with its type?

cdci_data_analysis/analysis/parameters.py
```diff
@@ -584,78 +555,74 @@ def _split(self,str_list):
...
-            # wtform_dict=wtform_dict)
-
-
-            self._set_angle(value, units=units)
+    def __init__(self, value=None, units=None, name=None):
+
+        super().__init__(value=value,
+                         units=units,
+                         # TODO can we safely make this assumption?
+                         default_units='deg',
```
volodymyrss, line 598: we can make it in principle, but we should not really add `default_units` without `units_name`.

tests/test_unit.py
```diff
@@ -54,3 +65,200 @@ def test_repeating_parameters(add_duplicate):
...
+
+        assert p_spectral_boundary.get_value_in_default_format() == p_spectral_boundary.value
+        assert p_spectral_boundary.get_value_in_default_format() == float(10.)
+        assert type(p_spectral_boundary.value) == float
+
+
+@pytest.mark.parametrize("value",  [25, 25., 25.64547871216879451687311211245117852145229614585985498212321,
+                                    "25", "25.", "25.64547871216879451687311211245117852145229614585985498212321",
+                                    "aaaa"])
+def test_integer_defaults(value):
```
volodymyrss, line 141: I currently see no difference between `float` and `int` which can not be absorbed in one test. See other suggested PR.

volodymyrss (changes requested):
Much nicer! Still few things.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -584,78 +555,74 @@ def _split(self,str_list):
...
-        if units=='names_list':
-            try:
-                #print(type(value))
-                assert (type(value) == list or type(value) == str  or type(str(value))== str)
-            except:
-                raise RuntimeError('par:',name,', value is not product list format : list of strings','it is',type(value),value)
+    def check_list_value(value, units, name='par'):
+        if units == 'names_list':
+            # TODO the condition 'isinstance(str(value), str))' was quite unclear to me, and probably useless since could lead to unexpected behavior
+            if not (isinstance(value, list) or isinstance(value, str)
```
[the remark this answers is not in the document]
burnout87: Oh cool, didn't know this was possible

volodymyrss:
We should require minimum dispatcher version for all used plugins, and test them in 21.10.0003

cdci_data_analysis/analysis/parameters.py
```diff
@@ -166,101 +162,103 @@ class Parameter(object):
...
         self._allowed_values = allowed_values
+        self._allowed_types = allowed_types
+        self._default_type = default_type
         self._default_units = default_units
         self.name = name
         self.units = units
         self.value = value
         self.units_name = units_name
-        #self._wtform_dict=wtform_dict
+        # TODO any ideas on what this represents? seems to be no longer in use
```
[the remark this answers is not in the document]
burnout87: then I will remove all this unused commented code

cdci_data_analysis/analysis/parameters.py
```diff
@@ -289,27 +287,38 @@ def set_value_from_form(self, form, verbose=False):
...
+    def check_units(units, allowed, name):
         if units not in allowed:
-            # TODO this exception is not properly formatted, it could be problematic
-            raise RuntimeError('wrong units for par: %s' % name, ' found: ', units, ' allowed:', allowed)
+            raise RuntimeError('wrong units for par: %s, found: %s, allowed: %s' % (name, units, allowed))
 
     @staticmethod
-    def check_value(val,units,par_name):
+    def check_type(par_type, allowed, name):
+        if type not in allowed:
```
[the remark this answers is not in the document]
burnout87: because the `default_type` was assigned without using the `setter` method, fixing now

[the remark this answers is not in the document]
burnout87: and the list allowed_type should also be defined

cdci_data_analysis/analysis/parameters.py
burnout87: yeah it makes sense, didnt think about it, also for the Integer class, have to consider the same case

cdci_data_analysis/analysis/parameters.py
burnout87: good point, let's not add it for now then

tests/test_unit.py
burnout87: yes I will remove it,

cdci_data_analysis/analysis/parameters.py
volodymyrss: If possible, there should be `units_name` for Energy. Something like `E1_units`. Actually, it might be defined in the plugin, depending on the parameter name.
Maybe then no change is needed here.

volodymyrss (changes requested):
Currently, this fails:

https://github.com/oda-hub/dispatcher-plugin-integral/pull/82

It's not the fault of the dispatcher, it's an unfortunate and very old choice of wildcard import.
But, this PR triggers that problem.
I suggest revert this change (`import six`) to avoid changing plugins... And a warning/todo.

volodymyrss:
still fails: https://github.com/oda-hub/dispatcher-plugin-integral/actions/workflows/python-app.yml ..

burnout87:
> still fails: https://github.com/oda-hub/dispatcher-plugin-integral/actions/workflows/python-app.yml ..

I am debugging it now indeed, some mroe changes here might be necessary

burnout87:
Ok...it seems now that all the plugins are passing the various checks

cdci_data_analysis/analysis/parameters.py
```diff
@@ -166,101 +163,100 @@ class Parameter(object):
...
 
-        if not ( name is None or type(name) in [ str ] ):
+        if allowed_types is not None:
+            allowed_types = allowed_types.copy()
+
+        if not (name is None or type(name) in [str]):
             raise RuntimeError(f"can not initialize parameter with name {name} and type {type(name)}")
 
         self._allowed_units = allowed_units
         self._allowed_values = allowed_values
```
volodymyrss, line 199:
```suggestion
self.default_units = default_units
```

volodymyrss:
@burnout87  could you please have a brief look why the plugin test fails? It seems to be places like this one https://github.com/oda-hub/dispatcher-plugin-integral/blob/0a2f0726b6f90f800679a7970f6a5777a1a942e0/tests/test_authorization.py#L277

I suspect it's unrelated to the current change: it also fails in previous version. 
But it fails if run locally.

volodymyrss (approved):

burnout87:
> @burnout87 could you please have a brief look why the plugin test fails? It seems to be places like this one https://github.com/oda-hub/dispatcher-plugin-integral/blob/0a2f0726b6f90f800679a7970f6a5777a1a942e0/tests/test_authorization.py#L277
> 
> I suspect it's unrelated to the current change: it also fails in previous version. But it fails if run locally.

So it's failing because it's missing this change [https://github.com/oda-hub/dispatcher-plugin-integral/pull/82/commits/5e37e1d5d4b4e25eaf668066faa34a6daf6ce1f8](https://github.com/oda-hub/dispatcher-plugin-integral/pull/82/commits/5e37e1d5d4b4e25eaf668066faa34a6daf6ce1f8)

volodymyrss:
> > > > @burnout87 could you please have a brief look why the plugin test fails? It seems to be places like this one https://github.com/oda-hub/dispatcher-plugin-integral/blob/0a2f0726b6f90f800679a7970f6a5777a1a942e0/tests/test_authorization.py#L277
> > > > I suspect it's unrelated to the current change: it also fails in previous version. But it fails if run locally.
> > > 
> > > 
> > > So it's failing because it's missing this change [oda-hub/dispatcher-plugin-integral@5e37e1d](https://github.com/oda-hub/dispatcher-plugin-integral/commit/5e37e1d5d4b4e25eaf668066faa34a6daf6ce1f8)
> > 
> > 
> > oh good. When was it merged? (was it?)
> 
> The error was to merge it inside [oda-hub/dispatcher-plugin-integral#82](https://github.com/oda-hub/dispatcher-plugin-integral/pull/82) and not isnide the master. If an error actually is

thanks good! means this freshly added plugin testing helps.

volodymyrss:
so you re-create the PR on master? the branch is restored.

burnout87:
> so you re-create the PR on master? the branch is restored.

yes, initally they will fail I guess on the master, but then once this is merged it will be ok, or did I confuse something?

volodymyrss:
> > so you re-create the PR on master? the branch is restored.
> 
> yes, initally they will fail I guess on the master, but then once this is merged it will be ok, or did I confuse something?

it does not even fail!

https://github.com/oda-hub/dispatcher-plugin-integral/pull/85

burnout87:
> 
> it does not even fail!
> 
> [oda-hub/dispatcher-plugin-integral#85](https://github.com/oda-hub/dispatcher-plugin-integral/pull/85)
> 
> once that is merged, we re-test the plugin integration here, and that's it

Yes, seen it, I mixed and confused some changes we did and thought it would not pass some of the tests.
 but I was wrong

volodymyrss:
52 conversations resolved, 12 successful checks. Can it really still have bugs? (yes, it can..)

PR #283, opened 2021-11-21
Empty sentry bugfix

dsavchenko:
dispatcher was not working with empty sentry url in config

### DIFF
--- a/cdci_data_analysis/pytest_fixtures.py
+++ b/cdci_data_analysis/pytest_fixtures.py
@@ -362,6 +362,36 @@ dispatcher:
 
     yield fn
 
+@pytest.fixture
+def dispatcher_test_conf_empty_sentry_fn(tmpdir):
+    fn = os.path.join(tmpdir, "test-dispatcher-conf.yaml")
+    with open(fn, "w") as f:
+        f.write("""
+dispatcher:
+    dummy_cache: dummy-cache
+    products_url: PRODUCTS_URL
+    dispatcher_callback_url_base: http://localhost:8001
+    sentry_url: 
+    logstash_host: 
+    logstash_port: 
+    secret_key: 'secretkey_test'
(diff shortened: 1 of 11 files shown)


cdci_data_analysis/pytest_fixtures.py
```diff
@@ -362,6 +362,36 @@ def dispatcher_test_conf_fn(tmpdir):
...
+        bind_port: 8001
+    email_options:
+        smtp_server: 'localhost'
+        sender_email_address: 'team@odahub.io'
+        cc_receivers_email_addresses: ['team@odahub.io']
+        bcc_receivers_email_addresses: ['teamBcc@odahub.io']
+        smtp_port: 61025
+        smtp_server_password: ''
+        email_sending_timeout: True
+        email_sending_timeout_default_threshold: 1800
```
burnout87, line 389: This an be avoided by simply removing removing the `sentry_url` value from the previous one, or maybe with a parameter that controls the value of `sentry_url`. No need to duplicate.

burnout87 (changes requested):
Perhaps we should not send any sentry in case the url is not specified.

Is it breaking specifically somewhere else?

volodymyrss:
eh, why so many changes, not related to sentry?

volodymyrss (commented):
how's that this one has diverged so much? could you re-create it maybe? (could be related to the force pushes)

volodymyrss:
@dsavchenko I emulated that recreating the PR shows correct changes from just a few commits.

dsavchenko:
Yes, let me recreate the PR.

volodymyrss:
> Yes, let me recreate the PR.

dont forget the [comment](https://github.com/oda-hub/dispatcher-app/pull/283#pullrequestreview-812150475) though!

PR #284, opened 2021-11-22
pathlib is builtin

volodymyrss:
turns out, current master is breaking because now pathlib only works as builtin

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,6 @@ simplejson
 flask
 astropy>=2.0.3
 pylogstash_context>=0.1.19
-pathlib
 gunicorn
 decorator
 python-logstash
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,6 @@ install_req = [
     "simplejson",
     "flask",
     "astropy>=2.0.3",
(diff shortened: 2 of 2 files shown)


burnout87 (changes requested):
No need for this, since it's already implemented in #280

volodymyrss:
> No need for this, since it's already implemented in #280

Of course, #280 deals with another issue. It circumstantially fixes this problem too. But it's not guaranteed #280 will be merged before other (simpler) changes. For example, a much simple #283 .

Merging this first has little impact. Just need to rebase other PRs, it will be automatic - no common files are affected.

So I propose to merge this first.

burnout87:
> > No need for this, since it's already implemented in #280
> 
> Of course, #280 deals with another issue. It circumstantially fixes this problem too. But it's not guaranteed #280 will be merged before other (simpler) changes. For example, a much simple #283 .
> 
> Merging this first has little impact. Just need to rebase other PRs, it will be automatic - no common files are affected.
> 
> So I propose to merge this first.

ok then

burnout87 (approved):

PR #289, opened 2021-11-23
more homogenous no-unit testing

volodymyrss:
Maybe a bit shorter. Mostly that more homogenous treatment between `int` and `float` and more clear table of what maps to what.

For more complex cases like with units this approach might not be better, not sure.

### DIFF
--- a/cdci_data_analysis/analysis/parameters.py
+++ b/cdci_data_analysis/analysis/parameters.py
@@ -25,7 +25,6 @@ from builtins import (bytes, str, open, super, range,
 
 __author__ = "Andrea Tramacere"
 
-import six
 import decorator
 import logging
 
@@ -197,14 +196,12 @@ class Parameter(object):
         self._allowed_units = allowed_units
         self._allowed_values = allowed_values
         self._allowed_types = allowed_types
-        self._default_type = default_type
         self._default_units = default_units
         self.name = name
+        self.default_type = default_type
         self.units = units
         self.value = value
         self.units_name = units_name
-        # TODO any ideas on what this represents? seems to be no longer in use
-        # self._wtform_dict=wtform_dict
 
     @property
     def value(self):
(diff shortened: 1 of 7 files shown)


cdci_data_analysis/analysis/parameters.py
```diff
@@ -408,7 +408,7 @@ def check_float_value(value, units=None, name=None):
             try:
                 value = float(value)
             except:
-                raise RuntimeError('type %s not valid for %s' % (type(value), name))
+                raise RuntimeError(f'unable to interpret value {value} (of type {type(value)}) as float')
```
burnout87, line 411: funny enough, I had applied this change also in the other PR #280 , before checking this

burnout87 (commented):
ok it helps in having two tests in one, but perhaps I would keep them separate to better inspect specific behaviors of the two classes (`Float`, `Integer`), also in case we apply changes in the future

cdci_data_analysis/analysis/parameters.py
volodymyrss: this is the merge on `parameters-default-format`, hence any visible change should not be yet in `parameters-default-format` - in #280 . If it is, rebase before merge will hide it

volodymyrss:
> ok it helps in having two tests in one, but perhaps I would keep them separate to better inspect specific behaviors of the two classes (`Float`, `Integer`), also in case we apply changes in the future

much of the behavior between `Float` and `Integer` is almost identical. In fact, whatever remains different, does not appear to me justified. This is why, to simplify dicussion, I suggest this.

However, if there is any particular difference between `Float` and `Integer` which is not covered by the test I suggest, please let me know, I could have missed it!

cdci_data_analysis/analysis/parameters.py
volodymyrss: I indeed appeared as a conflict and I resolved it - by picking a slightly longer and more detailed message. Do you have better suggestions for the text?

burnout87 (approved):
Just one suggestion for the text in error message, which should also applied to the Integer class

PR #290, opened 2021-11-25
url expiration time in email template

burnout87:


volodymyrss:
That's just it, thanks! 
I am only wondering, there is not test that breaks because of this change?..

burnout87:
> That's just it, thanks! I am only wondering, there is not test that breaks because of this change?..

I was just checking it, apparently not, we have only tests with permanent url

burnout87:
> That's just it, thanks! I am only wondering, there is not test that breaks because of this change?..

I will extend one test, so it'll check this too

burnout87:
> > That's just it, thanks! I am only wondering, there is not test that breaks because of this change?..
> 
> I will extend one test, so it'll check this too

Now it's there

volodymyrss (approved):
thanks!

Issue #273, opened 2021-11-09
refactor error propagation


volodymyrss:
An attempt to improve error reporting was done [here](https://github.com/oda-hub/dispatcher-app/pull/272) but it required deeper refactoring, which should  be done soon.

PR #291, opened 2021-11-25
raise InternalError during get_query_products errors

burnout87:


burnout87:
The way to deal with this is very similar to #264 , for which a major refactoring will be needed. In any case, now the same type of message will be visualized (and to sentry) as in the case of connection issues. This will include information also about the exception raised and the file involved

cdci_data_analysis/analysis/queries.py
```diff
@@ -523,25 +523,28 @@ def get_query_products(self,instrument,job,run_asynch,query_type='Real',logger=N
...
-            e_message = getattr(e, 'message', '')
-            debug_message = repr(e) + ' : ' + getattr(e, 'debug_message', '')
-
-            query_out.set_failed('get_dataserver_products found job failed',
-                                 logger=logger,
-                                 sentry_client=sentry_client,
-                                 excep=e,
-                                 e_message=e_message,
-                                 debug_message=debug_message)
+            raise InternalError(None)
```
volodymyrss, line 538: probably good to add some info from the original exception `e`?

cdci_data_analysis/analysis/queries.py
```diff
@@ -523,25 +523,28 @@ def get_query_products(self,instrument,job,run_asynch,query_type='Real',logger=N
...
+
+            # TODO: could we avoid these? they make error tracking hard
             # TODO we could use the very same approach used when test_communication fails
             logger.exception("failed to get query products")
 
             #status=1
             job.set_failed()
-
+            # TODO is this really helpful ?
             if os.environ.get('DISPATCHER_DEBUG', 'yes') == 'yes':
```
volodymyrss, line 535: It helps debugging for sure. It's set in the tests I think.

volodymyrss:
This should be fine, though I am only almost (not completely) sure it only happens in real issues.
But - ok.

Still, a small suggestion.

volodymyrss (changes requested):
This should be fine, though I am only almost (not completely) sure it only happens in real issues.
But - ok.

Still, a small suggestion.

cdci_data_analysis/analysis/queries.py
```diff
@@ -523,25 +523,19 @@ def get_query_products(self,instrument,job,run_asynch,query_type='Real',logger=N
...
-            e_message = getattr(e, 'message', '')
-            debug_message = repr(e) + ' : ' + getattr(e, 'debug_message', '')
-
-            query_out.set_failed('get_dataserver_products found job failed',
-                                 logger=logger,
-                                 sentry_client=sentry_client,
-                                 excep=e,
-                                 e_message=e_message,
-                                 debug_message=debug_message)
+            raise InternalError(e_message)
```
volodymyrss, line 538: yeah, but class also matters! What if it catches `ValueError` ? Will be quite hard to debug!

volodymyrss (dismissed):
ok, though it might benefit from a simple test

burnout87:
> ok, though it might benefit from a simple test

just included a simple test using a product type that intentionally make the request fails

cdci_data_analysis/analysis/queries.py
```diff
@@ -531,8 +531,9 @@ def get_query_products(self,instrument,job,run_asynch,query_type='Real',logger=N
 
             #status=1
             job.set_failed()
-            if os.environ.get('DISPATCHER_DEBUG', 'yes') == 'yes':
-                raise
+            # TODO are we sure this is still used? Because I instead found DISPATCHER_DEBUG_MODE env
```
volodymyrss, line 534: ok, then can we use DISPATCHER_DEBUG_MODE here?

volodymyrss (changes requested):

cdci_data_analysis/analysis/queries.py
burnout87: let's see if the tests, also for the plufgins, pass without this, otherwise it's making the debugging and proper execution of the test impossible

volodymyrss (approved):

cdci_data_analysis/analysis/queries.py
burnout87: indeed they're not passing for the `all-sky` plugin

burnout87:
After some investigation I figured that the `DISPATCHER_DEBUG_MODE` env var is used in the following cases:

- [https://github.com/oda-hub/dispatcher-app/blob/d05af3bc6dcee109f7cfd20d49632c17f867ecbc/cdci_data_analysis/configurer.py#L105-L109](https://github.com/oda-hub/dispatcher-app/blob/d05af3bc6dcee109f7cfd20d49632c17f867ecbc/cdci_data_analysis/configurer.py#L105-L109) when loading the dispatcher configuration and raise a warning and an error when a valus is set to `None` in case we're not debugging the dispatcher
- [https://github.com/oda-hub/dispatcher-app/blob/d05af3bc6dcee109f7cfd20d49632c17f867ecbc/cdci_data_analysis/plugins/importer.py#L58-L61](https://github.com/oda-hub/dispatcher-app/blob/d05af3bc6dcee109f7cfd20d49632c17f867ecbc/cdci_data_analysis/plugins/importer.py#L58-L61) to load the various testing instruments (empty etc etc)
- the discussed case
-

burnout87:
Now, we have two options:

- close this PR, and also the all-sky one, and postpone this refactoring when we'll work on #273 , that will include another big chunk of changes, that probably will also affect this specific case
- comment out those two discussed lines of code and go merging the other PR for all-sky

I tend to say that is better to go for the first approach, in order to minimize the changes, and also because we will probably have to deal with the same issue when we'll work on #273 . 

What do you think?

volodymyrss:
> Now, we have two options:
> 
>     * close this PR, and also the all-sky one, and postpone this refactoring when we'll work on [refactor error propagation #273](https://github.com/oda-hub/dispatcher-app/issues/273) , that will include another big chunk of changes, that probably will also affect this specific case
> 
>     * comment out those two discussed lines of code and go merging the other PR for all-sky
> 
> 
> I tend to say that is better to go for the first approach, in order to minimize the changes, and also because we will probably have to deal with the same issue when we'll work on #273 .
> 
> What do you think?

Ok, 1st is good. But we need to be sure we store the knowledge acquired in this one here in comments/todos - to avoid remembering it again after.

burnout87:
> 
> Ok, 1st is good. But we need to be sure we store the knowledge acquired in this one here in comments/todos - to avoid remembering it again after.

now the change has been reverted with the addition of some TODOs, also the test that was created is now temporarily marked to fail, so it will pass here

volodymyrss (approved):

volodymyrss:
by the way, notice how I enable auto-merge for this PR: you can also do this. Right now, it merges once all conditions are satisfied (reviews, conversations, tests)

burnout87:
> by the way, notice how I enable auto-merge for this PR: you can also do this. Right now, it merges once all conditions are satisfied (reviews, conversations, tests)

it's only for this PR right? you can enable per-PR I guess?

volodymyrss:
> > by the way, notice how I enable auto-merge for this PR: you can also do this. Right now, it merges once all conditions are satisfied (reviews, conversations, tests)
> 
> it's only for this PR right? you can enable per-PR I guess?

yes, auto-merge disabled by default: 

This one now has it enabled:

image

volodymyrss:
I hesitated adding codecov patch to requirements, since in cases like this it makes little sense. The only sense it makes here is to slightly improve the code, in little relevance to the change. But feel free to merge without.

burnout87:
> I hesitated adding codecov patch to requirements, since in cases like this it makes little sense. The only sense it makes here is to slightly improve the code, in little relevance to the change. But feel free to merge without.

I noticed it would have prevented the auto-merge to take place, and for cases like this is not that relevant.

PR #294, opened 2021-11-30
Limited and restricted plugin test with current app

volodymyrss:
test plugins if `test-plugins` PR label is set. This is one of the possible ways. Let's not set the label unless and until we mean it.


.github/workflows/python-plugins.yml
```diff
@@ -0,0 +1,70 @@
...
+
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip flake8 pytest mypy pylint pytest-cov    
+        
+        if [ -f plugin/requirements.txt ]; then 
+          < plugin/requirements.txt grep -v oda-hub/dispatcher-app | grep -v cdci-data-analysis | grep -v oda_api > plugin-req.txt
+          pip install -r plugin-req.txt ./plugin[test] .[test]; 
+        fi
```
burnout87, line 62: does this mean that the `plugin/requirements.txt` file will have to be included? Or is it created from one of the previous step? I am not familiar yet with github workflows

burnout87 (approved):
just one question

.github/workflows/python-plugins.yml
volodymyrss: yes, it is created in the other step. Plugin is cloned into `plugin` directory. This works.

Some tests fail for `integral` plugin. Strange. I look.

volodymyrss: That's unrelated. But strange.

PR #296, opened 2021-12-01
catch wrong input format

burnout87:
for both catalog and scw_list

### DIFF
--- a/cdci_data_analysis/analysis/instrument.py
+++ b/cdci_data_analysis/analysis/instrument.py
@@ -594,6 +594,7 @@ class Instrument:
                     raise RequestNotUnderstood(e_message)
                 self.set_par('user_catalog', user_catalog)
 
+        # TODO is this case still in use?
         if 'user_catalog_dictionary' in par_dic.keys() and par_dic['user_catalog_dictionary'] is not None:
             if type(par_dic['user_catalog_dictionary']) == dict:
                 catalog_to_build = par_dic['user_catalog_dictionary']
--- a/cdci_data_analysis/analysis/instrument.py
+++ b/cdci_data_analysis/analysis/instrument.py
@@ -601,19 +601,6 @@ class Instrument:
                     raise RequestNotUnderstood(e_message)
                 self.set_par('user_catalog', user_catalog)
 
-        # TODO this is not used from the frontend, perhaps it was in the past
-        if 'user_catalog_dictionary' in par_dic.keys() and par_dic['user_catalog_dictionary'] is not None:
-            if type(par_dic['user_catalog_dictionary']) == dict:
-                catalog_to_build = par_dic['user_catalog_dictionary']
-            else:
-                catalog_to_build = json.loads(par_dic['selected_catalog'])
-            if catalog_to_build is not None:
-                try:
-                    catalog_dic = build_catalog(catalog_to_build)
-                except RuntimeError:
-                    raise RequestNotUnderstood("catalog format not valid")
-                self.set_par('user_catalog', catalog_dic)
-
 
 def load_user_catalog(user_catalog_file):
     return BasicCatalog.from_file(user_catalog_file)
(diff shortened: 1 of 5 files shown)


burnout87:
@volodymyrss do we support scw_list in space separated format? Example:

'066500000010.001 066500010010.001 066500020010.001 066500030010.001 066500040010.001'

volodymyrss:
> @volodymyrss do we support scw_list in space separated format? Example:
> 
> '066500000010.001 066500010010.001 066500020010.001 066500030010.001 066500040010.001'

frontend field rejects it, so dispatcher should as well.

volodymyrss (approved):
seems ok. Thanks!

cdci_data_analysis/analysis/instrument.py
```diff
@@ -539,48 +553,67 @@ def set_input_products(self, par_dic, input_file_path,input_prod_list_name):
...
+                                           '<a href=https://docs.astropy.org/en/stable/api/astropy.io.ascii.Ecsv.html>ECSV</a> table.')
             self.set_par('user_catalog', catalog_object)
-            # TODO set selected_catalog ? in this way it will show up in the email url link
-            # and used consistently for the job_id generation
             self.set_par('selected_catalog', json.dumps(catalog_object.get_dictionary()))
-            # TODO verify that the file path is not needed in the frontend (should not)
+            # TODO not needed in the frontend
             par_dic.pop('user_catalog_file', None)
         else:
+            # TODO is this case still in use?
```
burnout87, line 586: Is this still used for catalogs?

cdci_data_analysis/analysis/instrument.py
```diff
@@ -539,48 +553,67 @@ def set_input_products(self, par_dic, input_file_path,input_prod_list_name):
...
+            if type(par_dic['user_catalog_dictionary']) == dict:
+                catalog_to_build = par_dic['user_catalog_dictionary']
+            else:
+                catalog_to_build = json.loads(par_dic['selected_catalog'])
+            if catalog_to_build is not None:
+                try:
+                    catalog_dic = build_catalog(catalog_to_build)
+                except RuntimeError:
+                    raise RequestNotUnderstood("catalog format not valid")
+                self.set_par('user_catalog', catalog_dic)
```
burnout87, line 615: and also this?

burnout87 (commented):
just one question regarding the setting of a catalog

cdci_data_analysis/analysis/instrument.py
volodymyrss: I think that's what passed in some cases when frontend send the table. Well, you can verify if you want.

cdci_data_analysis/analysis/instrument.py
volodymyrss: there are several ways of what frontend passes. It's good to verify, but it is not, in principle, up to the user. But ok.

volodymyrss (changes requested):
let's then verify all the ways frontend passes catalog

cdci_data_analysis/analysis/instrument.py
burnout87: Just checked, and it is indeed passed from the frontend!

burnout87: this is tested now

cdci_data_analysis/analysis/instrument.py
burnout87: to me, it's not entirely clear how the `user_catalog_dictionary` should be used and when.

it's not used from the frontend though, so I suppose it's some legacy code

volodymyrss (dismissed):

cdci_data_analysis/analysis/instrument.py
volodymyrss: possibly

burnout87: Just for the record, I made an advanced search and found out that `user_catalog_dictionary` was used in the past in some unit test, and nowhere else. So, we can get ride of it.

volodymyrss (approved):

PR #299, opened 2021-12-03
"jobs manager" => "job manager"

volodymyrss:
this is more how they say it (will need to create it also)


burnout87 (approved):

burnout87 (commented):
some tests needs to be also modified accordingly

burnout87 (commented):
some tests needs to be also modified accordingly

burnout87 (approved):

PR #312, opened 2021-12-10
considers token expiration time for submitted email

burnout87:


burnout87:
@volodymyrss do you think this criteria to evaluate whether or not an additional submitted email should be sent is make sense?

` time_from_last_submitted_email > min(time_to_token_expiration, email_sending_job_submitted_interval)`

PR #315, opened 2022-01-04
optional config values

volodymyrss:
it crashes without this. Other instances do not need gallery. The config may be skipped.


burnout87 (approved):
Thanks, haven't thought about it.

PR #316, opened 2022-01-05
additional log

burnout87:


volodymyrss (approved):

burnout87:
I just need the review before merging

volodymyrss:
> I just need the review before merging

ok, I have a look now!

volodymyrss (approved):
ok but I did not think too carefully.

could you modify the commit message to express the intent of the config test, it's to check that example config is compatible with both test config versions?

burnout87:
> could you modify the commit message to express the intent of the config test, it's to check that example config is compatible with both test config versions?

Yes, I was rushed a bit and didn't create a very explanatory message. 

I just added an additional test to verify the compatibility of the example config with the both test config versions (with and without the product-gallery related entries).

PR #318, opened 2022-01-05
conditionally ignore some tests

volodymyrss:


.github/workflows/python-app.yml
```diff
@@ -26,32 +30,40 @@ jobs:
...
-    - name: Test with pytest
+    - name: Test fast subset with pytest
       run: |
-        python -m pytest tests --cov=cdci_data_analysis -sv --full-trace --log-cli-level=DEBUG -m 'not test_drupal'
+        python -m pytest tests --cov=cdci_data_analysis -sv --full-trace --log-cli-level=DEBUG -m "not test_drupal and fast" --durations=0
+
+    - name: Test everything with pytest
+      if: ${{ env.TEST_ONLY_FAST != 'true' }}
+      run: |
+        python -m pytest tests --cov=cdci_data_analysis -sv --full-trace --log-cli-level=DEBUG -m "not test_drupal" --durations=0
```
burnout87, line 65: perhaps I would also have the `fast` tests not run since they are run in the previous step

burnout87 (commented):
Cool! Just a couple of questions regarding the conditions with the markers

.github/workflows/python-app.yml
volodymyrss: I left them since there might be some unwanted interaction.

Also, in fast mode, all dispatchers are long-living.

https://github.com/oda-hub/dispatcher-app/pull/318/files#diff-87a0b6a348b0b79bdf41f844da34925be0474bf965cc3a09ff199e93f0628d93R582-R584

So it helps to re-run them in normal mode, in case there is some unwanted state preservation.

.github/workflows/python-app.yml
```diff
@@ -26,32 +30,40 @@ jobs:
...
     - name: MyPy
+      if: ${{ env.TEST_ONLY_FAST != 'true' }}
       run: |
         mypy cdci_data_analysis --ignore-missing-imports || echo 'this is fine too!'
 
-    - name: Test with pytest
+    - name: Test fast subset with pytest
       run: |
-        python -m pytest tests --cov=cdci_data_analysis -sv --full-trace --log-cli-level=DEBUG -m 'not test_drupal'
+        python -m pytest tests --cov=cdci_data_analysis -sv --full-trace --log-cli-level=DEBUG -m "not test_drupal and fast" --durations=0
```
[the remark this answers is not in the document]
volodymyrss: yes, sure.

here are two variations:

https://github.com/oda-hub/dispatcher-app/runs/4715907832?check_suite_focus=true
https://github.com/oda-hub/dispatcher-app/runs/4715986894?check_suite_focus=true

.github/workflows/enforce-labels.yml
```diff
@@ -0,0 +1,15 @@
...
+    types: [labeled, unlabeled, opened, edited, synchronize]
+jobs:
+  enforce-label:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: yogevbd/enforce-label-action@2.1.0
+      with:
+        #REQUIRED_LABELS_ANY: "bug,enhancement,skip-changelog"
+        #REQUIRED_LABELS_ANY_DESCRIPTION: "Select at least one label ['bug','enhancement','skip-changelog']"
+        BANNED_LABELS: "test-only-fast"
```
burnout87, line 14: I dont understand what this is actually doing, preventing the user from assigning that label? Which purpose?

volodymyrss: Preventing merging if label is set. The test will fail and it's required.

It's so that we do not forget to remove the label before merging.

burnout87 (approved):

PR #320, opened 2022-01-05
dedicated function for request to the gallery and additional error log

burnout87:


cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -42,6 +42,22 @@ def analyze_drupal_output(drupal_output, operation_performed=None):
...
+                                params={**params},
+                                headers=headers)
+        elif method == 'post':
+            return requests.post(url,
+                                 data={**params},
+                                 files=files,
+                                 headers=headers
+                                 )
+    except ConnectionError as e:
+        logger.warning("connection to the drupal instance failed %s", e)
```
volodymyrss, line 58: what will the user  see? This way we get a warning and that's it?

User should get a notice that `gallery is inaccessible, we are investigating`
And we should get a sentry.

Actually we do now, so it's not so bad.
Just that user gets 500.

volodymyrss (changes requested):

cdci_data_analysis/analysis/drupal_helper.py
burnout87: It was not complete ye, indeed this is still a draft.

But thanks for the suggestion.

So, you actually got a sentry?

cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -42,6 +42,47 @@ def analyze_drupal_output(drupal_output, operation_performed=None):
...
+    except Exception as e:
+        logger.warning(f"an issue occurred when performing a request to the product gallery, "
+                       f"this is likely to be a connection related problem, we are investigating and "
+                       f"try to solve the it as soon as possible")
+        if sentry_client is not None:
+            sentry_client.capture('raven.events.Message',
+                                  message=f'exception when performing a request to the product gallery: {repr(e)}')
+        else:
+            logger.warning("sentry not used")
+
```
volodymyrss, line 84: funtion will return `None`

cdci_data_analysis/flask_app/app.py
```diff
@@ -431,10 +431,18 @@ def post_product_to_gallery():
...
     par_dic.pop('token')
 
+    sentry_url = getattr(app_config, 'sentry_url', None)
+    sentry_client = None
+    if sentry_url is not None:
+        from raven import Client
+
+        sentry_client = Client(sentry_url)
+
     output_post = drupal_helper.post_content_to_gallery(product_gallery_url=product_gallery_url,
```
volodymyrss, line 441: this might return `None`, if problem

cdci_data_analysis/flask_app/app.py
```diff
@@ -431,10 +431,18 @@ def post_product_to_gallery():
...
                                                         decoded_token=decoded_token,
                                                         gallery_secret_key=gallery_secret_key,
                                                         files=request.files,
+                                                        sentry_client=sentry_client,
                                                         **par_dic)
 
     return output_post
```
volodymyrss, line 448: `None` here is not too good

volodymyrss (changes requested):

cdci_data_analysis/analysis/drupal_helper.py
volodymyrss: I got a sentry. That was ok.

But user got 500. It's fine, but not ideal.

cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -42,6 +42,51 @@ def analyze_drupal_output(drupal_output, operation_performed=None):
...
+        logger.warning(f"an issue occurred when performing a request to the product gallery, "
+                       f"this is likely to be a connection related problem, we are investigating and "
+                       f"try to solve it as soon as possible")
+        if sentry_client is not None:
+            sentry_client.capture('raven.events.Message',
+                                  message=f'exception when performing a request to the product gallery: {repr(e)}')
+        else:
+            logger.warning("sentry not used")
+
+        raise BadRequest('issue when performing a request to the product gallery',
```
volodymyrss, line 85: `BadRequest` should mean user did something wrong, and should modify the requet. Some InternalError would be better here.

volodymyrss (approved):
small comment but ok

volodymyrss (approved):

PR #321, opened 2022-01-05
hide gallery secret key

volodymyrss:


volodymyrss:
I merge promptly since it's a vulnerability.

PR #325, opened 2022-01-10
multiple retries of the request for connection issues

burnout87:


cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -52,39 +55,81 @@ def execute_drupal_request(url,
...
+                    response_json = res.json()
+                    # a 403 has been noticed to be returned in two different cases:
+                    # * for not-valid token
+                    # * not-completed request
+                    error_msg = response_json['message']
+                except json.decoder.JSONDecodeError:
+                    error_msg = res.text
+                raise RequestNotAuthorized(error_msg)
+
+            if res.status_code in 404:
```
volodymyrss: `== 404`? `in [404]`?

cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -52,39 +55,81 @@ def execute_drupal_request(url,
...
+                logger.warning(f"there seems to be some problem in completing a request to the product gallery,"
+                               " this might be due to a an error in the url, "
+                               "or because we are attempting to load a page that no longer exists, "
+                               "please check it and try to issue again the request")
```
volodymyrss, line 97: could you add here in the warning and also in the exception the page which was requested, and `res.text`. Would be very helpful.

cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -52,39 +55,81 @@ def execute_drupal_request(url,
...
+                raise InternalError('issue when performing a request to the product gallery',
+                                    status_code=500,
+                                    payload={'error_message': str(e)})
+
```
volodymyrss, line 101: maybe a check for anything but `200` can be done? does it return `200` when posted successfully or more restfull `201`?

cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -52,39 +55,81 @@ def execute_drupal_request(url,
...
+            return res
+
+        except (ConnectionError,
+                RequestNotAuthorized,
+                requests.exceptions.ConnectionError,
+                requests.exceptions.Timeout) as e:
+            n_tries_left -= 1
+            if n_tries_left > 0:
+                if n_tries_left < round(n_max_tries / 2):
```
volodymyrss, line 110: I have a small suggestion which I found very useful in such cases. There is a better way to do it and easy one. Easy one is better here.

We can make a global (module-scope) counter of these issues. Global state is dangerous, but it's applicable here, if just for logging.
Variables `total_n_successful_post_requests`, `total_n_post_request_retries`.
Increment them accordinly, and in the log message add "average post retries per request since dispatcher start: total_n_post_request_retries/total_n_successful_post_requests".
You could then also add an extra warning `when n_max_tries - n_tries_left  > total_n_post_request_retries/total_n_successful_post_requests`

What people often do is to report such metric e.g. in last hour, as self-health check. But it's maybe unnecessarily complex.

cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -223,9 +268,10 @@ def get_observations_for_time_range(product_gallery_url, gallery_jwt_token, t1=N
         # format the time fields, from the format request, with +/- 1ms
         t1_minor = parser.parse(t1) - datetime.timedelta(seconds=1)
         t2_plus = parser.parse(t2) + datetime.timedelta(seconds=1)
-        t1_minor_formatted = t1_minor.strftime('%Y-%m-%dT%H:%M:%S')
-        t2_plus_formatted = t2_plus.strftime('%Y-%m-%dT%H:%M:%S')
-        # eg /mmoda-pg/observations/range/2018-12-31T23%3A59%3A59--2021-12-01T00%3A00%3A01
+        # using the format accepted by the product gallery
+        t1_minor_formatted = t1_minor.strftime('%Y-%d-%mT%H:%M:%S')
```
volodymyrss, line 271: omg! We should stick to ISO. Are you sure it uses this weird mix of formats? If yes, it should be exterimnated at the level of drupal, by changing the defaults. It's a source of frequent confusion, and this sort of issues commonly cause issues in astronomy!

volodymyrss (changes requested):

cdci_data_analysis/analysis/drupal_helper.py
burnout87: yes was a typo, already changed

cdci_data_analysis/analysis/drupal_helper.py
burnout87: `201` when posted something successfully (article or image), `200` when successfully returns a `GET` call

burnout87: > maybe a check for anything but `200` can be done?

like I do now?

cdci_data_analysis/analysis/drupal_helper.py
burnout87: Thanks, very nice suggestion!

cdci_data_analysis/analysis/drupal_helper.py
volodymyrss: I mean, if it's not possible to avoid it, we will face troubles, and we will have to keep track very cautiously. It's especially dangerous because of how similar (but seemingly specific) the two formats appear.

burnout87: I have searched and it is affected only in the way the timerange is formatted for this REST call, but I perhaps I will figure out how to customize also this.

I am also worried regarding the potential issues we might face, so I will keep track of it.

In any case, over the product gallery, it is possible to fully customize the format of the time we want to display.

volodymyrss: For presentation - sure. But two-way translation will be done, in part implicitly.

also you quote this:

> eg /mmoda-pg/observations/range/2018-31-12T23%3A59%3A59--2021-01-12T00%3A00%3A0

does it mean in the urls the time will be the wrong format sometimes?

burnout87: > 
> > eg /mmoda-pg/observations/range/2018-31-12T23%3A59%3A59--2021-01-12T00%3A00%3A0
> 
> does it mean in the urls the time will be the wrong format sometimes?

This is an example of the resulting request directly copied from the views page, the time format will always be this one. What do you mean with wrong?

volodymyrss: > > > eg /mmoda-pg/observations/range/2018-31-12T23%3A59%3A59--2021-01-12T00%3A00%3A0
> > 
> > 
> > does it mean in the urls the time will be the wrong format sometimes?
> 
> This is an example of the resulting request directly copied from the views page, the time format will always be this one. What do you mean with wrong?

So this sort of URLs will occur, may be copied, and may need to be shared along?

This string: `2018-31-12T23:A5:59` - maybe I could call it not so much wrong, as evil. If it only occurs somewhere in drupal internals - it's ok as long as we do not see if often. If they appear in emails which might be shared -  it's worse.

It's not about matter of taste, we rely on quick look assessment often, and introducing like that an opportunity for mistake.
I have seen multiple times this exact kind of mixed representation causing mistakes, loss of time (of people and observatories) and loss of valuable opportunity. This is no joke. 

I suppose it may be possible to change the default format in drupal. Maybe we can look into that later, if it's not trivial.
If so, let's make an issue.
And discuss it on a meeting.

burnout87: > 
> So this sort of URLs will occur, may be copied, and may need to be shared along?

No, this will only be used internally, no need to share this around.

> 
> I suppose it may be possible to change the default format in drupal. Maybe we can look into that later, if it's not trivial. If so, let's make an issue. And discuss it on a meeting.

This is what I am investigating right now. 

The tricky point here is that the feature to use the timerange as a filter is provided by an external module, and if this does not come with the option to configure the format of the input.

> I suppose it may be possible to change the default format in drupal. Maybe we can look into that later, if it's not trivial.
If so, let's make an issue.

Let's talk about it in a separate issue and decide there.

volodymyrss: https://github.com/oda-hub/dispatcher-app/issues/329

cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -52,39 +58,90 @@ def execute_drupal_request(url,
...
+                raise RequestNotAuthorized(error_msg)
+
+            if res.status_code not in [200, 201]:
+                logger.warning(f"there seems to be some problem in completing a request to the product gallery:\n"
+                               f"the requested url {url} lead to the error {res.text}, "
+                               "this might be due to an error in the url or the page requested no longer exists, "
+                               "please check it and try to issue again the request")
+                raise InternalError('issue when performing a request to the product gallery',
+                                    status_code=500,
+                                    payload={'error_message': str(e)})
```
volodymyrss, line 105: `e` is undefined here

cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -52,39 +58,90 @@ def execute_drupal_request(url,
...
+                    response_json = res.json()
+                    # a 403 has been noticed to be returned in two different cases:
+                    # * for not-valid token
+                    # * not-completed request
+                    error_msg = response_json['message']
+                except json.decoder.JSONDecodeError:
+                    error_msg = res.text
+                raise RequestNotAuthorized(error_msg)
+
+            if res.status_code not in [200, 201]:
```
volodymyrss, line 98: `elif`

volodymyrss (changes requested):
test does not pass. There are at least two clear issues. Maybe I am looking on wrong code?

burnout87:
> test does not pass. There are at least two clear issues. Maybe I am looking on wrong code?

mmmm I am investigating

cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -52,39 +58,90 @@ def execute_drupal_request(url,
...
+                    # a 403 has been noticed to be returned in two different cases:
+                    # * for not-valid token
+                    # * not-completed request
+                    error_msg = response_json['message']
+                except json.decoder.JSONDecodeError:
+                    error_msg = res.text
+                raise RequestNotAuthorized(error_msg)
+
+            elif res.status_code not in [200, 201]:
+                logger.warning(f"there seems to be some problem in completing a request to the product gallery:\n"
```
volodymyrss, line 99: I get this:

```
the requested url http://cdciweb02.isdc.unige.ch/mmoda/gallery/observations/range/2018-31-12T23:59:59--2021-01-12T00:00:01 lead to the error {"message":"No route found for \u0022GET \/observations\/range\/2018-31-12T23:59:59--2021-01-12T00:00:01\u0022"}, this might be due to an error in the url or the page requested no longer exists, please check it and try to issue again the request
```

did it pass for you before the push?

burnout87: I am still investigating, I suspect some configuration within the view for the observation, I am seeing the very same error

burnout87: ok it works for me now, for some reasons I had disabled the view :( I must have been struggling too much on these formats

volodymyrss: works for me too.

volodymyrss (approved):

PR #331, opened 2022-01-12
using different endpoint for observations range

burnout87:


volodymyrss (approved):

burnout87:
BTW once the `test-only-fast` label is removed, again the rest of the complete checks re-run

volodymyrss:
> BTW once the `test-only-fast` label is removed, again the rest of the complete checks re-run

It could be adapted, but that's the simple behavior default. 
It's somehow preferable, since larger danger is if an outdated version of test runs somehow remains at the time of merge. 
In this case it's not too long to run any of the tests, right?

volodymyrss:
> BTW once the `test-only-fast` label is removed, again the rest of the complete checks re-run

yes, it's kind of intentional. So that the tests are always up-to-date.

Issue #334, opened 2022-01-24
 Angle parameter gets converted to degrees when passing to backend


dsavchenko:
As an example, this
https://github.com/oda-hub/dispatcher-plugin-legacysurvey/blob/b1fee836889b530380c1fe1556fe7e2d1e9f395b/dispatcher_plugin_legacysurvey/queries.py#L143
parameter when set to 3. in the frontend becomes 0.05 in param_dict here 
https://github.com/oda-hub/dispatcher-plugin-legacysurvey/blob/b1fee836889b530380c1fe1556fe7e2d1e9f395b/dispatcher_plugin_legacysurvey/queries.py#L153
but I expect it not to convert unless get_value_in_default_format() is explicitly called.

As we can see from test https://github.com/oda-hub/dispatcher-app/blob/15b294af455f754ed5c986d76a636f88d1b4ba28/tests/test_unit.py#L130
the Angle class itself works well. 

I suspect that unintended conversion happens when setting dict here https://github.com/oda-hub/dispatcher-app/blob/15b294af455f754ed5c986d76a636f88d1b4ba28/cdci_data_analysis/analysis/instrument.py#L149 and then the values are somehow    
pushed back to the Parameter objects. At this point, I am a bit lost.
@burnout87 @volodymyrss you were working with the units of parameters recently, can you help?

burnout87:
Which parameter is giving you issues? `pixel_size`?

volodymyrss:
Could you @dsavchenko please demonstrate the effect in a test?

dsavchenko:
> Which parameter is giving you issues? `pixel_size`?
> 
> I don't understand why from a value of 3. (which is in arc_sec I guess) you get a value of 0.05 when a conversion to a default format is applied.

Sorry. This one becomes 0.0008(3) of course

dsavchenko:
> Could you @dsavchenko please demonstrate the effect in a test? 

I should think again, how to properly construct such a test

burnout87:
also, at which step do you notice that the value is not the one you were expecting?

dsavchenko:
> also, at which step do you notice that the value is not the one you were expecting?

I noticed a very strange and cryptic exception in the backend of legacysurvey. After spending a few hours on that I finally realised that the values of some parameters passed to the backend are not the same as being set in the frontend.

What's interesting, is that it was working some time ago.

burnout87:
> 
> What's interesting, is that it was working some time ago.

[...]

PR #335, opened 2022-01-25
Improve support for Parameter units/format/type

dsavchenko:
addressing #334

### DIFF
--- a/cdci_data_analysis/analysis/parameters.py
+++ b/cdci_data_analysis/analysis/parameters.py
@@ -161,11 +161,22 @@ class ParameterTuple(object):
 
 class Parameter(object):
     """
+    # General notes
+
     format:
-    every parameter has a format. Format defines representation of the parameter value. 
-    Default format of  defines unique representation 
+        * Every `Parameter` has a *format*. *Format* defines representation of the parameter value in parameter URL, analysis_parameters dictionary, oda_api, and embedded in fits files.       
+        * Parameter dictionary may include format specifiers for each parameter.  If non-default format is used the parameter, parameter format specifier is required. If not specified, the default is used.
+        * *default format* of defines **unique** representation of the parameter. Any parameter value can be converted to default *format*. By using default parameter representations, it is possible to construct unique and deterministic request parameter dictionaries, URLs, oda_api codes.
+
+    unit:
+    * physical quantities are represented as floats (and should inherit from `Float`) and are scaled with *units*. The same value may be represented as different floats if units are correspondingly different.
+    * units are treated similarly to formats: there are unit specifiers and default units. 
     
-    TODO
+    type:
+    * each parameter is constructed with a value of some types. Types is verified at construction, and converted to default type. Further representations are defined by units and formats
+
+    
+    TODO see through that this is implemented
     """
     def __init__(self,
                  value=None,
--- a/cdci_data_analysis/analysis/parameters.py
+++ b/cdci_data_analysis/analysis/parameters.py
@@ -160,18 +160,18 @@ class ParameterTuple(object):
 
 
 class Parameter(object):
+    """
+    format:
+    every parameter has a format. Format defines representation of the parameter value. 
+    Default format of  defines unique representation 
+    
+    TODO
+    """
     def __init__(self,
                  value=None,
                  units=None,
                  name: Union[str, None] = None,
                  # TODO should we remove units/type/format knowledge from the Parameter class?
-"""
-format:
-   every parameter has a format. Format defines representation of the parameter value. 
-   Default format of  defines unique representation 
-  
-TODO
-"""
                  allowed_units=None,
                  default_units=None,
                  units_name=None,
--- a/cdci_data_analysis/analysis/parameters.py
+++ b/cdci_data_analysis/analysis/parameters.py
@@ -164,19 +164,18 @@ class Parameter(object):
     # General notes
 
     format:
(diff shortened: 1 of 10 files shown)


cdci_data_analysis/analysis/parameters.py
```diff
@@ -164,13 +164,29 @@ def __init__(self,
                  value=None,
                  units=None,
                  name: Union[str, None] = None,
+                 # TODO should we remove units/type/format knowledge from the Parameter class?
+"""
+format:
+   every parameter has a format. Format defines representation of the parameter value. 
+   Default format of  defines unique representation 
+  
+TODO
```
volodymyrss, line 173: I did not finish the string, here it continues:

```suggestion
"""
# General notes

format:
    * Every `Parameter` has a *format*. *Format* defines representation of the parameter value in parameter URL, analysis_parameters dictionary, oda_api, and embedded in fits files.       
    * Parameter dictionary may include format specifiers for each parameter.  If non-default format is used the parameter, parameter format specifier is required. If not specified, the default is used.
    * *default format* of defines **unique** representation of the parameter. Any parameter value can be converted to default *format*. By using default parameter representations, it is possible to construct unique and deterministic request parameter dictionaries, URLs, oda_api codes.

units:
   * physical quantities are represented as floats and are scaled with *units*. The same value may be represented as different floats if units are correspondingly different.
   * units are treated similarly to formats: there are unit specifiers and default units. 

  
TODO see through that this is implemented
"""
```

cdci_data_analysis/analysis/parameters.py
```diff
@@ -164,13 +164,29 @@ def __init__(self,
...
+"""
```
burnout87, line 174: I think this needs correct indentation

volodymyrss: yeah, the suggestion did not embed correctly. Please feed free to adapt @dsavchenko ? Also I expended it with another suggestion.

dsavchenko: Done

cdci_data_analysis/analysis/parameters.py
volodymyrss:
```suggestion
format:
    * Every `Parameter` has a *format*. *Format* defines representation of the parameter value in parameter URL, analysis_parameters dictionary, oda_api, and embedded in fits files.       
    * Parameter dictionary may include format specifiers for each parameter.  If non-default format is used the parameter, parameter format specifier is required. If not specified, the default is used.
    * *default format* of defines **unique** representation of the parameter. Any parameter value can be converted to default *format*. By using default parameter representations, it is possible to construct unique and deterministic request parameter dictionaries, URLs, oda_api codes.

unit:
   * physical quantities are represented as floats (and should inherit from `Float`) and are scaled with *units*. The same value may be represented as different floats if units are correspondingly different.
   * units are treated similarly to formats: there are unit specifiers and default units. 
   
type:
   * each parameter is constructed with a value of some types. Types is verified at construction, and converted to default type. Further representations are defined by units and formats

  
TODO see through that this is implemented https://github.com/oda-hub/dispatcher-app/issues/339
```

cdci_data_analysis/analysis/parameters.py
volodymyrss: See I expanded it.

dsavchenko: Hurried again. Sorry

cdci_data_analysis/analysis/parameters.py
burnout87: Does it make sense to add the following point to the unit part:

```
* for some parameters units might not be specified (eg Time) since not necessary, the same applies the other way around
```

Also, regarding the type: at the moment this might not be specified for certain parameters (eg Time, Name), shall we mention that?

volodymyrss: as we use astropy Time, it would not inherit from `Float`. And only `Float` can have units. 
In principle, for Time, this is not the only option. We could have used time since some epoch. This would be a float (measured in days, or seconds etc since epoch).
But, if it's astropy Time - it does not have inherit  from Float, and has no type. In principle time since epoch is more like TimeDelta than Time - timestamp.
I add this small example.

Time and Name would have type,  why not? astropy.time.Time and string correspondingly.

burnout87:
any ideas on why the plugins are all failing?

cdci_data_analysis/analysis/parameters.py
burnout87: > 
> Time and Name would have type, why not? astropy.time.Time and string correspondingly.

Ok then this will have to be considered when dealing with #339

volodymyrss:
> any ideas on why the plugins are all failing?

Probably some non-frozen dependency changed. Does it work locally?

burnout87:
> > any ideas on why the plugins are all failing?
> 
> Probably some non-frozen dependency changed. Does it work locally?

Yes, it seems so

burnout87 (approved):

burnout87:
> 
> [oda-hub/oda_api#120](https://github.com/oda-hub/oda_api/pull/120) adapts the test

as #341 shows, just changing that test, `tests/test_cli.py::test_get` passes

burnout87:
ok this can be merged, anything else to add @dsavchenko  ?

volodymyrss (approved):

dsavchenko:
Let's merge.

PR #336, opened 2022-01-26
support for different units

burnout87:
probably not  the best solution, but it behaves in the way we expect and the way we wanted since we applied this approach

### DIFF
--- a/.github/workflows/python-oda-api.yml
+++ b/.github/workflows/python-oda-api.yml
@@ -56,5 +56,5 @@ jobs:
     
     - name: Test oda_api with pytest
       run: |
-        cd oda_api && DISPATCHER_MOCK_KB=yes python -m pytest tests -sv --full-trace --log-cli-level=DEBUG 
+        cd oda_api && DISPATCHER_MOCK_KB=yes python -m pytest tests -m 'not slow and not test_drupal' -sv --full-trace --log-cli-level=DEBUG 
     
--- a/cdci_data_analysis/analysis/instrument.py
+++ b/cdci_data_analysis/analysis/instrument.py
@@ -270,7 +270,7 @@ class Instrument:
         #  this was removed by 2f5b5dfb7e but turns out it is used by some plugins, see test_server_plugin_integral_all_sky
         self._current_par_dic=par_dic
 
-        # set pars values from the input parameters
+        # # set pars values from the input parameters
         query_out = self.set_pars_from_form(par_dic, verbose=verbose, sentry_client=sentry_client)
         if verbose:
             self.show_parameters_list()
--- a/cdci_data_analysis/analysis/instrument.py
+++ b/cdci_data_analysis/analysis/instrument.py
@@ -271,7 +271,9 @@ class Instrument:
         self._current_par_dic=par_dic
 
         # # set pars values from the input parameters
-        query_out = self.set_pars_from_form(par_dic, verbose=verbose, sentry_client=sentry_client)
+        # query_out = self.set_pars_from_form(par_dic, verbose=verbose, sentry_client=sentry_client)
+        query_out = QueryOutput()
+        query_out.set_done()
         if verbose:
             self.show_parameters_list()
 
--- a/cdci_data_analysis/analysis/instrument.py
+++ b/cdci_data_analysis/analysis/instrument.py
@@ -154,6 +154,12 @@ class Instrument:
                     else:
                         raise InternalError("Error when setting the parameter %s: "
                                             "default unit not specified" % par.name)
+                if par.par_format_name is not None:
+                    if par.par_default_format is not None:
+                        updated_par_dic[par.par_format_name] = par.par_default_format
+                    else:
+                        raise InternalError("Error when setting the parameter %s: "
+                                            "default format not specified" % par.name)
                 else:
                     self.logger.warning("units_name for the parameter %s not specified", par.name)
 
--- a/cdci_data_analysis/analysis/parameters.py
+++ b/cdci_data_analysis/analysis/parameters.py
@@ -662,8 +662,8 @@ class Energy(Float):
 
     # TODO re-introduced for retro-compatibility
     @staticmethod
-    def check_energy_value(value, units=None, name=None):
-        Float.check_float_value(value, units, name)
+    def check_energy_value(value, units=None, name=None, par_format=None):
+        Float.check_float_value(value, units=units, name=name, par_format=par_format)
 
 
 class SpectralBoundary(Energy):
@@ -691,5 +691,5 @@ class UserCatalog(Parameter):
                          allowed_units=_allowed_units)
 
     @staticmethod
-    def check_name_value(value, units=None, name=None):
+    def check_name_value(value, units=None, name=None, par_format=None):
         pass
--- a/cdci_data_analysis/analysis/parameters.py
+++ b/cdci_data_analysis/analysis/parameters.py
@@ -677,7 +677,7 @@ class DetectionThreshold(Float):
         super().__init__(value=value,
                          units=units,
                          # TODO to check if it's correct
-                         check_value=self.check_value,
+                         check_value=self.check_float_value,
                          name=name,
                          allowed_units=_allowed_units)
 
--- a/cdci_data_analysis/analysis/parameters.py
+++ b/cdci_data_analysis/analysis/parameters.py
@@ -309,6 +309,10 @@ class Parameter(object):
     def get_value_in_default_units(self):
         return self.value
 
+    # to ensure backward compatibility
+    def get_value_in_default_format(self):
+        return self.get_value_in_default_units()
+
     def get_value_in_units(self, units):
         return self.value
 
--- a/cdci_data_analysis/analysis/parameters.py
+++ b/cdci_data_analysis/analysis/parameters.py
@@ -394,7 +394,7 @@ class Float(Parameter):
                          name=name,
                          default_type=float,
                          # TODO added for consistency with Integer
-                         allowed_types=[int],
+                         allowed_types=[float],
                          allowed_units=allowed_units)
 
     @property
--- a/cdci_data_analysis/analysis/parameters.py
+++ b/cdci_data_analysis/analysis/parameters.py
@@ -564,12 +564,12 @@ class InputProdList(Parameter):
 
 
 class Angle(Parameter):
-    def __init__(self, value=None, units=None, name=None):
+    def __init__(self, value=None, units=None, default_units='deg', name=None):
 
         super().__init__(value=value,
                          units=units,
                          # TODO can we safely make this assumption?
-                         default_units='deg',
+                         default_units=default_units,
                          name=name,
                          allowed_units=None)
 
@@ -599,7 +599,7 @@ class Angle(Parameter):
 
 
 class Energy(Float):
-    def __init__(self, value=None, E_units='keV', name=None, check_value=None):
+    def __init__(self, value=None, E_units='keV', default_units='keV', name=None, check_value=None):
         if check_value is None:
             check_value = self.check_energy_value
 
@@ -607,7 +607,7 @@ class Energy(Float):
 
         super().__init__(value=value,
                          units=E_units,
-                         default_units='keV',
+                         default_units=default_units,
                          check_value=check_value,
                          name=name,
                          allowed_units=_allowed_units)
--- a/cdci_data_analysis/analysis/parameters.py
+++ b/cdci_data_analysis/analysis/parameters.py
@@ -302,11 +302,21 @@ class Parameter(object):
     def set_par(self, value, units=None):
         if units is not None:
             self.units = units
-        self.value = value
-        return self.get_value_in_default_format()
 
-    def get_value_in_default_format(self):
+        if self.units is not None and self.default_units is not None and self.units != self.default_units:
+            # convert value in the default units and assigns it to the value
+            value = self.get_value_in_units(value, self.units, self.default_units)
+            self.units = self.default_units
+
+        self.value = value
+        return self.get_value_in_default_units()
+
+    def get_value_in_default_units(self):
         return self.value
+
+    @staticmethod
+    def get_value_in_units(value, units_in, units_out):
+        return value
 
     def get_form(self, wtform_cls, key, validators, defaults):
         return wtform_cls('key', validators=validators, default=defaults)
@@ -373,7 +383,7 @@ class Float(Parameter):
         else:
             self._v = None
 
-    def get_value_in_default_format(self):
+    def get_value_in_default_units(self):
         self.check_value(self.value, name=self.name, units=self.units)
         return float(self.value) if self.value is not None else None
 
@@ -417,7 +427,7 @@ class Integer(Parameter):
         else:
             self._v = None
 
-    def get_value_in_default_format(self):
+    def get_value_in_default_units(self):
         self.check_value(self.value, name=self.name, units=self.units)
         return int(self.value) if self.value is not None else None
 
@@ -447,7 +457,7 @@ class Time(Parameter):
                          default_units='isot',
                          name=name)
 
-    def get_value_in_default_format(self) -> Union[str, float, None]:
+    def get_value_in_default_units(self) -> Union[str, float, None]:
         return getattr(self._astropy_time, self.default_units)
 
     @property
@@ -473,7 +483,7 @@ class TimeDelta(Parameter):
                          default_units='sec',
                          name=name)
 
-    def get_value_in_default_format(self) -> Union[str, float, None]:
+    def get_value_in_default_units(self) -> Union[str, float, None]:
         return getattr(self._astropy_time_delta, self.default_units)
 
     @property
@@ -563,8 +573,13 @@ class Angle(Parameter):
                          name=name,
                          allowed_units=None)
 
-    def get_value_in_default_format(self) -> Union[str, float, None]:
+    def get_value_in_default_units(self) -> Union[str, float, None]:
         return getattr(self._astropy_angle, self.default_units)
+
+    @staticmethod
+    def get_value_in_units(value, units_in, units_out):
+        astropy_angle = astropyAngle(value, unit=units_in)
+        return getattr(astropy_angle, units_out)
 
     @property
     def value(self):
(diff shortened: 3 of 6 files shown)


tests/test_server_basic.py
```diff
@@ -1526,4 +1526,4 @@ def test_param_value(dispatcher_live_fixture):
     print("content:", c.text)
     jdata=c.json()
 
-    assert jdata['products']['echo']['ang'] == 2.0
\ No newline at end of file
+    assert jdata['products']['echo']['ang'] == 0.0005555555555555556
```
dsavchenko, line 1529: But my ultimate goal was to have 2.0 preserved here

tests/test_unit.py
```diff
@@ -143,11 +143,13 @@ def constructor():
             parameter = constructor()
 
             assert parameter.value == outcome
-            assert parameter.get_value_in_default_format() == outcome_default_format
+            assert parameter.get_value_in_default_units() == outcome_default_format
 
             # setting value during request
             assert parameter.set_par(input_value) == outcome_default_format
-            assert parameter.value == outcome
+            assert parameter.value == outcome_default_format
```
dsavchenko, line 150: So then get_value_in_default_units() becomes redundant. Previous behaviour was totally logical for me.

tests/test_server_basic.py
burnout87: As discussed in #280 , we decided to set the all parameters to their default value.

This discussion started with #227 for the Time parameter, but we forgot to consider the case you just encountered.
In any case we agreed to set by default all the parameters to their default units.

dsavchenko: OK, finally I grasp the reason it is changed to work this way.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -563,9 +578,14 @@ def __init__(self, value=None, units=None, name=None):
                          name=name,
                          allowed_units=None)
 
-    def get_value_in_default_format(self) -> Union[str, float, None]:
+    def get_value_in_default_units(self) -> Union[str, float, None]:
         return getattr(self._astropy_angle, self.default_units)
 
+    @staticmethod
```
dsavchenko, line 584: Why not instance method?

tests/test_unit.py
burnout87: I am adjusting also this.

tests/test_server_basic.py
burnout87: check now, I might have found a better way, but perhaps some more testing is required

cdci_data_analysis/analysis/parameters.py
burnout87: not a particular reason, but we have used `staticmethod` also within the rest of the class

burnout87: also, the intention here is to have a utility function, so `staticmethod` makes sense here

dsavchenko: I am asking just because it is way prettier to use something like `parameter.value_in_units('arcsec')` if one wants to pass the parameter to backend in these units, instead of using internal representation like `parameter._astropy_angle.arcsec` or `Angle.get_value_in_units(parameter.value, parameter.default_units, parameter.units)`

burnout87:
the oda-api test fails since it does not take into account the `echo` product. Once this is merged the relative test will be adjusted accordingly

tests/test_server_basic.py
```diff
@@ -1525,5 +1525,6 @@ def test_param_value(dispatcher_live_fixture):
     assert c.status_code == 200
     print("content:", c.text)
     jdata=c.json()
-
-    assert jdata['products']['echo']['ang'] == 2.0
\ No newline at end of file
+    # TODO notice the difference, is this acceptable?
+    assert jdata['products']['analysis_parameters']['ang'] == 0.0005555555555555556
```
dsavchenko, line 1529: This dictionary is also used when, e.g., generating api_code. But dispatcher expects it in arcsec, not in deg, so the api_code is wrong.
This works for Time, as there is T_format parameter in the request. Introducing units as parameter for every parameter available is definitely overkill.

cdci_data_analysis/analysis/instrument.py
```diff
@@ -270,8 +270,10 @@ def run_query(self, product_type,
         #  this was removed by 2f5b5dfb7e but turns out it is used by some plugins, see test_server_plugin_integral_all_sky
         self._current_par_dic=par_dic
 
-        # set pars values from the input parameters
-        query_out = self.set_pars_from_form(par_dic, verbose=verbose, sentry_client=sentry_client)
+        # # set pars values from the input parameters
+        # query_out = self.set_pars_from_form(par_dic, verbose=verbose, sentry_client=sentry_client)
```
dsavchenko, line 274: So this place is where the second unnecessary conversion occurred!

dsavchenko (changes requested):
What if we add default_units to Angle (and possible others, like Energy) constructor as parameter? 
Or simply do default_units = units in call to super().__init__ there?
Isn't it a solution?

burnout87:
> What if we add default_units to Angle (and possible others, like Energy) constructor as parameter? Or simply do default_units = units in call to `super().__init__` there? Isn't it a solution?

Where exaclty? Because for the `Angle` class, the default_units is already set to `deg`.

cdci_data_analysis/analysis/instrument.py
burnout87: Exactly, we didn't notice during the first refactoring. But now it is quite clear the "damage" it creates.

dsavchenko:
Like this https://github.com/oda-hub/dispatcher-app/commit/4e9a846003df30e4480633456857691b2d9cc176 so you can set defaults when declaring parameter. 
Or like this https://github.com/oda-hub/dispatcher-app/commit/72ccfdda9f573915cb60c2ac3ae6109c6420bdea which is less flexible, but parameters will always work as expected.

tests/test_server_basic.py
burnout87: Good point regarding the api_code. A possible solution could be a "formatting" of the dict to be prepared for the generation of the api_code

burnout87:
> Like this [4e9a846](https://github.com/oda-hub/dispatcher-app/commit/4e9a846003df30e4480633456857691b2d9cc176) so you can set defaults when declaring parameter. 

This is the way I would go for.

> Or like this [72ccfdd](https://github.com/oda-hub/dispatcher-app/commit/72ccfdda9f573915cb60c2ac3ae6109c6420bdea) which is less flexible, but parameters will always work as expected.

This is indeed less flexible, I'd rather choose the first option if we have to make this change.

burnout87:
> What if we add default_units to Angle (and possible others, like Energy) constructor as parameter? 

Currently the `Energy` does not contain any conversion functionality, is that intended? Should it be provided?

dsavchenko:
> Currently the `Energy` does not contain any conversion functionality, is that intended? Should it be provided?

I suspect it was just always in keV before. The conversion functionality is needed. But it's not blocking anything for now, so could be done as another issue

burnout87:
> > Currently the `Energy` does not contain any conversion functionality, is that intended? Should it be provided?
> 
> I suspect it was just always in keV before. The conversion functionality is needed. But it's not blocking anything for now, so could be done as another issue. I can do it, for example.

I will then not apply the default_units change for `Energy` for now.

tests/test_server_basic.py
```diff
@@ -1519,11 +1519,21 @@ def test_param_value(dispatcher_live_fixture):
                            'product_type': 'echo',
                            'query_status': 'new',
                            'query_type': 'Real',
-                           'ang': 2.0},
+                           'ang': 2.0,
+                           'ang_deg': 2.0,
+                           'energ': 2.0},
```
dsavchenko, line 1524:
```suggestion
                           'energ': 2.0,
                           'T1': 57818.560277777775,
                           'T2': 57819.560277777775,
                           'T_format': 'mjd'},
```

tests/test_server_basic.py
```diff
@@ -1519,11 +1519,21 @@ def test_param_value(dispatcher_live_fixture):
...
+    # TODO notice the difference, is this acceptable?
+    assert jdata['products']['analysis_parameters']['ang'] == 2
+    assert jdata['products']['echo']['ang'] == 2
+    # converted in the default units, which for the ang_deg parameter is arcsec
+    assert jdata['products']['analysis_parameters']['ang_deg'] == 7200
+    # in the products it instead remains in deg
+    assert jdata['products']['echo']['ang_deg'] == 2
+
+    assert jdata['products']['analysis_parameters']['energ'] == 2
+    assert jdata['products']['echo']['energ'] == 2
```
dsavchenko, line 1539:
```suggestion
    assert jdata['products']['echo']['energ'] == 2
    
    assert jdata['products']['analysis_parameters']['T1'] == '2017-03-06T13:26:48.000'
    assert jdata['products']['echo']['T1'] == 57818.560277777775
```

cdci_data_analysis/analysis/parameters.py
```diff
@@ -473,9 +480,12 @@ def __init__(self, value=None, delta_T_format='sec', name=None, delta_T_format_n
                          default_units='sec',
                          name=name)
 
-    def get_value_in_default_format(self) -> Union[str, float, None]:
+    def get_value_in_default_units(self) -> Union[str, float, None]:
         return getattr(self._astropy_time_delta, self.default_units)
```
dsavchenko, line 485:
```suggestion

    def get_value_in_default_format(self):
        return self.get_value_in_default_units()

```
To ensure backward compatibility. Also, _format_ for time sounds more natural than _units_

dsavchenko (changes requested):
Just a few suggestions

cdci_data_analysis/analysis/parameters.py
```diff
@@ -447,9 +451,12 @@ def __init__(self, value=None, T_format='isot', name=None, Time_format_name=None
                          default_units='isot',
                          name=name)
 
-    def get_value_in_default_format(self) -> Union[str, float, None]:
+    def get_value_in_default_units(self) -> Union[str, float, None]:
         return getattr(self._astropy_time, self.default_units)
```
dsavchenko, line 456:
```suggestion

    def get_value_in_default_format(self):
        return self.get_value_in_default_units()

```

tests/test_server_basic.py
burnout87: inserted and test passes!

tests/test_server_basic.py
burnout87: inserted and test passes!

dsavchenko (approved):

.github/workflows/python-oda-api.yml
```diff
@@ -56,5 +56,5 @@ jobs:
     
     - name: Test oda_api with pytest
       run: |
-        cd oda_api && DISPATCHER_MOCK_KB=yes python -m pytest tests -sv --full-trace --log-cli-level=DEBUG 
+        cd oda_api && DISPATCHER_MOCK_KB=yes python -m pytest tests -m 'not slow and not test_drupal' -sv --full-trace --log-cli-level=DEBUG
```
volodymyrss, line 59: this is already in master, right?

burnout87: no it is not

cdci_data_analysis/analysis/parameters.py
```diff
@@ -447,9 +455,12 @@ def __init__(self, value=None, T_format='isot', name=None, Time_format_name=None
...
```
volodymyrss, line 459: this could call `get_value_in_units`

cdci_data_analysis/analysis/parameters.py
```diff
@@ -554,18 +568,21 @@ def check_list_value(value, units, name='par'):
...
                          units=units,
                          # TODO can we safely make this assumption?
-                         default_units='deg',
+                         default_units=default_units,
                          name=name,
                          allowed_units=None)
 
-    def get_value_in_default_format(self) -> Union[str, float, None]:
+    def get_value_in_default_units(self) -> Union[str, float, None]:
         return getattr(self._astropy_angle, self.default_units)
```
volodymyrss, line 581: `return self.get_value_in_units(self.default_units)1

.github/workflows/python-oda-api.yml
volodymyrss: Oh! Where is it then? Did it not come from gallery pr?

burnout87: I just added the `-m 'not slow and not test_drupal'` option, and that was not there when inserted this the first time.

burnout87: Whereas the python-oda-api workflow is already in the master, I just added an option to skip certain tests

volodymyrss: test_drupal is testing gallery, right?

burnout87: exactly

volodymyrss: why was it not excluded in gallery PR then?

burnout87: because the gallery related functionalities and tests inside oda_api were introduced later

cdci_data_analysis/analysis/parameters.py
```diff
@@ -302,10 +302,18 @@ def set_value_from_form(self, form, verbose=False):
...
+        return self.get_value_in_default_units()
+
+    def get_value_in_default_units(self):
+        return self.value
 
+    # to ensure backward compatibility
     def get_value_in_default_format(self):
+        return self.get_value_in_default_units()
+
+    def get_value_in_units(self, units):
```
volodymyrss, line 316: why is this function not using any units?

cdci_data_analysis/analysis/parameters.py
```diff
@@ -302,10 +302,18 @@ def set_value_from_form(self, form, verbose=False):
     def set_par(self, value, units=None):
         if units is not None:
             self.units = units
+
         self.value = value
-        return self.get_value_in_default_format()
+        return self.get_value_in_default_units()
+
+    def get_value_in_default_units(self):
+        return self.value
```
volodymyrss, line 310: does it make sense if  this calls `get_value_in_units`?

cdci_data_analysis/analysis/parameters.py
burnout87: Because it's supposed to be extended by each child class of the class `Parameter`

cdci_data_analysis/analysis/parameters.py
burnout87: well in this case yes

volodymyrss:
We have both formats and units. It's relevant, for example, for floats: there are different ways to represent a float, even if it's units are the same. The representation is relevant for embedding in the URL in unique way.

cdci_data_analysis/analysis/parameters.py
volodymyrss: and therefore avoid repeating in inherited classes

cdci_data_analysis/analysis/parameters.py
volodymyrss: if there is no generic way, this function should raise an exception. Currently it does not do what it implies.

.github/workflows/python-oda-api.yml
volodymyrss: but how did the test pass in workflows when the tests in oda_api were introduced?

cdci_data_analysis/analysis/parameters.py
volodymyrss: name of the parameter, "Time" is a bit confusing. It implies TimeStamp, a moment of time.Such an entity has no units, but has a format. Only things that inherit from Float can have units. Methods could be arranged accordingly: unit-specific methods can be moved from `Parameter` to `Float`.

burnout87:
> We have both formats and units. It's relevant, for example, for floats: there are different ways to represent a float, even if it's units are the same. The representation is relevant for embedding in the URL in unique way.

Formats area used also for time parameters (eg `Time` and `TimeDelta`).

For Integer and Floats we also use types.

cdci_data_analysis/analysis/parameters.py
burnout87: Several Parameter(s) do not have an implementation of such a method, thus causing several tests (also in the plugin) to fail.

We can assume to have it returning the very same value in case no implementation is provided.

.github/workflows/python-oda-api.yml
burnout87: probably the drupal gallery tests were not included yet

cdci_data_analysis/analysis/parameters.py
volodymyrss: I think it's because only `Float` parameters should have  units. And also all `Float` parameters have to have units.
The fact that the tests fail while this change appears natural means that the logic is a bit unfortunate.
Maybe moving units to `Float` and inheriting those parameters which have units from `Float` would restore the logic.

burnout87: > Maybe moving units to `Float` and inheriting those parameters which have units from `Float` would restore the logic.

You mean to have parameters like `Angle` to inherit from `Float`? But for certain parameters, that cannot really apply, like `Name`.

cdci_data_analysis/analysis/parameters.py
burnout87: And having the concepts of units/type/format removed from the class Parameter ?

burnout87:
Quite some changes have been applied, let's discuss if we want them and if they make sense

cdci_data_analysis/analysis/parameters.py
```diff
@@ -164,13 +164,23 @@ def __init__(self,
                  value=None,
                  units=None,
                  name: Union[str, None] = None,
+                 # TODO should we remove units/type/format knowledge from the Parameter class?
+
```
volodymyrss, line 168: I suggest to add a docstring for this class. The structure starts to become convoluted.

```suggestion
format:
   every parameter has a format. Format defines representation of the parameter value. 
   Default format of  defines unique representation o
```

dsavchenko:
As agreed, I merge this PR into angle-units branch

Issue #279, opened 2021-11-12
make generic if possible treatment of parameter transformations, with reasonable defaults


volodymyrss:

burnout87:
Various Parameter might need tailored solutions (eg `SpectralBoundary`).

PR #337, opened 2022-01-26
Alternative fix for units problem

dsavchenko:
This simple change fixes https://github.com/oda-hub/dispatcher-app/issues/334. 

But we need to assure it doesn't break something elsewhere

### DIFF
--- a/cdci_data_analysis/analysis/parameters.py
+++ b/cdci_data_analysis/analysis/parameters.py
@@ -303,7 +303,7 @@ class Parameter(object):
         if units is not None:
             self.units = units
         self.value = value
-        return self.get_value_in_default_format()
+        return self.value
 
     def get_value_in_default_format(self):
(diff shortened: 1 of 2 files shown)


cdci_data_analysis/analysis/parameters.py
```diff
@@ -303,7 +303,7 @@ def set_par(self, value, units=None):
         if units is not None:
             self.units = units
         self.value = value
-        return self.get_value_in_default_format()
+        return self.value
```
burnout87, line 306: This is exactly how it used to before, and we changed such behavior [here](https://github.com/oda-hub/dispatcher-app/pull/280/commits/30e2399977e2b1130178de620f813c3b8a666534), in order to solve the #279 .

burnout87 (changes requested):

PR #344, opened 2022-02-03
check if any of the two roles is present

burnout87:
now, any of the two roles, `job manager` and `administrator`, is sufficient


volodymyrss:
thanks. But I think we can resort to only use `job manager` here. Else we could check `administrator` in addition to any other check. `administrator` can set their own role `job manager` if they want to. Role represents certain activity, and that's it. No need to make it more complex!

volodymyrss (approved):

volodymyrss (approved):

PR #346, opened 2022-02-04
catalog normalization function placeholder

burnout87:


burnout87:
@volodymyrss this is a starting point for the catalogs normalization

PR #348, opened 2022-02-07
Astro entity reference

burnout87:
forgot to include a reference to a astrophysical entity, more specifically in this case to the source

### DIFF
--- a/cdci_data_analysis/analysis/drupal_helper.py
+++ b/cdci_data_analysis/analysis/drupal_helper.py
@@ -348,6 +348,22 @@ def post_observation(product_gallery_url, gallery_jwt_token, t1=None, t2=None, s
     return observation_drupal_id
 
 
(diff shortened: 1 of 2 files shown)


cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -348,6 +348,22 @@ def post_observation(product_gallery_url, gallery_jwt_token, t1=None, t2=None, s
     return observation_drupal_id
 
 
+def get_source_astrophysical_entity_id(product_gallery_url, gallery_jwt_token, source_name=None, sentry_client=None) \
```
volodymyrss, line 351: I would suggest to call it more clearly (even if longer)

`get_source_astrophysical_entity_id_by_source_name`

volodymyrss (approved):
thanks!

PR #349, opened 2022-02-08
endpoint to push in a new renku branch

burnout87:


burnout87:
Branches are successfully created, now I will have to generate the appropriate URL to be returned

image

tests/test_server_basic.py
```diff
@@ -1510,6 +1512,55 @@ def test_product_gallery_post_article(dispatcher_live_fixture_with_gallery):
...
+
+    assert c.status_code == 200
+
+    # parse the repo url and build the renku one
+    repo_url = dispatcher_test_conf_with_renku_options['renku_options']['renku_gitlab_repository_url']
+    parsed_repo_url = urlparse(repo_url)
+    namespace = parsed_repo_url.path.split('/')[2]
+    project_name = parsed_repo_url.path.split('/')[-1]
+
+    assert c.text == f'{parsed_repo_url.scheme}://{parsed_repo_url.hostname}/projects/{namespace}/{project_name}/sessions/new?autostart=1&branch=mmoda_request_{job_id}'
```
volodymyrss, line 1561: could you checkout the repo after, and check that the content is valid?

cdci_data_analysis/pytest_fixtures.py
```diff
@@ -393,16 +395,16 @@ def dispatcher_test_conf_with_gallery_fn(dispatcher_test_conf_fn):
...
 
     with open(fn, "w") as f:
         with open(dispatcher_test_conf_fn) as f_default:
             f.write(f_default.read())
 
         f.write('\n    renku_options:'
                 '\n        renku_gitlab_repository_url: "https://renkulab.io/gitlab/gabriele.barni/test-dispatcher-endpoint"'
-                f'\n        renku_gitlab_token_name: "{os.getenv("RENKU_GITLAB_TOKEN_NAME", "token_name")}"'
-                f'\n        renku_gitlab_token: "{os.getenv("RENKU_GITLAB_TOKEN", "token")}"')
+               f'\n        renku_gitlab_token_name: "{os.getenv("RENKU_GITLAB_TOKEN_NAME", "token_name")}"'
```
volodymyrss, line 406: that's `user name` I think

cdci_data_analysis/pytest_fixtures.py
```diff
@@ -730,11 +732,57 @@ def dispatcher_fetch_dummy_products(dummy_product_pack: str, reuse=False):
...
+
+    if branch_name is None:
+        branch_name = 'master'
+
+    url_parsed = urlparse(repository_url)
+
+    if gitlab_token_name is not None and gitlab_token is not None:
+        url_parsed = url_parsed._replace(netloc=f'{gitlab_token_name}:{gitlab_token}@{url_parsed.hostname}')
+
+    repo = Repo.clone_from(url_parsed.geturl(), repo_dir, branch=branch_name)
```
volodymyrss, line 747: most if not all of this would work with test on a filesystem repo

cdci_data_analysis/pytest_fixtures.py
burnout87: no I need the name of the token

volodymyrss: what is "the name of the token", I do not understand the term. Internal token names of github? They should not be used.
It's https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html yes? And you use it in the URL as username, between schema and hostname.

burnout87: Yes it's the personal access token, that has a name assigned at creation.

What do you suggest to use instead?

burnout87: btw I see the security implications, they should not be used within URL

volodymyrss: I would have used ssh key here actually. Or still token but in netrc. So it's not logged.
Maybe this git library can pass the key more covertly.

I would suggest to call it username, because that's what you have to put in it's place. In practice, this username is apparently not currently used https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html . So it does not matter what would be put there.

burnout87: Now uses ssh, and no longer any token. However I needed to provide the 'renku_user_name' so to properly build the output url.

cdci_data_analysis/pytest_fixtures.py
burnout87: next step!

volodymyrss (approved):
if it seems to work, maybe we can just:
* make an issue  to add test with local repo later
* add example in oda_api

burnout87:
> if it seems to work, maybe we can just:

I confirm it, now I am adjusting the formatting, but the link is visualized

> 
>     * make an issue  to add test with local repo later

yes makes sense

> 
>     * add example in oda_api

once this is merged, I will extend the tests in oda_api as well as the example notebook

burnout87:
Not clear to me why anteres test is failing. Any idea @dsavchenko ?

dsavchenko:
> Not clear to me why antares test is failing. Any idea @dsavchenko ?

Test are using real backend container, and it seems that the server for the Antares data (opendata.km3net.de) is currently down.

PR #356, opened 2022-02-14
Optional job id and session

burnout87:


volodymyrss (changes requested):
could you also add "produced_by" argument to mitigate reduction of reproducibility?

and also we'll need an example in oda_api. An issue there?

burnout87:
> 
> and also we'll need an example in oda_api. An issue there?

actually oda_apii will need some extension, I will provided that once the field `produced_by` is implemented

burnout87:
ok the produced by field is now thgere, I am working now on the best formatting for that link

burnout87:
> ok the produced by field is now thgere, I am working now on the best formatting for that link

Output example

image


[http://cdciweb02.isdc.unige.ch/mmoda/gallery/node/1298](http://cdciweb02.isdc.unige.ch/mmoda/gallery/node/1298)

PR #361, opened 2022-02-16
Better error message no api code

burnout87:


volodymyrss (approved):

burnout87:
Sorry for the double-review, I spotted a bug right while you were approving.

volodymyrss:
> Sorry for the double-review, I spotted a bug right while you were approving.

If only a test could catch that! Well, later, hopefully..

volodymyrss (approved):

burnout87:
> > Sorry for the double-review, I spotted a bug right while you were approving.
> 
> If only a test could catch that! Well, later, hopefully..

I was quite an edgy case, but thankfully I noticed it

PR #363, opened 2022-02-17
More test observation product

burnout87:


burnout87:
It now checks an observation is actually assigned when it should or it is not

volodymyrss (dismissed):

volodymyrss (approved):

volodymyrss:
I am not entirely sure the logic of naming variables is currently the most natural. But it's ok for now.

PR #365, opened 2022-02-22
Logging for renku

volodymyrss:


cdci_data_analysis/analysis/renku_helper.py
```diff
@@ -67,22 +71,10 @@ def push_api_code(api_code,
...
-    generated_renku_new_session_url = new_session_autostart_url.format(
-        renku_project_url=renku_project_url,
-        namespace=namespace,
-        project_name=project_name,
-        branch=f'&branch={branch_name}')
-
-    return generated_renku_new_session_url
+    # original_url = repo.remotes.origin.url
+    # TODO: project url could be derived from renku projects base path and original_url. the config values should be renamed then    
+    return f"{renku_project_url}/sessions/new?autostart=1&branch={branch_name}"
```
burnout87, line 76: - The resulting URL does not seem to be working. Is this supposed to be this way for the moment and the configuration are adjusted as you suggested?
-

volodymyrss: The URL is working. We checked. There are even screenshots in, see emails.

There was a confusion between naming things, and some confusion remains.

burnout87: > The URL is working. We checked. There are even screenshots in, see emails.
> 
Yes, I noticed and I saw the configuration you applied, but yes renaming is necessary at this point.

Issue #371, opened 2022-03-21
`resolve_source` endpoint not deployed on staging


burnout87:
This endpoint is not available when I try to use it.

It has been implemented with #369

volodymyrss:
please check, to be sure.

burnout87:
I still get the same error

volodymyrss:
the version which is currently deployed has the latest commit, from 7 days ago. I just re-deployed it, not sure it changes anything.
I suggest we change the version of the app, the one returned here:

https://frontend.staging.odahub.io/mmoda/dispatch-data/run_analysis

it's in cdci_data_analysis/pkg_info.json

that way we can see explicitly what's the version

burnout87:
Ok, I see the problem, the endpoint is there and working, but some configuration is missing, namely `name_resolver_url` and `entities_portal_url` .

Those should be:
``` language:yml
name_resolver_url: "http://cdsweb.u-strasbg.fr/cgi-bin/nph-sesame/-oxp/NSV?"
entities_portal_url: "http://cdsportal.u-strasbg.fr/?target={}"
```

burnout87:
> could you make them fall-back on these by default? I suppose the parsing is tuned to their output? so making them configurable without some modularity is only so much useful.
> 

Ok I will implement that.

> By the way, you could use proxy resolver, which also uses and merges other sources, not available in simbad. Like so:
> 
> [https://resolver-prod.obsuks1.unige.ch/api/v1.1/byname/{}](https://resolver-prod.obsuks1.unige.ch/api/v1.1/byname/%7B%7D)
> 
> 
> It's what is used in frontend.
> 
frontend actually attempts first to query this resolver, and if this fails then it moves on to try the one I am currently using

> 
> I note that your might need other properties but time/location, so moving to internal resolver could be another stage. For now, please just set the defaults to what you expect.

Indeed it might need some further development, so first stage is to apply the defaults, then will extend also with the usage of the other resolver

volodymyrss:
> > could you make them fall-back on these by default? I suppose the parsing is tuned to their output? so making them configurable without some modularity is only so much useful.
> 
> Ok I will implement that.
> 
> > By the way, you could use proxy resolver, which also uses and merges other sources, not available in simbad. Like so:
> > [https://resolver-prod.obsuks1.unige.ch/api/v1.1/byname/{}](https://resolver-prod.obsuks1.unige.ch/api/v1.1/byname/%7B%7D)
[...]

PR #373, opened 2022-03-24
Use local resolver

burnout87:


burnout87:
The resolver now uses the local one, as discussed in #371

volodymyrss (approved):

PR #375, opened 2022-03-29
Revnum service

burnout87:


cdci_data_analysis/analysis/drupal_helper.py
```diff
@@ -718,3 +719,32 @@ def resolve_name(name_resolver_url: str, entities_portal_url: str = None, name:
...
+
+    if res.status_code == 200:
+        resolved_obj['revnum'] = int(res.content)
+    else:
+        logger.warning(f"there seems to be some problem in completing the request for the conversion of the time: {time_to_convert}\n"
+                       f"the request lead to the error {res.text}, "
+                       "this might be due to an error in the url or the service "
+                       "requested is currently not available, "
+                       "please check your request and try to issue it again")
+        raise InternalError('issue when performing a request to the local resolver',
```
volodymyrss, line 747: to timesystem service

volodymyrss (approved):
I am not entirely sure why such a indirection is needed, one could call directly the `timesystem` service from wherever this would be called. 
Could you explain a bit?
But, in principle, ok.

burnout87:
I am using the same approach used for the source name resolution (#369), just that this will be used for the revolution dedicated page that will be built in the product gallery, some discussions also started [here](https://gitlab.astro.unige.ch/oda/product-gallery/doc/-/issues/17).

Yes, a direct call to the service from, for example, the oda_api ,is a much easier to implement approach, but in this way we have only one point where we configure the various services and the related changes of configurations will be easier in the future.

volodymyrss (approved):

burnout87:
Ok I didn't know any of the details related to their implementation of those services, that is why I thought this approach would have created a negligible overhead.

In any case, I will "monitor" the performances over time and if necessary I will move to a more efficient approach as discussed here

burnout87:
Certain tests are failing and it seems to be related to a connection issue, any other ideas on what the reasons could be? I am really struggling

volodymyrss:
> Certain tests are failing and it seems to be related to a connection issue, any other ideas on what the reasons could be? I am really struggling

connection to what? Normally it should not connect to anything..

burnout87:
> > Certain tests are failing and it seems to be related to a connection issue, any other ideas on what the reasons could be? I am really struggling
> 
> connection to what? Normally it should not connect to anything..

It seems that those failing are using the `dispatcher_long_living_fixture` fixture, still I am struggling understanding the reason

volodymyrss:
> > > Certain tests are failing and it seems to be related to a connection issue, any other ideas on what the reasons could be? I am really struggling
> > 
> > 
> > connection to what? Normally it should not connect to anything..
> 
> It seems that those failing are using the `dispatcher_long_living_fixture` fixture, still I am struggling understanding the reason

so to dispatcher? 
Do they run locally? If run one-by-one?

burnout87:
> 
> so to dispatcher? 

Exactly

> Do they run locally? If run one-by-one?

I made a test yesterday removing the `fast` mark to a test, and that passed right away, so actually I am not even sure where the problem lies

burnout87:
> > so to dispatcher?
> 
> Exactly
> 
> > Do they run locally? If run one-by-one?
> 
> I made a test yesterday removing the `fast` mark to a test, and that passed right away, so actually I am not even sure where the problem lies

To be even more precise, the fails happen when running the `fast` tests

burnout87:
Ok it seems to be working now, thanks for the help!

volodymyrss:
> Ok it seems to be working now, thanks for the help!
> 
> But I wonder what the issue was...

There were two issues, both related to update of some dependency versions, most of all flask.
* service url incorrectly detected from new flask output, causing connection refused
* request.json previously returned None if unavailable, but now it was causing exception? rather strange. Does not matter. This was causing 400.

volodymyrss (approved):

burnout87:
> > Ok it seems to be working now, thanks for the help!
> > But I wonder what the issue was...
> 
> There were two issues, both related to update of some dependency versions, most of all flask.
> 
>     * service url incorrectly detected from new flask output, causing connection refused

I actually noticed on my local instance, when running various tests, that the url generated by flask when the application was starting was actually different, so it has to do with the new version of flask indeed

> 
>     * request.json previously returned None if unavailable, but now it was causing exception? rather strange. Does not matter. This was causing 400.

Just yesterday I fixed one import in the `oda_api` related to a json library [here](https://github.com/oda-hub/oda_api/pull/133/commits/dc29c03bc7947336b51ad4126cc4e59d99be7717)

> 
> 
> edit: I squashed commits, to hide all the tries. You might need to be careful pulling.

ok

volodymyrss:
> Just yesterday I fixed one import in the `oda_api` related to a json library [here](https://github.com/oda-hub/oda_api/pull/133/commits/dc29c03bc7947336b51ad4126cc4e59d99be7717)

I am not sure this is related, but it's similar. There is some complexity with different json packages: simplejson et al.

Anyway, now there is a meaningful error with oda_api test.

burnout87:
> 
> Anyway, now there is a meaningful error with oda_api test.

I saw it, it's just a drupal test that should be ignored. In the merge in the oda_api it will be fixed.

burnout87:
>
> Anyway, now there is a meaningful error with oda_api test.

And oda_api tests pass now!

Issue #377, opened 2022-04-04
clone gitlab/renku repo duplicated in fixtures?


volodymyrss:
The staging renku integration does not work. It broke in one of the PRs in the last month, I did not follow enough.
I tried to see if the test works, but found that the clone function is copied in the tests:

https://github.com/oda-hub/dispatcher-app/blob/c6091d0198b2ec4d317ce13fcf4def084464dc4f/cdci_data_analysis/pytest_fixtures.py#L770-L783

While the one which is used is this one:

https://github.com/oda-hub/dispatcher-app/blob/c6091d0198b2ec4d317ce13fcf4def084464dc4f/cdci_data_analysis/analysis/renku_helper.py#L102-L116

So what's tested is not used.
Why would it be so, do you know @burnout87?

After this test works with the code which is actually used in the application, it will be necessary to have some more log about the ssh key (and other parameters) used.

burnout87:
the tests work for me after applying the fix  in #378, would you give it a try and see if the tests pass with that?

volodymyrss:
> the tests work for me after applying the fix in #378, would you give it a try and see if the tests pass with that?

I would check, but could you help me understand why an older version of the clone function is copied in the tests? This way the function which is used live is not tested. So if I understand correctly, the test is not too useful, no?

burnout87:
> I would check, but could you help me understand why an older version of the clone function is copied in the tests?

Unless I understood  wrong, the difference is related to the git_ssh command, which was changed with #365 , but not in the test. 

Also I changed the names of the renku-relaterd configurations, was this change applied on the staging?

volodymyrss:
> > I would check, but could you help me understand why an older version of the clone function is copied in the tests?
> 
> Unless I understood wrong, the difference is related to the git_ssh command, which was changed with #365 , but not in the test.

But why can't you just `from cdci_data_analysis.analysis.renku_helper import clone..repo`? That way you test it. 
This function should not have been copied, but imported. Or was there a special reason to copy it?
Copying leads precisely to the kind issue we have: the two copies diverged, and we are not testing the function used live.

> Also I changed the names of the renku-relaterd configurations, was this change applied on the staging?

you can see the list of staging deployments in the mattermost. So presumably yes.

burnout87:
> > > I would check, but could you help me understand why an older version of the clone function is copied in the tests?
> > 
> 
[...]

PR #378, opened 2022-04-04
fix path test

burnout87:
applying this fix makes the renku dedicated work ( at least for me)

### DIFF
--- a/cdci_data_analysis/pytest_fixtures.py
+++ b/cdci_data_analysis/pytest_fixtures.py
@@ -767,23 +767,6 @@ def dispatcher_fetch_dummy_products(dummy_product_pack: str, reuse=False):
     open(dispatcher_dummy_product_pack_state_fn, "w").write("%s"%time.time())
 
 
-def get_repo_name(repository_url):
-    repo_name = repository_url.split('/')[-1]
-    if repo_name.endswith('.git'):
-        repo_name = repo_name[0:-4]
-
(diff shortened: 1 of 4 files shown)


volodymyrss:
that's cool, but could we also do from `cdci_data_analysis.analysis.renku_helper import clone..repo` as per https://github.com/oda-hub/dispatcher-app/issues/377, so that we test the function which is used live? That way it's more reliable.

cdci_data_analysis/pytest_fixtures.py
```diff
@@ -774,7 +774,7 @@ def clone_gitlab_repo(repository_url, repo_dir=None, renku_gitlab_ssh_key_path=N
     if branch_name is None:
         branch_name = 'master'
 
-    git_ssh_cmd = f'ssh -i {renku_gitlab_ssh_key_path}'
+    git_ssh_cmd = f'ssh -i {renku_gitlab_ssh_key_path} -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no'
```
volodymyrss, line 777: could you just remove this function entirely and only import it?

burnout87: oh yes I was planning to do it, forgot

volodymyrss (approved):

volodymyrss (approved):

Issue #379, opened 2022-04-05
renku push is sometimes failing for existing branches


volodymyrss:
maybe it should be force-push

burnout87:
This is unexpected because in case of existing branch it should return the url, I also wrote a test that covers this specific case.

Which kind of error did you see? Because I just tried to push on Renku after the very same request, and it didn't complain.

volodymyrss:
well it does. Maybe conflicts with existing older versions of the branch.
Are there any disadvantages of force-push?
Also, we should auto-cleanup old branches.

burnout87:
> well it does. Maybe conflicts with existing older versions of the branch. 
Well this could be, though I was not expecting this to happen, especially because I developed everything on a separate personal test repository, but appareantly...

> Are there any disadvantages of force-push? 
No not all, I will implement it.

PR #380, opened 2022-04-06
Force push renku branch, token asked to the user

burnout87:


burnout87:
Example [here](https://renkulab.io/gitlab/gabriele.barni/test-dispatcher-endpoint/-/blob/mmoda_request_a540c2205e73a704/api_code_a540c2205e73a704.ipynb), where the previous version of the api_code was containing the token in clear text.

And that is perhaps the reason why #379 happens, but I am still not 100% sure.

cdci_data_analysis/analysis/renku_helper.py
```diff
@@ -173,11 +170,11 @@ def create_new_notebook_with_code(repo, api_code, job_id, file_name=None):
 def commit_and_push_file(repo, file_path):
     try:
         add_info = repo.index.add(file_path)
-        commit_info = repo.index.commit("commit code from MMODA")
+        commit_info = repo.index.commit("commit code from MMODA " + time.strftime('%Y-%m-%dT%H:%M:%S'))
```
volodymyrss, line 173: normally, time is always already in the commit. So this is duplication.
What's less obvious is user. Maybe we can create commit authored by the user from the token of the request? The commiter will be still set from the config, ODA user.

cdci_data_analysis/analysis/renku_helper.py
```diff
@@ -30,22 +31,18 @@ def push_api_code(api_code,
...
-        job_id_branch_already_exists = check_job_id_branch_is_present(repo, job_id)
+        step = f'checkout branch {branch_name}'
+        repo = checkout_branch_renku_repo(repo, branch_name)
 
-        if not job_id_branch_already_exists:
-            step = f'checkout branch {branch_name}'
-            repo = checkout_branch_renku_repo(repo, branch_name)
+        step = f'removing token from the api_code'
+        token_pattern = r"(\'|\")token(\'|\"):.\s?(\'|\").*?(\'|\"),?"
+        api_code = re.sub(token_pattern, '"token": "<INSERT_YOUR_TOKEN_HERE>",', api_code, flags=re.DOTALL)
```
volodymyrss, line 39: after trying this, I find it would be better if this line is commented-out in the API code. That way the code could be run directly (if allowed)

cdci_data_analysis/analysis/renku_helper.py
burnout87: I wanted to have something generating a different commit message every time, otherwise the commit history would not change, especially now that we force-push always. 

The user is definitely an option, but perhaps I would also add a rand short string.

cdci_data_analysis/analysis/renku_helper.py
burnout87: Like so? [https://renkulab.io/gitlab/gabriele.barni/test-dispatcher-endpoint/-/blob/mmoda_request_9a5e3646291b26da/api_code_9a5e3646291b26da.ipynb](https://renkulab.io/gitlab/gabriele.barni/test-dispatcher-endpoint/-/blob/mmoda_request_9a5e3646291b26da/api_code_9a5e3646291b26da.ipynb)

cdci_data_analysis/analysis/renku_helper.py
volodymyrss: the commit history for each branch should consist of a single ODA commit. Does it not?

burnout87: Ah ok I thought it would have been useful to see the history of the commits.

But then we should not put anything and leave the message as it is (just `commit code from MMODA`) otherwise two requests from two different users might cause the commit history for a branch to have more than one commit.

volodymyrss: Wait, which history? Every new branch should start from master, and have exactly one MMODA commit.

burnout87: > Wait, which history? 

The commit history for a new branch

> Every new branch should start from master, and have exactly one MMODA commit.

It does, but if we change the commit message including the author, two force-push on the same branch create two MMODA message.

volodymyrss: force-push rewrites history. So the resulting branch will be that of the last made request, but only one. The history of the branch before force-push will be lost - so there is never two messages in the same branch.

burnout87: > force-push rewrites history. 

Ok I was not considering this aspect: it rewrites the history on involved files (which for us is the notebook with the api_code, [example](https://renkulab.io/gitlab/gabriele.barni/test-dispatcher-endpoint/-/commits/2fcaf6e112729a8c0b184632d24464ed01f7847d/api_code_9a5e3646291b26da.ipynb) ), but for the branch will be still visible ( [example](https://renkulab.io/gitlab/gabriele.barni/test-dispatcher-endpoint/-/commits/mmoda_request_9a5e3646291b26da) )

Does it make sense or am I missing something?

burnout87: It is now including the username extracted from the token, if this is available (which should always be the case)

tests/test_server_basic.py
```diff
@@ -1810,7 +1810,9 @@ def test_posting_renku(dispatcher_live_fixture_with_renku_options, dispatcher_te
 
     extracted_api_code = DispatcherJobState.extract_api_code(session_id, job_id)
     token_pattern = r"(\'|\")token(\'|\"):.\s?(\'|\").*?(\'|\")"
```
volodymyrss, line 1812: would this work? it's simpler

```suggestion
    token_pattern = r"[\'\"]token[\'\"]:\s*?[\'\"].*?[\'\"]"
```

volodymyrss (approved):

cdci_data_analysis/analysis/renku_helper.py
burnout87: so, is the user name, extracted from the token, in the commit message ok?

tests/test_server_basic.py
burnout87: yes thanks!

volodymyrss (approved):

cdci_data_analysis/analysis/renku_helper.py
volodymyrss: > > force-push rewrites history.
> 
> Ok I was not considering this aspect: it rewrites the history on involved files (which for us is the notebook with the api_code, [example](https://renkulab.io/gitlab/gabriele.barni/test-dispatcher-endpoint/-/commits/2fcaf6e112729a8c0b184632d24464ed01f7847d/api_code_9a5e3646291b26da.ipynb) ), but for the branch will be still visible ( [example](https://renkulab.io/gitlab/gabriele.barni/test-dispatcher-endpoint/-/commits/mmoda_request_9a5e3646291b26da) )
> 
> Does it make sense or am I missing something?

I think there is some confusion about what is rewritten with what.

force push rewrites commit history (including all files, whether some files were added not in first commit is circumstantial) on the remote.
it overwrites the remote history with local history.
these mmoda branches are started from master.
in your example you made new commit on master.
so new branches have this new commit too.

what was in this branch on remote is completely discarded, removing all history.

volodymyrss: > so, is the user name, extracted from the token, in the commit message ok?

I rather meant to set commit author, not so much the message. Both commit author and timestamp are part  of the commit metadata, and are usually shown, e.g. in gitlab

volodymyrss: what could be more useful in the message, is the sort of metadata which is not naturally in the commit.

For example, it could say 
"stored API code of MMODA request for isgri image by Carlo Ferrigno

the original request was through to https://astrooda... .
To retrieve the result, please follow this URL
"

Commit messages can be quite long, multi-line.
I am not sure if URL will comply with restrictions on commit text.

volodymyrss: > > so, is the user name, extracted from the token, in the commit message ok?
> 
> I think I didn't write this previous message of mine properly and therefore was not clear.

Unclear indeed!

I also deduced from this example:

image

Gitlab can show distinction between commiter and author, e.g.:

image

We'd want authored-by the token-user and commited-by MMODA.
Authored-by can be set in git commit command. Commited-by will be set automatically.

> 
> > "stored API code of MMODA request for isgri image by Carlo Ferrigno
> 
> > the original request was through to https://astrooda/... .
> > To retrieve the result, please follow this URL"
> 
> Regarding these information to include inside the commit message, some can be extracted from the token (user name) and some from the request. Is this what you mean with metadata that cannot be found in the token?

Information which can not be found in **commit metadata**, not in token. This information comes from token and request, yes. I suppose you have it right there.

> 
> As for the link you mention, that is the link we also include within the email correct?

yes.

burnout87: For example:

image

The relative gitlab account is also recognized and set.

Out of curiosity, I also made a test and applied a mock author, that does not have a git lab account:

image

burnout87: Better now:
image

burnout87: A first commit with the extended message and the url:

image

volodymyrss:
Oh forgot another related thing which bothered me: when the push failed, there was no sentry alert.
There should be. 
Since it is our problem, not user interaction problem.

cdci_data_analysis/analysis/renku_helper.py
```diff
@@ -170,14 +173,57 @@ def create_new_notebook_with_code(repo, api_code, job_id, file_name=None):
     return file_path
 
 
-def commit_and_push_file(repo, file_path):
+def generate_commit_request_url(params_dic, products_url):
```
volodymyrss, line 176: can you reuse the function used elsewhere? I am sure this URL is constructed simularly in a different place.

burnout87: Yes it is, it's here: [https://github.com/oda-hub/dispatcher-app/blob/9ad3167723a116e4b485606feaf7c90dbc209075/cdci_data_analysis/flask_app/dispatcher_query.py#L1014](https://github.com/oda-hub/dispatcher-app/blob/9ad3167723a116e4b485606feaf7c90dbc209075/cdci_data_analysis/flask_app/dispatcher_query.py#L1014) .

I thought about it but this is an instance method of the `InstrumentQueryBackEnd`, sure we can make an instance of that class, but perhaps we add some overhead and complexity and it might become more complex to maintain.

What do you think?

volodymyrss: > Yes it is, it's here:
> 
> https://github.com/oda-hub/dispatcher-app/blob/9ad3167723a116e4b485606feaf7c90dbc209075/cdci_data_analysis/flask_app/dispatcher_query.py#L1014
> 
> This is an instance method of the `InstrumentQueryBackEnd` class. I thought about using it by making an instance of that class, but perhaps we add some overhead and complexity and it might become more complex to maintain.
> 
> What do you think?

I think `InstrumentQueryBackEnd.generate_products_url_from_par_dict` can be split in two parts:
* part which uses in self.use_scws - to remain in  InstrumentQueryBackEnd
* part which just generates url from par_dict  - to be taken out of the class in some other file.

cdci_data_analysis/analysis/renku_helper.py
```diff
@@ -170,18 +172,42 @@ def create_new_notebook_with_code(repo, api_code, job_id, file_name=None):
...
-        # e.g. push_info = origin.push(refspec='origin:' + str(repo.head.ref))
-        push_info = repo.git.push("--set-upstream", repo.remote().name, str(repo.head.ref))
-        logger.info("push operation complete")
-    except Exception as e:
-        logger.warning(f"something happened while pushing the the file {file_path}, {e}")
-        raise e
+def generate_commit_request_url(params_dic, products_url):
+    # generate the url for the commit message
+    # this is a "default" value for use_scws
+    params_dic['use_scws'] = 'no'
```
volodymyrss, line 178: is this logic not also applicable to the url in email?

burnout87: yes, but for the email the instance member `self.use_scws` is used:

[https://github.com/oda-hub/dispatcher-app/blob/9ad3167723a116e4b485606feaf7c90dbc209075/cdci_data_analysis/flask_app/dispatcher_query.py#L1019-L1026](https://github.com/oda-hub/dispatcher-app/blob/9ad3167723a116e4b485606feaf7c90dbc209075/cdci_data_analysis/flask_app/dispatcher_query.py#L1019-L1026)

volodymyrss: so maybe we can first insert the value from the member in `par_dict`, and then do the logic inside the generate_url function?

volodymyrss: that's just annoying, since it's a delicate logic and it's unfortunate to have it twice.
ok, well.

volodymyrss (dismissed):

volodymyrss (dismissed):

volodymyrss (approved):

PR #387, opened 2022-04-12
sentry alter not submitted status run_query

burnout87:


cdci_data_analysis/analysis/email_helper.py
```diff
@@ -353,25 +353,60 @@ def store_email_api_code_attachment(api_code, status, scratch_dir, sending_time=
...
+        history_info_obj['additional_information'] = additional_info_obj
+    email_history_log_fn = os.path.join(path_email_history_folder, 'email_history_log.log')
+    email_history_log_obj_list = []
+    # Opening JSON list file
+    if os.path.exists(email_history_log_fn):
+        with open(email_history_log_fn, 'r') as openfile:
+            email_history_log_obj_list = json.load(openfile)
+    # append new log object
+    email_history_log_obj_list.append(history_info_obj)
+    with open(email_history_log_fn, 'w') as outfile:
```
volodymyrss, line 373: will this happen with every `run_query`?

cdci_data_analysis/analysis/email_helper.py
```diff
@@ -416,35 +481,82 @@ def is_email_to_send_callback(logger, status, time_original_request, config, job
...
-    return False
+            email_sending_failed = tokenHelper.get_token_user_fail_email(decoded_token)
+            log_additional_info_obj['email_sending_failed'] = email_sending_failed
+            if email_sending_failed:
+                log_additional_info_obj['check_result_message'] = 'the email can be sent'
+            else:
+                log_additional_info_obj['check_result_message'] = 'the email cannot be sent'
+            sending_ok = email_sending_failed
+
+    log_email_sending_info(logger=logger,
```
volodymyrss, line 555: this will happen with every callback, right? This is guaratneed to make the service inoperable, it's often 10 req/second for hours.

Let's make a short log message (no writing files!) in all cases but done/failed/submitted.
Also, better to create new file for every case when the record is actually written (these should be kept few), it's safer.

cdci_data_analysis/analysis/email_helper.py
burnout87: yes, at every call of the `log_email_sending_info` method

volodymyrss: It creates not quite as much load as callback, but still a lot. Could you restrict this to log, unless the email is actually submitted?

burnout87: now, in the case of `run_query`, it logs in case the email cannot be sent

> unless the email is actually submitted?

you mean sent ? so to log in case the email sending fails?

volodymyrss: > now, in the case of `run_query`, it logs in case the email cannot be sent
> 
> > unless the email is actually submitted?
> 
> you mean sent ? so to log in case the email sending fails?

I really mean when it's sent. 
Because the problem now is that several near-identical emails are produced and sent for the same event.
When fails - we get a sentry. Nothing new should be done in this case.

cdci_data_analysis/analysis/email_helper.py
burnout87: > Let's make a short log message (no writing files!) in all cases but done/failed/submitted.

So basically when the email cannot be sent?

volodymyrss: when it **is** sent - write file with explanations why was it sent. that's the problem we face - too many emails.
For callback only "done" should be lead to sending. When callback returns failed/submitted - maybe write too, it does not happen often.

write a reasonably short log message when email is not sent with explanation why.

burnout87: Now record is written on a file when email is actually sent, and some explanation on why an email is not sent has been added.

cdci_data_analysis/analysis/email_helper.py
```diff
@@ -395,19 +424,42 @@ def is_email_to_send_run_query(logger, status, time_original_request, scratch_di
...
-        logger.info("email_sending_job_submitted: %s", email_sending_job_submitted)
         logger.info("interval_ok: %s", interval_ok)
+        log_additional_info_obj['interval_ok'] = interval_ok
+
+        status_ok = True
+        if status != 'submitted':
+            status_ok = False
+            logger.info(f'status {status} not a valid one for sending an email after a run_query')
+            if sentry_client is not None:
+                sentry_client.capture('raven.events.Message',
```
volodymyrss, line 438: oh, could you make a dispatcher config option to disable this, and keep this disabled by default? I am concerned this will also create load.

cdci_data_analysis/analysis/email_helper.py
```diff
@@ -416,35 +468,64 @@ def is_email_to_send_callback(logger, status, time_original_request, config, job
...
 
-    return False
+        # not valid status
+        else:
+            logger.info(f'status {status} not a valid one for sending an email after a callback')
+    else:
+        logger.info(f'an email cannot be sent because a token was not provided')
+
+    if sending_ok:
+        log_additional_info_obj['check_result_message'] = 'the email can be sent'
```
volodymyrss, line 522: "the email will be sent"

cdci_data_analysis/analysis/email_helper.py
```diff
@@ -416,35 +468,64 @@ def is_email_to_send_callback(logger, status, time_original_request, config, job
...
+            email_sending_failed = tokenHelper.get_token_user_fail_email(decoded_token)
+            log_additional_info_obj['email_sending_failed'] = email_sending_failed
+            sending_ok = email_sending_failed
 
-    return False
+        # not valid status
+        else:
+            logger.info(f'status {status} not a valid one for sending an email after a callback')
+    else:
+        logger.info(f'an email cannot be sent because a token was not provided')
```
volodymyrss, line 519: "will not be sent"

cdci_data_analysis/analysis/email_helper.py
```diff
@@ -353,25 +353,53 @@ def store_email_api_code_attachment(api_code, status, scratch_dir, sending_time=
...
+def log_email_sending_info(logger, status, time_request, scratch_dir, job_id, additional_info_obj=None):
+    path_email_history_folder = os.path.join(scratch_dir, 'email_history')
+    if not os.path.exists(path_email_history_folder):
+        os.makedirs(path_email_history_folder)
+    history_info_obj = dict(time=timestamp2isot(time_request),
+                            status=status,
+                            job_id=job_id)
+    if additional_info_obj is not None:
+        history_info_obj['additional_information'] = additional_info_obj
+    email_history_log_fn = os.path.join(path_email_history_folder, f'email_history_log_{time_request}.log')
```
volodymyrss, line 365: this filename is not sufficiently unique. Please add some other fields: e.g. status, and also maybe just add a random hash.

cdci_data_analysis/analysis/email_helper.py
burnout87: you mean to disable this whole logging part (and sentry) in case the status is not valid for sending an email?

cdci_data_analysis/analysis/email_helper.py
burnout87: it includes now status and the hash of the record object

cdci_data_analysis/analysis/email_helper.py
volodymyrss: I meant just this part: sentry when unexpected status is returned. 
But, to be safe, let's just disable this part entirely. 
Only log, not sentry, if other status than "submitted" is returned.

volodymyrss (approved):

PR #389, opened 2022-04-26
job_id parameter for a data_product

burnout87:
…r for the data_product


volodymyrss (approved):

volodymyrss:
I do not really look as closely for this interface as for the other features, but it seems ok.

PR #396, opened 2022-05-09
incident report endpoint and refactoring email_helper.py

burnout87:


burnout87:
An example of an incident report email

image

volodymyrss (approved):
ok I guess,thanks

burnout87:
> ok I guess,thanks

The information included can be further extended in the future if need be

PR #413, opened 2022-06-22
Fetch extra information from dispatcher before `done` email

burnout87:


cdci_data_analysis/pytest_fixtures.py
```diff
@@ -644,6 +656,24 @@ def dispatcher_live_fixture(pytestconfig, dispatcher_test_conf_fn, dispatcher_de
                 
         kill_child_processes(pid, signal.SIGINT)
         os.kill(pid, signal.SIGINT)
+
+
+@pytest.fixture
+def second_dispatcher_live_fixture(pytestconfig, second_dispatcher_test_conf_fn, dispatcher_debug, request):
```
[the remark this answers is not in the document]
burnout87: A multi-threaded dispatcher probably would be the way, but it might take more development.

Also aren't we supposed to have load-balancer, that would assign the request to another dispatcher running instance? If so, in this way is closer to the situation we would have in production/staging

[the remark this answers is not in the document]
volodymyrss: > A multi-threaded dispatcher probably would be the way, but it might take more development.

But did you check if it's not already being multi-threaded? 

Also I do not think it's more development to make sure it's multi-threaded if it is not. It's just more involved development, you'd need to remember how the [dispatcher is started](https://github.com/oda-hub/dispatcher-app/blob/c3d66dda0adf9cdc1ac589b63050f6ae5fee45d5/cdci_data_analysis/pytest_fixtures.py#L466).
 
> Also aren't we supposed to have load-balancer, that would assign the request to another dispatcher running instance? If so, in this way is closer to the situation we would have in production/staging

Dispatchers are running multi-threaded in production. There may be also several of them. But the complex situation occurs when the requests are to another thread, if it occurs, there may be some interaction between the requests. This is the situation which should be tested primarily.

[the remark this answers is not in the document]
burnout87: > 
> But did you check if it's not already being multi-threaded?
> 
> Also I do not think it's more development to make sure it's multi-threaded if it is not. It's just more involved development, you'd need to remember how the [dispatcher is started](https://github.com/oda-hub/dispatcher-app/blob/c3d66dda0adf9cdc1ac589b63050f6ae5fee45d5/cdci_data_analysis/pytest_fixtures.py#L466).
> 

It is started as single threaded https://github.com/oda-hub/dispatcher-app/blob/c3d66dda0adf9cdc1ac589b63050f6ae5fee45d5/bin/run_osa_cdci_server.py#L109

Has the dispatcher ever been used multi-threaded? Do we have experience with that?

> 
> Dispatchers are running multi-threaded in production. There may be also several of them. 

So does that mean that each dispatcher is started with the option `threaded=True` ?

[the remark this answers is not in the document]
volodymyrss: > > But did you check if it's not already being multi-threaded?
> > Also I do not think it's more development to make sure it's multi-threaded if it is not. It's just more involved development, you'd need to remember how the [dispatcher is started](https://github.com/oda-hub/dispatcher-app/blob/c3d66dda0adf9cdc1ac589b63050f6ae5fee45d5/cdci_data_analysis/pytest_fixtures.py#L466).
> 
> It is started as single threaded
> 
> https://github.com/oda-hub/dispatcher-app/blob/c3d66dda0adf9cdc1ac589b63050f6ae5fee45d5/bin/run_osa_cdci_server.py#L109
> 
> Has the dispatcher ever been used multi-threaded? Do we have experience with that?

 
> > Dispatchers are running multi-threaded in production. There may be also several of them.
> 
> So does that mean that each dispatcher is started with the option `threaded=True` ?

Not really, because it's started with gunicorn (of course, since flask server is no good for production). 
A variety of options were tried to optimize load.

I checked and actually, right now, we only run multiprocessing with gunicorn

https://github.com/oda-hub/dispatcher-container/blob/7c5ca47dcb643c85539e245f3109b534669a7bbd/entrypoint.sh#L24-L31

It might be easy to just use this command in the fixture.
Or test with threading, which is more demanding and will be useful.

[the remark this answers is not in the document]
burnout87: > 
> I checked and actually, right now, we only run multiprocessing with gunicorn
> 
> https://github.com/oda-hub/dispatcher-container/blob/7c5ca47dcb643c85539e245f3109b534669a7bbd/entrypoint.sh#L24-L31

Isn't this equivalent to the unicorn-based start within the dispatcher?

https://github.com/oda-hub/dispatcher-app/blob/c3d66dda0adf9cdc1ac589b63050f6ae5fee45d5/bin/run_osa_cdci_server.py#L89-L107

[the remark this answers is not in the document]
burnout87: I am testing with the options `threaded=False`, but it does not seem to be working fully here on github, no issue at all locally on my machine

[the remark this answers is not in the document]
burnout87: It is using now gunicorn

bin/run_osa_cdci_server.py
```diff
@@ -85,28 +80,31 @@ def main(argv=None):
...
     if use_gunicorn is True:
         # let's use the bind options configuration
         dispatcher_bind_host = conf.bind_host
         dispatcher_bind_port = conf.bind_port
-        # dispatcher_url = conf.dispatcher_url
-        # port = conf.dispatcher_port
-
+        dispatcher_url = conf.dispatcher_url
+        port = conf.dispatcher_port
+        # TODO I suspect there is a bug in the gunicorn library
```
volodymyrss, line 91: which bug?

tests/conftest.py
```diff
@@ -5,19 +5,24 @@
...
             dispatcher_long_living_fixture,
+            gunicorn_dispatcher_long_living_fixture,
             dispatcher_test_conf,
             dispatcher_test_conf_with_gallery,
             dispatcher_test_conf_with_gallery_no_resolver,
             dispatcher_test_conf_empty_sentry_fn,
             dispatcher_test_conf_with_gallery_fn,
             dispatcher_test_conf_with_gallery_no_resolver_fn,
             dispatcher_test_conf_fn,
+            second_dispatcher_test_conf_fn,
```
volodymyrss, line 21: not used anymore?

bin/run_osa_cdci_server.py
burnout87: While investigating on why gunicorn was not working (using the StandaloneApplication class), I kept getting the same error:

`dictionary changed size during iteration`

Did some research and one of the suggestion was to use `list` instead `items` as the second returns a view of the dictionary whereas the first one returns a whole new list. But still it didn't help.

That's why I thought of a bug within the gunicorn library, but I cannot know it for sure since then I changed implementation approach.

This should reply also to the previous question.

tests/reference_emails/done-failing.html
```diff
@@ -0,0 +1,81 @@
...
+you receive this message because at 2021-08-19 17:25:52 ( 0.0 seconds ago ) you submitted a request <br>
+for a failing from the service provided by University of Geneva<br>
+available at the URL PRODUCTS_URL.<br>
+<br>
+
+Your request is now <b>ready</b>.<br>
+
+The following additional information have been returned regarding your request:<br>
+failing query
+Instrument: empty, product: failing failed!
```
volodymyrss, line 20: That's just beatiful, reminds of the classic:

image

has to be adapted

tests/conftest.py
burnout87: no...will remove it

bin/run_osa_cdci_server.py
volodymyrss: You should then `copy` the values. List also contains just references. But if it's not used, can the comment reflect that?

burnout87: I can also remove it completely....

tests/reference_emails/done-failing.html
burnout87: It's kind of a funny coincidence:

The error message is built here:

https://github.com/oda-hub/dispatcher-app/blob/master/cdci_data_analysis/analysis/instrument.py#L340-L358

This is then included in the `error_message` field of the `RemoteException`

cdci_data_analysis/analysis/instrument.py
```diff
@@ -253,6 +255,32 @@ def parse_inputs_files(self,
...
+        updated_par_dic['product'] = updated_par_dic['product_type']
+        updated_par_dic['product_type'] = updated_par_dic['query_type']
+        updated_par_dic.pop('query_type')
+
+        logger.info(f"getting products for a more in-depth analysis for the results within run_call_back with args {updated_par_dic}")
+        disp = DispatcherAPI(url=config.dispatcher_callback_url_base, instrument='mock')
+        try:
+            disp.get_product(**updated_par_dic)
+        except RemoteException as re:
+            logger.info('RemoteException detected when retrieving additional information from a completed job')
```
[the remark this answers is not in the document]
burnout87: > I think this is the case when we do not need to send an email, since it will not give user anything actionable. Instead, we should make a sentry report.
> 

So we should include within the email any error details/message? Returned from this call I mean.

>
> Except maybe if this is returned in case of, say, empty result. This should be sent to user with an appropriate, careful, comment. Could you test this empty result case?
>

How would we detect an empty result?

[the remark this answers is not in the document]
volodymyrss: > > I think this is the case when we do not need to send an email, since it will not give user anything actionable. Instead, we should make a sentry report.
> 
> So we should include within the email any error details/message? Returned from this call I mean.

Yes, see example email in test.

> 
> > Except maybe if this is returned in case of, say, empty result. This should be sent to user with an appropriate, careful, comment. Could you test this empty result case?
> 
> How would we detect an empty result?

Check what oda_api raises. Sometimes it's connection error - this should not be sent. Everything that successfully returns from backend but raises an exception in processing is effectivly empty.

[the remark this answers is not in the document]
burnout87: > 
> Check what oda_api raises. Sometimes it's connection error - this should not be sent. Everything that successfully returns from backend but raises an exception in processing is effectivly empty.
>

What do you mean with processing? Processing, by the oda_api, of what has been returned?
There are several exception raised by oda_api, and most of them derive from `RemoteException`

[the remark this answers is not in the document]
volodymyrss: can you distinguish connection error?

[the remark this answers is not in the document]
burnout87: Yes that can be done. However also other exceptions generated within the oda_api can be intercepted, namely:

- `DispatcherNotAvailable`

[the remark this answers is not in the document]
burnout87: Where can I find an example of an empty product? I am trying to design a fake product that would indeed return an empty product for tests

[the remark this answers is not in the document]
volodymyrss: You can put a fake ScW ID:

```bash
$ oda-api get -i isgri -p isgri_image -a scw_list=351700110010.000 -a integral_data_rights=all-private
```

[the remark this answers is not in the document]
burnout87: SO I tried to intercept (almost) all the different sort of exceptions raised by the `oda_api`, until a RemoteException is detected that could possibly indicate an `empty_product`

cdci_data_analysis/analysis/instrument.py
```diff
@@ -253,6 +255,56 @@ def parse_inputs_files(self,
...
+                                               f'{de}'))
+        except ConnectionError as ce:
+            logger.info(f'A connection error exception has been detected when retrieving additional information from a completed job: {ce}')
+            if sentry_client is not None:
+                sentry_client.capture('raven.events.Message',
+                                      message=(f'ConnectionError detected when retrieving additional '
+                                               f'information from a completed job '
+                                               f'{ce}'))
+        except RemoteException as re:
+            logger.info(f'A connection error exception has been detected when retrieving additional information from a completed job: {re}')
```
volodymyrss, line 298: this is really not too useful for the user. See the example instead

volodymyrss: another problem is that `RemoteException` might occur with a correctly completed job with empty result

volodymyrss: > > another problem is that `RemoteException` might occur with a correctly completed job with empty result
> 
> And how can this be detected? By inspecting the content of the exception?

yes

burnout87: The inspection I perform now is based on the output returned by the `oda_api`.

In particular, searching for these two cases:
- https://github.com/oda-hub/oda_api/blob/master/oda_api/api.py#L158-L190 (probably connection error)
- https://github.com/oda-hub/oda_api/blob/master/oda_api/api.py#L813-L832 (probably empty results)

cdci_data_analysis/analysis/email_helper.py
```diff
@@ -268,6 +269,10 @@ def send_job_email(
                                              'attachment',
                                              filename="api_code.py")
 
+    status_details_generated = False
+    if status_details is not None:
```
volodymyrss, line 273: so if it contains exception, it will be sent in the email, and not the nice message?
It should be the nice message and exception details.

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -926,6 +927,15 @@ def run_call_back(self, status_kw_name='action') -> typing.Tuple[str, typing.Uni
                     product_type = original_request_par_dic['product_type']
                 except KeyError as e:
                     raise MissingRequestParameter(repr(e))
+                # get more info regarding the status of the request
+                status_details = None
+                if status == 'done':
+                    # set instrument
+                    self.set_instrument(self.instrument_name)
+                    status_details = self.instrument.get_status_details(par_dic=original_request_par_dic,
```
volodymyrss, line 935: exception in fetching would return something here, yes?
It's true that it's hard to always distinguish from the status_details if it's good.
We would probably need to have some defined structure for the output json. Something that contains {... "status": "empty"/"exception"/"successful" ...} etc

bin/run_osa_cdci_server.py
```diff
@@ -85,28 +80,32 @@ def main(argv=None):
...
-            'threads': 4,
+            'workers': 4,
+            'timeout': 900,
+            'limit-request-line': 0
         }
         if debug:
-            options['loglevel'] = 'debug'
+            options['log-level'] = 'debug'
 
         if True:
```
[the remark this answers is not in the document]
burnout87: just the `if True` or the whole gunicorn one? I thought about it since we dont use it...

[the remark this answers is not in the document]
volodymyrss: yeah, good to cleanup

cdci_data_analysis/analysis/email_helper.py
burnout87: are you referring to a log message? if so that is generated within the dedicated function within the `Instrument` class

volodymyrss: yes. Yes, it is generated there. But what will be sent if the generated message contains exception?

burnout87: in the email it is mentioned that it contains an empty product

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: > exception in fetching would return something here, yes?
>
`status_details` is `None` in case of no exception, Connection related exception and Dispatcher related exception.

burnout87: now a more detailed informative object is included also in the output json

cdci_data_analysis/analysis/email_helper.py
```diff
@@ -268,6 +269,21 @@ def send_job_email(
...
+        if status_details['status'] == 'empty_product':
+            status_details_message = '''Unfortunately, after a quick automated assessment of the request, it has been found that it contains an <b>empty, product</b>.
+To the best of our knowledge, no unexpected errors occurred during processing,
+and if this is not what you expected, you probably need to modify the request parameters. We are sorry.<br>'''
+        elif status_details['status'] == 'empty_result':
+            status_details_message = '''Unfortunately, after a quick automated assessment of the request, it has been found that it contains an <b>empty, result</b>.
+To the best of our knowledge, no unexpected errors occurred during processing,
+and if this is not what you expected, you probably need to modify the request parameters. We are sorry.<br>'''
+        elif status_details['status'] == 'exception' or status_details['status'] == 'connection_error' or status_details['status'] == 'dispatcher_exception':
```
volodymyrss, line 282:
```suggestion
        elif status_details['status'] in ['exception', 'connection_error', dispatcher_exception':    
```

cdci_data_analysis/analysis/email_helper.py
```diff
@@ -268,6 +269,21 @@ def send_job_email(
...
+            status_details_message = '''Unfortunately, after a quick automated assessment of the request, a problem has been found.
+To the best of our knowledge, no unexpected errors occurred during processing,
+and if this is not what you expected, you probably need to modify the request parameters. We are sorry.<br>'''
+
```
volodymyrss, line 286: also why not add `else: raise NotImplementedError`

cdci_data_analysis/analysis/email_helper.py
```diff
@@ -268,6 +270,30 @@ def send_job_email(
                                              'attachment',
                                              filename="api_code.py")
 
+    status_details_message = None
+    status_details_title = status
+    if status_details is not None and status_details['status'] != 'successful':
+        if status_details['status'] == 'empty_product':
+            status_details_message = '''Unfortunately, after a quick automated assessment of the request, it has been found that it contains an <b>empty, product</b>.
```
volodymyrss, line 277: why "empty, product" and not "empty product"?

cdci_data_analysis/analysis/email_helper.py
```diff
@@ -268,6 +270,30 @@ def send_job_email(
...
+To the best of our knowledge, no unexpected errors occurred during processing,
+and if this is not what you expected, you probably need to modify the request parameters. We are sorry.<br>'''
+            status_details_title = 'finished: with empty product'
+        elif status_details['status'] == 'empty_result':
+            status_details_message = '''Unfortunately, after a quick automated assessment of the request, it has been found that it contains an <b>empty, result</b>.
```
volodymyrss, line 282: for user, there is no way to understand what is the difference between empty product and empty result.

cdci_data_analysis/analysis/email_helper.py
```diff
@@ -268,6 +270,30 @@ def send_job_email(
...
+To the best of our knowledge, no unexpected errors occurred during processing,
+and if this is not what you expected, you probably need to modify the request parameters. We are sorry.<br>'''
+            status_details_title = 'finished: with empty result'
+        elif status_details['status'] in ['exception', 'connection_error', 'dispatcher_exception']:
+            status_details_message = '''Unfortunately, after a quick automated assessment of the request, a problem has been found.
```
volodymyrss, line 287: I do not this this should be ever sent to user. The message self-contradicting: problem, but no errors. 
If the error is indeed unexpected, we should get a sentry and that's it - expect our reaction. 

The only thing is that I am not sure if some reasonable cases, some untreated empty outcomes, fall into this category. Maybe let's keep it as sentry only and see.

burnout87: ok I agree, nothing to the user for the moment, but will add a `TODO` for potential future adjustments

cdci_data_analysis/analysis/email_helper.py
volodymyrss: yes. The meaning is the same.

cdci_data_analysis/analysis/email_helper.py
burnout87: `TODO` added

volodymyrss (approved):

volodymyrss:
@dsavchenko please have a look if you manage, else let's merge and see in staging.

dsavchenko (approved):

PR #422, opened 2022-07-31
The region file is shown with the onload syntax in js9

ferrigno:


volodymyrss:
Could you clarify @ferrigno why is this a `bug`? It seems like a feature? Did I misunderstand some intentions?

volodymyrss (approved):

ferrigno:
I put it as bug, because the command that was present before was unable to load the region file and show the regions. It is matter of semantics to define it as bug or feature, I think.

volodymyrss:
> I put it as bug, because the command that was present before was unable to load the region file and show the regions. It is matter of semantics to define it as bug or feature, I think.

Yes, but semantics (as I understand it) is not arbitrary. Bug is something that violates feature which we assume to work. Bug needs to be addressed ASAP, sometimes with a hotfix, or at least included in "Known Issues". Feature is something we know we did not have but would like to have. Feature development is to be scheduled according to the requirements and availability.

In this case it does not matter since it's already implemented and no development should be scheduled or organized. It's just for clarity in general. Here I add https://github.com/oda-hub/doc-handbook/pull/23

PR #427, opened 2022-08-09
using captureMessage Sentry client from Flask library

burnout87:


burnout87:
Currently the only change applied is within the Flask app, for which the change `capture` -> `captureMessage` is applied

volodymyrss (approved):

PR #429, opened 2022-08-09
retries when sending email

burnout87:


cdci_data_analysis/analysis/email_helper.py
```diff
@@ -368,65 +377,82 @@ def send_email(smtp_server,
...
                smtp_server_password,
                logger,
+               sending_time=None,
+               scratch_dir=None,
                attachment=None
                ):
 
     server = None
-    logger.info("Sending email")
+    logger.info(f"Sending email from the smtp server: {smtp_server}:{smtp_port}")
```
volodymyrss, line 386: **through** the SMTP server?

cdci_data_analysis/analysis/email_helper.py
```diff
@@ -368,65 +377,82 @@ def send_email(smtp_server,
...
+            if n_tries_left > 0:
+                logger.warning(f"there seems to be some problem in sending the email with title {email_subject}, "
+                               f"another attempt will be made")
+
+                logger.debug(f"{e} exception while attempting to send the email with title {email_subject}\n"
+                             f"{n_tries_left} tries left, sleeping {email_sending_retry_sleep_s} seconds until retry\n")
+                time.sleep(email_sending_retry_sleep_s)
+            else:
+                logger.warning(f"an issue occurred when sending the email with title {email_subject}, "
+                               f"we are investigating and try to solve the issue as soon as possible")
```
volodymyrss, line 448: "we are investigating" is good to say to users, but it's not much use to write in log that only we can read. This way we tell ourselves we are investigating, which is of course reassuring, but not too useful.

cdci_data_analysis/analysis/email_helper.py
```diff
@@ -368,65 +377,82 @@ def send_email(smtp_server,
...
+
+            return message
+        except Exception as e:
+            n_tries_left -= 1
+
+            if n_tries_left > 0:
+                logger.warning(f"there seems to be some problem in sending the email with title {email_subject}, "
+                               f"another attempt will be made")
+
+                logger.debug(f"{e} exception while attempting to send the email with title {email_subject}\n"
```
volodymyrss, line 443: really this should be "error" or even "exception" (if it looks ok)

cdci_data_analysis/analysis/email_helper.py
```diff
@@ -368,65 +377,82 @@ def send_email(smtp_server,
...
+                             f"{n_tries_left} tries left, sleeping {email_sending_retry_sleep_s} seconds until retry\n")
+                time.sleep(email_sending_retry_sleep_s)
+            else:
+                logger.warning(f"an issue occurred when sending the email with title {email_subject}, "
+                               f"we are investigating and try to solve the issue as soon as possible")
+
+                store_not_sent_email(email_body_html, scratch_dir, sending_time=sending_time)
```
volodymyrss, line 450: I would capture sentry here to be sure. Even if it also caputres elsewhere.

volodymyrss (changes requested):
few comments!

cdci_data_analysis/analysis/email_helper.py
burnout87: but a sentry is already captured at upper levels, we would end up with an additional sentry log that would not add much to what we already receive

volodymyrss: yes, that's why I am saying that I would capture sentry here to be sure. Even if it also captures elsewhere.
Clearly, there is a problem with what sentry captures and where. Maybe there is a mistake somewhere.
So, to be sure, we can also capture here. In the worst case we get two messages with slightly different context.

volodymyrss (approved):

PR #433, opened 2022-08-10
report extra version

volodymyrss:
to give better introspection on the dev versions deployed

### DIFF
--- a/cdci_data_analysis/__init__.py
+++ b/cdci_data_analysis/__init__.py
@@ -24,6 +24,6 @@ _dir=os.path.dirname(__file__)
 with open('%s/pkg_info.json'%_dir) as fp:
     _info = json.load(fp)
 
-__version__ = _info['version']
+__version__ = _info['version'] + os.getenv('DISPATCHER_EXTRA_VERSION', '')
 
 conf_dir = os.path.dirname(__file__)+'/config_dir'


cdci_data_analysis/__init__.py
```diff
@@ -24,6 +24,6 @@
 with open('%s/pkg_info.json'%_dir) as fp:
     _info = json.load(fp)
 
-__version__ = _info['version']
+__version__ = _info['version'] + os.getenv('DISPATCHER_EXTRA_VERSION', '')
```
burnout87, line 27: What does the `DISPATCHER_EXTRA_VERSION` env var contain?

Because it might not be distinguishable between that and the version. Better to add a space in between.

burnout87 (changes requested):

cdci_data_analysis/__init__.py
volodymyrss: space is no good, this better be a word. the variable would also contain the separator, e.g. a dash.
But I changed a bit the approach now.

burnout87: actually I intended a separator like a dash, space is not good indeed

burnout87: where is the `DISPATCHER_EXTRA_VERSION` var set?

volodymyrss: in the deployment

burnout87 (approved):

Issue #426, opened 2022-08-08
refactor sentry usage


volodymyrss:
as pointed out in #425 

edit: we could use https://pypi.org/project/pytest-sentry/ also

volodymyrss:
refactor not actually done yet.

Also we need to move to sentry-sdk:

https://docs.sentry.io/platforms/python/

raven is old, see here:

https://docs.sentry.io/clients/python/

it says

```
Note

A new Python SDK has superseded this deprecated version. Sentry preserves this documentation for customers using the old client. We recommend using the [updated Python SDK](https://docs.sentry.io/platforms/python/) for new projects.
```

This deprectated API might cause issues sending some events

PR #434, opened 2022-08-10
Hotfix manydone

volodymyrss:


volodymyrss:
this will be tested when we use sentry test in #426

burnout87 (approved):

PR #435, opened 2022-08-11
using sentry-sdk, init and capture message, removing raven

burnout87:


volodymyrss (changes requested):
ok, could you also add https://pypi.org/project/pytest-sentry/? just in one case, to see that it's working

burnout87:
> ok, could you also add https://pypi.org/project/pytest-sentry/? just in one case, to see that it's working

There's now a dedicated fixture that setup the sentry-sdk `Hub`, and used, for now, within context of a test that explicitly make an email sending fail

burnout87:
I can see the sentry notifications captuerd and raised

volodymyrss (approved):

PR #447, opened 2022-08-30
refresh token endpoint

burnout87:
… of the token payload


volodymyrss:
could be good to add a role to allow only some account tokens to be refreshed.

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -789,12 +789,17 @@ def validate_job_id(self, request_parameters_from_scratch_dir=False):
...
     # potentially this can be extended to support more modification of the token payload (e.g. roles)
-    def update_token(self, update_email_options=False):
+    def update_token(self, update_email_options=False, refresh_token=False):
 
         if update_email_options:
             self.token = tokenHelper.update_token_email_options(self.token, self.app.config.get('conf').secret_key,
                                                                 self.restricted_par_dic(self.par_dic))
 
+        if refresh_token:
+            refresh_interval = int(self.restricted_par_dic(self.par_dic).get('refresh_interval', None))
```
volodymyrss, line 799: it should not be allowed to refresh to get an very long-live token, there should be a limit (e.g. 1 week)
and there should be a reasonable default (say, also 1 week)

also, only tokens with some special role should be refreshable, e.g. "refresh-tokens"

volodymyrss (changes requested):

volodymyrss (approved):

Issue #444, opened 2022-08-29
use parent classes to determine both the type (like ISOT) and the other use (like StartTime)


volodymyrss:
we can parse https://odahub.io/ontology/ontology.ttl with rdflib

PR #450, opened 2022-08-31
accept any keyword in specific parameter constructor

dsavchenko:
to avoid exception in from_owl_uri


burnout87 (dismissed):

volodymyrss:
> to avoid exception in from_owl_uri

that's good, but it should not probably be completely quiet. We want to avoid accidentally constructing something with loss of information.

Test here would be useful, simple, and also demonstrate well what we intend.

volodymyrss (changes requested):
need test to demonstrate the intent

dsavchenko:
I reverted previous changes, will do it in a better way.
The test demonstrates initial intent.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -397,77 +401,38 @@ def matches_owl_uri(cls, owl_uri: str) -> bool:
...
+
         for x in subclasses_recursive(cls):
             logger.debug("searching for class with owl_uri=%s, found %s", owl_uri, x)
             if x.matches_owl_uri(owl_uri):
                 logger.info("will construct %s by url %s", x, owl_uri)
+                call_kwargs = kwargs.copy()
+                call_signature = signature(x)
+                for par_name in kwargs.keys():
+                    if par_name not in call_signature.parameters:
+                        logger.warning("parameter %s with value %s not used to construct %s", par_name, kwargs[par_name], x)
```
volodymyrss, line 417: can we trigger this in the test? when is it even expected to be triggered?

volodymyrss (dismissed):

cdci_data_analysis/analysis/parameters.py
dsavchenko: Actually, it is not triggered "in the wild". Previously the wrong parameters used to be passed. I added test to trigger it. 

But probably, it is not needed and the exception should be raised...

volodymyrss: Yeah it's a bit uncertain. Oh, could we try to parse some real output? Store it in test data and parse it? As an additional test.

burnout87 (approved):

cdci_data_analysis/analysis/parameters.py
dsavchenko: You mean what output? The tests in nb2workflow plugin use real output. But this is irrelevant here, as we need to test only one method `from_owl_url` 
Previously, the exception was caused by bad code, it will not be replicated currently with any data.

Accepting unneeded parameters could potentially simplify future code. On the other hand, it could be more error-prone.

burnout87: perhaps I would extend this warning message also mentioning that the parameter `par_name` won't be used for the instantiation of the object

volodymyrss: yeah so let's continue there https://github.com/oda-hub/dispatcher-plugin-nb2workflow/issues/23#issuecomment-1241649054

volodymyrss: > > perhaps I would extend this warning message also mentioning that the parameter `par_name` won't be used for the instantiation of the object
> 
> But it will be.

I am lost what on we are talking about. `call_kwargs` is used to construct, and you pop from it, right? And there is a warning?

dsavchenko: Sorry, I've read wrongly, confused myself, then confused everyone. I've removed this comment

burnout87: my suggestion was to extend the log message mentioning that the object will be built without using `par_name` since this is not part of the signature, I also was not clear enough probably

dsavchenko: > my suggestion was to extend the log message mentioning that the object will be built without using `par_name` since this is not part of the signature, I also was not clear enough probably

Could you propose the exact phrasing?

burnout87: `parameter %s, with value %s, is not used to construct a %s object, therefore this will be discarded for the instantiation`

cdci_data_analysis/analysis/parameters.py
```diff
@@ -397,77 +401,38 @@ def matches_owl_uri(cls, owl_uri: str) -> bool:
...
-                            par_default_format=par_default_format,
-                            par_format_name=par_format_name,
-        
-                            default_type=default_type,
-                            allowed_types=allowed_types,
-        
-                            check_value=check_value,
-                            allowed_values=allowed_values)
+                    parameter = x(**call_kwargs)
                     break
```
volodymyrss, line 421: I am a bit concerned what happens when there are multiple matches. In principle, it gets the most generic one, but maybe it's good to warn too.

dsavchenko: I think that we can just avoid multiple matches. Subclasses should have more specific uris than parents. 
What's exactly selected is related to https://github.com/oda-hub/dispatcher-app/issues/444

volodymyrss: we'll be dealing with these sort of challenges as the ontology improves, that's why I left todo's instead of invasive changes

volodymyrss (approved):

PR #451, opened 2022-09-01
plugin reload

dsavchenko:


volodymyrss (changes requested):
looks reasonable. Could you make a test please?

burnout87 (changes requested):
I agree, having a simple test is a good idea

dsavchenko:
I couldn't guess the way to test it here. For e.g. `empty` instrument, how to check, if it was actually reloaded? Maybe you could propose something

In the use case for the https://github.com/oda-hub/dispatcher-plugin-nb2workflow/ it's clear, but it should be tested in the plugin.

volodymyrss:
> I couldn't guess the way to test it here. For e.g. `empty` instrument, how to check, if it was actually reloaded? Maybe you could advise something.

you'd probably need to modify some of the mock plugins to optionally reveal more than one instrument. What do you think @burnout87, you made much of these test plugins ? 

> In the use case for the https://github.com/oda-hub/dispatcher-plugin-nb2workflow/ it's clear, but it should be tested in the plugin.

maybe the test should be there as well, but since the feature is in dispatcher, it's nicer if there is a self-sufficient test in the dispatcher too

dsavchenko:
@burnout87 see this https://github.com/oda-hub/dispatcher-plugin-nb2workflow/blob/5f66be9ab88f18c56e3adc0a143b2521df0549f0/tests/test_plugin.py#L123 test

burnout87:
> @burnout87 see this https://github.com/oda-hub/dispatcher-plugin-nb2workflow/blob/5f66be9ab88f18c56e3adc0a143b2521df0549f0/tests/test_plugin.py#L123 test

Ok thanks...those fake instruments provided built-in in the dispatcher (`empty`, `empty-async` etc etc) do not effectively belong to any plugin but are just added to the list of instruments solely for testing purposes.

An option could be to build a simple fake plugin that includes those instruments (or just some of them), probably just by providing a simple fake `exposer` module would do the job.

I can give do it and we can see if it helps.

volodymyrss:
> > @burnout87 see this https://github.com/oda-hub/dispatcher-plugin-nb2workflow/blob/5f66be9ab88f18c56e3adc0a143b2521df0549f0/tests/test_plugin.py#L123 test
> 
> Ok thanks...those fake instruments provided built-in in the dispatcher (`empty`, `empty-async` etc etc) do not effectively belong to any plugin but are just added to the list of instruments solely for testing purposes.
> 
> An option could be to build a simple fake plugin that includes those instruments (or just some of them), probably just by providing a simple fake `exposer` module would do the job.
> 
> I can give do it and we can see if it helps.

Ok, but let's make sure this does not take a lot of time. If it does, it could be better to make a simple test in dispatcher (like just to check that the endpoint responds and does not do anything crazy) and do the testing in plugin. Anyway the plugin is always tested (#454) and this feature is required only for this plugin.

volodymyrss (approved):

burnout87 (approved):

PR #455, opened 2022-09-06
dummy plugin and relative reload test

burnout87:
here an empty plugin with a relative test

surely not the neatest solution, so open to suggestions for improvements

### DIFF
--- a/cdci_data_analysis/flask_app/app.py
+++ b/cdci_data_analysis/flask_app/app.py
@@ -206,9 +206,12 @@ def common_exception_payload():
     plugins = {}
     payload['config']['plugins'] = plugins
     for plugin_name, plugin_module in importer.cdci_plugins_dict.items():
-        plugins[plugin_name] = {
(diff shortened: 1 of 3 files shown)


volodymyrss (approved):
nice, seems ok for me, but I did not look deep.

dsavchenko:
The functionality and it's mechanics do not really mimic the actual usecase, so it looks for me like unneded modifications. 
Still ok, because I don't currently have another solution (but I will try)

burnout87:
> The functionality and it's mechanics do not really mimic the actual usecase, so it looks for me like unneded modifications. Still ok, because I don't currently have another solution (but I will try)

You're right but it would probably take a lot more time to build an entire `dummy` plugin that would allow us to fully reproduce the same behavior of re-loading of a standard plugin

volodymyrss:
so we do not need this? @burnout87

burnout87:
No, can be closed as #457 provides a more complete approach

PR #457, opened 2022-09-06
dummy plugin and relative reload test (alternative solution)

dsavchenko:


dsavchenko:
The test for integral plugin should pass with https://github.com/oda-hub/dispatcher-plugin-integral/pull/100

burnout87 (approved):

dsavchenko:
So I propose to merge this approach, then the reload-plugin brunch will be ready to merge into master

volodymyrss (approved):

Issue #443, opened 2022-08-29
combine two warning kinds


volodymyrss:
try to avoid always reporting all default parameters for all instruments

burnout87:
Is this related to  https://github.com/oda-hub/oda_api/issues/152

volodymyrss:
> Is this related to [oda-hub/oda_api#152](https://github.com/oda-hub/oda_api/issues/152) ?

it's related.

see how antares returns a warning because T1/T2 are not used.

Similar warning should be sent for every parameter which is not used.

what oda_api does is an independent check. It is not exceedingly useful, but let's keep it for a bit.
Instead, oda_api should make sure to show the warning returned by dispatcher (like frontend does)

PR #462, opened 2022-09-14
Detect unknown arguments in the request and warn about them; optionally remove

dsavchenko:


tests/test_server_basic.py
```diff
@@ -2322,3 +2322,25 @@ def test_param_value(dispatcher_live_fixture):
...
+                           'unknown': 2.0},
+                  )
+    
+    assert c.status_code == 200
+    print("content:", c.text)
+    jdata=c.json()
+    
+    assert re.match(r'Please note that arguments?.*unknown.*not used', jdata['exit_status']['comment'])
+    assert 'T_format' not in jdata['exit_status']['comment']
+    #assert 'unknown' not in jdata['products']['analysis_parameters'].keys()
```
burnout87, line 2345: Were you also planning to add a test for the `job_id`?

dsavchenko: If the arguments are not removed from par_dic then job_id is calculated as before, and available tests cover this

tests/test_server_basic.py
```diff
@@ -41,16 +41,16 @@
                     query_type="Real",
                     instrument="isgri",
                     product_type="isgri_image",
-                    osa_version="OSA10.2",
-                    E1_keV=20.,
-                    E2_keV=40.,
+                    # osa_version="OSA10.2",
+                    # E1_keV=20.,
+                    # E2_keV=40.,
```
burnout87, line 46: Will you uncomment those back? I am not sure this won't brake any test...

cdci_data_analysis/analysis/instrument.py
```diff
@@ -139,41 +142,63 @@ def set_pars_from_dic(self, par_dic, verbose=False):
...
+                                'api',
+                                'oda_api_version',
+                                'off_line',
+                                'job_id',
+                                'async_dispatcher',
+                                'allow_unknown_args'] + self.get_arguments_name_list()
+        self.unknown_arguments_name_list = []
+        for k in list(updated_arg_dic.keys()):
+            if k not in known_argument_names:
+                if not (self.allow_unknown_arguments or arg_dic.get('allow_unknown_args', 'False') == 'True'):
```
burnout87, line 194: Why can the `allow_unknown_args` arg be found in the input dic `arg_dic`?

Can the `allow_unknown_args` member be assigned directly from the `arg_dic, if present?

burnout87 (changes requested):

tests/test_server_basic.py
dsavchenko: These are integral-specific parameters, which are not really used (they are not defined in the empty* instruments, which are used in every test of this file). That's why I commented them, otherwise, many tests will break with this new proposed behavior of filtering unknown arguments.
Like this, all tests pass.

cdci_data_analysis/analysis/instrument.py
dsavchenko: I made two ways of allowing undefined arguments: either instrument-wide by self.allow_unknown_arguments or by passing allow_unknown_args=True in request. The former is not used, though. 

> Can the `allow_unknown_args` member be assigned directly from the `arg_dic` entry, if present?
I'm not sure, I understand your question

burnout87: > > Can the `allow_unknown_args` member be assigned directly from the `arg_dic` entry, if present?
> I'm not sure, I understand your question

Basically:
```suggestion
                self.allow_unknown_arguments = arg_dic.get('allow_unknown_args', 'False')
```
during the set_pars_from_dic method

tests/test_server_basic.py
burnout87: which tests dont pass?

dsavchenko: Those where job_id is being compared with the one, calculated in test using make_hash on params

burnout87: Then those tests should be modified accordingly: by removing those entries (eg E1, E2 etc) from the dict used to calculate the `job_id`for the comparison

burnout87 (approved):

cdci_data_analysis/analysis/instrument.py
```diff
@@ -63,6 +63,18 @@
 # TODO: this is not preserved between requests, and is not thread safe. Why not pass it in class instances?
 params_not_to_be_included = ['user_catalog',]
 
+general_use_args = ['instrument',
```
volodymyrss, line 66: word "general" is to general. if there is no less general word, could you please add a comment above explaining it's use?

cdci_data_analysis/analysis/instrument.py
```diff
@@ -79,7 +91,8 @@ def __init__(self,
                  data_serve_conf_file=None,
                  product_queries_list=None,
                  data_server_query_class=None,
-                 query_dictionary={}):
+                 query_dictionary={},
+                 allow_unknown_arguments=False):
```
volodymyrss, line 95: that's nice, for compatibility

cdci_data_analysis/analysis/instrument.py
```diff
@@ -105,6 +118,8 @@ def __init__(self,
         self.input_product_query=input_product_query
 
         self.query_dictionary = query_dictionary
+        
+        self.allow_unknown_arguments = allow_unknown_arguments
```
volodymyrss, line 122: we have to write somewhere in a comment what's the difference between `argument` and `parameter`

cdci_data_analysis/analysis/instrument.py
```diff
@@ -139,41 +154,53 @@ def set_pars_from_dic(self, par_dic, verbose=False):
...
-        return updated_par_dic
+        self.allow_unknown_arguments = self.allow_unknown_arguments or arg_dic.get('allow_unknown_args', 'False') == 'True'
+        known_argument_names = general_use_args + self.get_arguments_name_list()
+        self.unknown_arguments_name_list = []
+        for k in list(updated_arg_dic.keys()):
+            if k not in known_argument_names:
+                if not self.allow_unknown_arguments:
+                    updated_arg_dic.pop(k) 
+                    self.logger.warning("argument '%s' is in the request but not used by instrument '%s', removing it", k, self.name)
+                    self.unknown_arguments_name_list.append(k)
```
volodymyrss, line 199: why set attribute if return is used for the principal structure?

cdci_data_analysis/analysis/parameters.py
```diff
@@ -205,6 +205,7 @@ def __init__(self,
                  allowed_values=None,
                  **kwargs
                  ):
+
```
volodymyrss, line 208: ?

cdci_data_analysis/analysis/parameters.py
```diff
@@ -248,6 +249,14 @@ def __init__(self,
...
         self.par_format_name=par_format_name
         self.value = value
+        
+        self._arg_list = [self.name]
+        if par_format_name is not None:
+            self._arg_list.append(par_format_name)
+        
+    @property
+    def argument_names_list(self):
+        return self._arg_list
```
volodymyrss, line 259: potentially dangerous to return mutable member

cdci_data_analysis/analysis/parameters.py
```diff
@@ -391,8 +400,10 @@ def check_value(val, units=None, name=None, par_format=None):
         pass
 
     def reprJSON(self):
-        return dict(name=self.name, units=self.units, value=self.value)
-    
+        reprjson = [dict(name=self.name, units=self.units, value=self.value)]
+        if self.par_format_name is not None:
+            reprjson.append(dict(name=self.par_format_name, units="str", value=self.par_format))
+        return reprjson
```
volodymyrss, line 406: to be clear, json is a string generated from this object. I find sometimes confusion about this, also in the dispatcher, resulting in occasional `try: dat=json.loads(data): except: pass`.

cdci_data_analysis/analysis/queries.py
```diff
@@ -246,9 +246,13 @@ def get_parameters_list_as_json(self,prod_dict=None):
...
         for par in self._parameters_list:
-            l.append(par.reprJSON())
-
-        return json.dumps(l)
+            l.extend(par.reprJSON())
+        # remove duplicates like T_format
+        seen = set()
+        l1 = [x for x in l if (x.get('name') is None) or not (x.get('name') in seen or seen.add(x.get('name')))]
+        
+        print(l1)
```
volodymyrss, line 254: logging?

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -1106,6 +1107,13 @@ def build_dispatcher_response(self,
         if query_out is not None:
             out_dict['products'] = query_out.prod_dictionary
             out_dict['exit_status'] = query_out.status_dictionary
+            if getattr(self.instrument, 'unknown_arguments_name_list', []):
+                if len(self.instrument.unknown_arguments_name_list) == 1:
+                    comment = f'Please note that argument {self.instrument.unknown_arguments_name_list[0]} is not used'
+                else:
+                    comment = f'Please note that arguments {", ".join(self.instrument.unknown_arguments_name_list)} are not used'
+                out_dict['exit_status']['comment'] = \
+                    out_dict['exit_status']['comment'] + ' ' + comment if out_dict['exit_status']['comment'] else comment
```
volodymyrss, line 1116: do we not have a warning field too?

tests/test_server_basic.py
```diff
@@ -1420,6 +1428,7 @@ def test_default_values(dispatcher_live_fixture, additional_parameter):
 
     if additional_parameter:
         params['additional_param'] = 'no_value'
+        params['allow_unknown_args'] = True
```
volodymyrss, line 1431: do we have test when it's not allowed?

volodymyrss (changes requested):

cdci_data_analysis/analysis/instrument.py
dsavchenko: possibly "non_parameter_args"?

cdci_data_analysis/analysis/instrument.py
volodymyrss: but anyway

cdci_data_analysis/flask_app/dispatcher_query.py
dsavchenko: We do. We manually putted the message about unused arguments (e.g. coords for GW) into the comment field, so I used it here also.

tests/test_server_basic.py
dsavchenko: In this specific test, it doesn't make sense, but it's not allowed by default. 
In test_unknown_argument the behavior is tested directly (but need to uncomment). There are also changes for several tests where job_id is asserted

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: The comment field is used within the frontend to be visualized with the yellow background. We had discussed here

https://github.com/oda-hub/dispatcher-app/issues/443#issuecomment-1233034425

cdci_data_analysis/analysis/instrument.py
```diff
@@ -63,6 +63,22 @@
...
+                      'token',
+                      'api',
+                      'oda_api_version',
+                      'off_line',
+                      'job_id',
+                      'async_dispatcher',
+                      'allow_unknown_args']
+# NOTE: arguments are passed in the request to the dispatcher
+# some arguments are used to set the values of the analysis parameters
+# the parameter is a subclass of Parameter and may use several arguments to set it's value
```
volodymyrss, line 80: cool, but there is a bit more to it.
something should be said about Parameter (and their subclasses) instances,which have names.
Notable that several parameters, even built from from different classes may use some of the same arguments.

cdci_data_analysis/analysis/instrument.py
```diff
@@ -139,41 +158,53 @@ def set_pars_from_dic(self, par_dic, verbose=False):
...
 
             self.logger.info("after normalizing, set_pars_from_dic>> par: %s par.name: %s par.value: %s par_dic[par.name]: %s",
-                             par, par.name, par.value, par_dic.get(par.name, None))
+                             par, par.name, par.value, arg_dic.get(par.name, None))
 
             if par.name == "scw_list":
                 self.logger.info("set_pars_from_dic>> scw_list is %s", par.value)
 
-        return updated_par_dic
+        self.allow_unknown_arguments = self.allow_unknown_arguments or arg_dic.get('allow_unknown_args', 'False') == 'True'
```
volodymyrss, line 195: it seems intentional that if the attribute is already set to True, the argument is ignored. But why?

cdci_data_analysis/analysis/parameters.py
```diff
@@ -391,8 +400,11 @@ def check_value(val, units=None, name=None, par_format=None):
         pass
 
     def reprJSON(self):
-        return dict(name=self.name, units=self.units, value=self.value)
-    
+        # produces json-serialisable list
```
volodymyrss, line 403: ok, but could you name the function something like "reprJSONifiable"? I know it's long but it's less likely to cause mistakes.

cdci_data_analysis/analysis/instrument.py
```diff
@@ -554,11 +585,24 @@ def get_parameters_name_list(self, prod_name=None):
...
+        for par in self.get_parameters_list(prod_name = prod_name):
+            l.extend(par.argument_names_list)
+        l = list(dict.fromkeys(l)) # remove duplicates preserving order
+        return l
+        
+    def get_parameters_name_list(self, prod_name=None):
+        l = []
+        for par in self.get_parameters_list(prod_name = prod_name):
+            l.append(par.name)
+        # TODO: what about duplicates here?
```
volodymyrss, line 604: it's not really a problem I guess, not sure. at least add a warning?

cdci_data_analysis/analysis/queries.py
```diff
@@ -246,9 +246,13 @@ def get_parameters_list_as_json(self,prod_dict=None):
         l=[ {'query_name':self.name}]
 
         for par in self._parameters_list:
-            l.append(par.reprJSON())
-
-        return json.dumps(l)
+            l.extend(par.reprJSON())
+        # remove duplicates like T_format
+        seen = set()
+        l1 = [x for x in l if (x.get('name') is None) or not (x.get('name') in seen or seen.add(x.get('name')))]
```
volodymyrss, line 252: I am not sure this is the most readable way to do it. It's not like the goal is to minimize the number of lines. Also the line is a bit too long.

cdci_data_analysis/analysis/queries.py
```diff
@@ -344,10 +348,14 @@ def get_parameters_list_as_json(self,prod_dict=None):
...
         for par in self._parameters_list:
-            l.append(par.reprJSON())
-
-        print(l)
-        return json.dumps(l)
+            l.extend(par.reprJSON())
+        
+        # remove duplicates like T_format
+        seen = set()
+        l1 = [x for x in l if (x.get('name') is None) or not (x.get('name') in seen or seen.add(x.get('name')))]
```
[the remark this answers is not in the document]
volodymyrss: or call parent class if possible?

cdci_data_analysis/analysis/queries.py
```diff
@@ -344,10 +348,14 @@ def get_parameters_list_as_json(self,prod_dict=None):
...
+        
+        print(l1)
```
volodymyrss, line 357: another print - fixed in parent class but not here. duplication bites!

cdci_data_analysis/flask_app/dispatcher_query.py
volodymyrss: yeah that's what I meant. there is another comment field, behaves like comment but more alarming.

tests/test_server_basic.py
volodymyrss: yeah I meant elsewhere. Uncomment is good.

volodymyrss (changes requested):

cdci_data_analysis/analysis/instrument.py
burnout87: true, I hadn't seen it in that way...perhaps it is better to set it with the value within the `arg_dic` if present, otherwise it doesn' change...but it depends on which one has the "priority"...

cdci_data_analysis/analysis/parameters.py
dsavchenko: I used the existing function name. But ok, it's better to change

volodymyrss: Yeah, and approach like that probably caused dejsonifcation by trail in some dispatcher and oda_api places...

cdci_data_analysis/flask_app/dispatcher_query.py
dsavchenko: We used 'comment' field before to warn about unused parameters, so why change it?

volodymyrss: it does seem like warning should go into warning field.
but ok, this is more of a comment than a warning. It does not break the result, just means that the request is redundant.

volodymyrss (approved):
let's just not forget of this https://github.com/oda-hub/oda_api/issues/152 which will show these warnings

PR #463, opened 2022-09-15
Obsid and attachments observation

burnout87:


burnout87:
Will include rest endpoint to upload an observation

volodymyrss (approved):

PR #473, opened 2022-11-02
searching the token from the api_code and using it in the output link

burnout87:


burnout87:
to address https://github.com/oda-hub/smartsky/issues/66

volodymyrss (changes requested):

volodymyrss:
once complete tests pass, it needs cautious review since it's security-sensitive

cdci_data_analysis/analysis/renku_helper.py
```diff
@@ -147,7 +150,7 @@ def check_job_id_branch_is_present(repo, job_id):
     r = re.compile(f".*_{job_id}")
     filtered_list = list(filter(r.match, list_branches))
 
-    return len(filtered_list) == 1
```
volodymyrss, line 150: so this is only used for test, right?

why do we end up with >= 1 now, several sessions? Would several different sessions for the same jobid have the same exact content?

burnout87: > so this is only used for test, right?
> 
yes!

> why do we end up with >= 1 now, several sessions? Would several different sessions for the same jobid have the same exact content?
>

Because there could be the autosave ones:

image

tests/test_server_basic.py
```diff
@@ -2704,15 +2704,16 @@ def test_posting_renku(dispatcher_live_fixture_with_renku_options, dispatcher_te
     repo.git.pull("--set-upstream", repo.remote().name, str(repo.head.ref))
     api_code_file_name = generate_notebook_filename(job_id=job_id)
 
-    assert c.text == f"{renku_project_url}/sessions/new?autostart=1&branch=mmoda_request_{job_id}&commit={repo.head.commit.hexsha}&notebook={api_code_file_name}"
+    assert c.text == f"{renku_project_url}/sessions/new?autostart=1&branch=mmoda_request_{job_id}" \
+                     f"&commit={repo.head.commit.hexsha}" \
+                     f"&notebook={api_code_file_name}" \
+                     f"&env[ODA_TOKEN]={encoded_token}"
```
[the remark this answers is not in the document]
burnout87: I am testing again now, but it seems renkulab has some connection issue ATM, and doesn't  build the image.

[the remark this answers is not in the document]
burnout87: I can confirm it works! I just to upgrade the renku version of the test repository on renkulab, but the env_var is successfully passed.

image

cdci_data_analysis/analysis/renku_helper.py
volodymyrss: could you actively distinguish autosaves, since they have distinctly different format? To avoid missing actual problem creating multiple branches - this should not happen.

Maybe only match branches with a more strict pattern?

In principle, users can also create similar branches.

volodymyrss (approved):

cdci_data_analysis/analysis/renku_helper.py
burnout87: will address via #478

PR #474, opened 2022-11-09
allow encoding of ODAAstropyTable

dsavchenko:


dsavchenko:
It's so simple that I can't imagine any

burnout87 (approved):

volodymyrss:
what is going to be done with this encoded output? how to take care to decode it adequately?

volodymyrss:
if decoding/interpretting is not guaranteed, it could be safer not to encode it to highlight possible issue.

volodymyrss:
> It's so simple that I can't imagine any  

encoding of the object should be tested somewhere, but rather in oda_api

but for dispatecher, is there a case of a request when it is actually added in the result, and encoded?

dsavchenko:
it's needed for this PR https://github.com/oda-hub/dispatcher-plugin-nb2workflow/pull/32 where I add ODAAstropyTable as data product

volodymyrss:
> it's needed for this PR [oda-hub/dispatcher-plugin-nb2workflow#32](https://github.com/oda-hub/dispatcher-plugin-nb2workflow/pull/32) where I add ODAAstropyTable as data product

so can it be tested there?

dsavchenko:
> so can it be tested there?
The full cycle can be tested in the plugin, of course.

But as the data product itself is in oda_api, it's better to test the encoding and decoding there. I'm not sure if there are such tests, need to check

volodymyrss:
> > so can it be tested there?
> 
> The full cycle can be tested in the plugin, of course.
> 
> But as the data product itself is in oda_api, it's better to test the encoding and decoding there. I'm not sure if there are such tests, need to check

It's good to test them there.

But it's also good to make sure that encoding and decoding of this object are actually correctly used in the dispatcher.
This can be done in the dispatcher or in a plugin which uses these codings and decodings.

dsavchenko:
The testing of encoding and decoding in oda_api is a bit strange and convoluted. 
This https://github.com/oda-hub/oda_api/blob/6b5d953b0f86be4e7238c295cbaa6f7243a1eb90/tests/test_data_products.py#L32 uses CustomJSONEncoder from dispatcher code which in turn just references encode-decode implementation of oda_api product classes.

dsavchenko:
We agreed on https://github.com/oda-hub/dispatcher-app/issues/475 and https://github.com/oda-hub/oda_api/issues/167
to improve code and tests

volodymyrss (approved):

Issue #478, opened 2022-12-05
Distinguish Renku autosaves branches


burnout87:
When verifying the presence of of a branch for a specific `job_id`.

In relation to the comment https://github.com/oda-hub/dispatcher-app/pull/473#discussion_r1039207579

PR #480, opened 2022-12-06
including os library and stop force pushing

burnout87:


volodymyrss (approved):

burnout87:
> Will address also #481

It checks if the branch already exists in the remote, if so, makes a git pull so to avoid the error when pushing to the branch, and also avoid to force push

volodymyrss:
> > Will address also #481
> 
> It checks if the branch already exists in the remote, if so, makes a git pull so to avoid the error when pushing to the branch, and also avoid to force push

Usually, there should be no need to commit anything, right? 
The commit is only needed when new version of the code is generated.
In such a case, the branch will also contain previous version of the code. It's not really useful for the user and may cause confusion.
We want this live session to be potentially directly used for developing added-value workflow.

So I'd suggest to add in the branch name the version of the code. Actually, you can just use hash of the notebook content.

We can also keep to pull-and-commit behavior you add now as a fallback if something unexpected happens. We should also send a sentry alert in this case.

burnout87:
addressing also #478

burnout87:
> 
> Usually, there should be no need to commit anything, right? The commit is only needed when new version of the code is generated. In such a case, the branch will also contain previous version of the code. It's not really useful for the user and may cause confusion. We want this live session to be potentially directly used for developing added-value workflow.
> 

Yes this is true. Indeed now, it pulls in case the branch already exists, and get the commit hex, that uses to generate the url.
If the branch does not exists, it makes the push.


> So I'd suggest to add in the branch name the version of the code. Actually, you can just use hash of the notebook content.
> 
I generate a has using the content of the notebook. I had to remove two fieds, two ids, that were generate at the moment the notebook object was created.

> We can also keep to pull-and-commit behavior you add now as a fallback if something unexpected happens. We should also send a sentry alert in this case.

As of now. it sends a sentry in case something goes wrong, let's if it is enough

cdci_data_analysis/analysis/renku_helper.py
```diff
@@ -175,11 +200,22 @@ def generate_notebook_filename(job_id):
...
 
 
-def create_new_notebook_with_code(repo, api_code, file_name):
+def write_notebook_file(repo, nb, file_name):
     repo_dir = repo.working_dir
-
     file_path = os.path.join(repo_dir, file_name)
+    nbf.write(nb, file_path)
 
+    return file_path
```
volodymyrss, line 208: some new lines here

volodymyrss (changes requested):
just some formatting point

volodymyrss (approved):

PR #487, opened 2022-12-12
including sentry for renku integration

burnout87:


volodymyrss:
would be good also to avoid showing raw internal error html, something meanigful and maybe a reference to https://renkulab.statuspage.io/ and https://mmoda.statuspage.io/ ?

burnout87:
> would be good also to avoid showing raw internal error html, something meanigful and maybe a reference to https://renkulab.statuspage.io/ and https://mmoda.statuspage.io/ ?

That is a good idea, and it will require some change on the frontend side. Now I need this to get more information since, I run it locally, and I couldn't reproduce it.

volodymyrss:
> > would be good also to avoid showing raw internal error html, something meanigful and maybe a reference to https://renkulab.statuspage.io/ and https://mmoda.statuspage.io/ ?
> 
> That is a good idea, and it will require some change on the frontend side. Now I need this to get more information since, I run it locally, and I couldn't reproduce it.

You output a line returned from the dispatcher, it's cleared from html?

I see this problem in particular:

```
Traceback (most recent call last):
  File "/pyenv/versions/3.8.5/lib/python3.8/site-packages/cdci_data_analysis/analysis/renku_helper.py", line 52, in push_api_code
    notebook_hash = generate_nb_hash(nb_obj)
  File "/pyenv/versions/3.8.5/lib/python3.8/site-packages/cdci_data_analysis/analysis/renku_helper.py", line 214, in generate_nb_hash
    copied_nb['cells'][0].pop('id')
KeyError: 'id'
```

burnout87:
> ```
> Traceback (most recent call last):
>   File "/pyenv/versions/3.8.5/lib/python3.8/site-packages/cdci_data_analysis/analysis/renku_helper.py", line 52, in push_api_code
>     notebook_hash = generate_nb_hash(nb_obj)
>   File "/pyenv/versions/3.8.5/lib/python3.8/site-packages/cdci_data_analysis/analysis/renku_helper.py", line 214, in generate_nb_hash
>     copied_nb['cells'][0].pop('id')
> KeyError: 'id'
> ```

where could you see this?

volodymyrss:
> > ```
> > Traceback (most recent call last):
> >   File "/pyenv/versions/3.8.5/lib/python3.8/site-packages/cdci_data_analysis/analysis/renku_helper.py", line 52, in push_api_code
> >     notebook_hash = generate_nb_hash(nb_obj)
> >   File "/pyenv/versions/3.8.5/lib/python3.8/site-packages/cdci_data_analysis/analysis/renku_helper.py", line 214, in generate_nb_hash
> >     copied_nb['cells'][0].pop('id')
> > KeyError: 'id'
> > ```
> 
> where could you see this?

in the dispatcher exception.

burnout87:
> > > would be good also to avoid showing raw internal error html, something meanigful and maybe a reference to https://renkulab.statuspage.io/ and https://mmoda.statuspage.io/ ?
> > 
> > 
> > That is a good idea, and it will require some change on the frontend side. Now I need this to get more information since, I run it locally, and I couldn't reproduce it.
> 
> You output a line returned from the dispatcher, it's cleared from html?
> 

I parse the object returned from the dispatcher, if this contains errors, this is contained in the error_message

volodymyrss:
> > > > would be good also to avoid showing raw internal error html, something meanigful and maybe a reference to https://renkulab.statuspage.io/ and https://mmoda.statuspage.io/ ?
> > > 
> > > 
> > > That is a good idea, and it will require some change on the frontend side. Now I need this to get more information since, I run it locally, and I couldn't reproduce it.
> > 
> > 
> > You output a line returned from the dispatcher, it's cleared from html?
> 
> I parse the object returned from the dispatcher, if this contains errors, this is contained in the error_message

and if `error_message` has html in it, will it be shown?

burnout87:
> 
> ```
> Traceback (most recent call last):
>   File "/pyenv/versions/3.8.5/lib/python3.8/site-packages/cdci_data_analysis/analysis/renku_helper.py", line 52, in push_api_code
>     notebook_hash = generate_nb_hash(nb_obj)
>   File "/pyenv/versions/3.8.5/lib/python3.8/site-packages/cdci_data_analysis/analysis/renku_helper.py", line 214, in generate_nb_hash
>     copied_nb['cells'][0].pop('id')
> KeyError: 'id'
> ```

BTW it would be really helpful for me for the debug to see the full stack trace and the dispatcher output,

volodymyrss:
> > and if `error_message` has html in it, will it be shown?
> 
> in case of error, this does not happen
> 
> in what happened today, the html was returned from the dispatcher, as a result of an handled expecption
> 
> the frontend could not parse what the dispatcher returned and so the whole HTML was displayed

maybe it should happen in case of a controller error, if the output contains json, the html can be rendered.

Really, the 500 error formatted by flask with no details should never be returned by the endpoint. Instead, a useful message should be shown with details about where to check the status.

What do you think?

burnout87:
> maybe it should happen in case of a controller error, if the output contains json, the html can be rendered.
> 

if the output is json, there is no html, the html that appeared in the error was the result of an exception, and for some reason the dispatcher did not return an object

> Really, the 500 error formatted by flask with no details should never be returned by the endpoint. Instead, a useful message should be shown with details about where to check the status.
> 
> What do you think?

I agree that it should never, that is why I would ideally need to investigate the dispatcher log, in order to properly udnerstand what went wrong and an html was returned

volodymyrss:
> > maybe it should happen in case of a controller error, if the output contains json, the html can be rendered.
> 
> if the output is json, there is no html, the html that appeared in the error was the result of an exception, and for some reason the dispatcher did not return an object
> 
> > Really, the 500 error formatted by flask with no details should never be returned by the endpoint. Instead, a useful message should be shown with details about where to check the status.
> > What do you think?
> 
> I agree that it should never, that is why I would ideally need to investigate the dispatcher log, in order to properly udnerstand what went wrong and an html was returned

We'll organize access to sentry.

But could nevertheless make a try-except block which catches any relevant exceptions and returns something useful?

burnout87:
> 
> But could nevertheless make a try-except block which catches any relevant exceptions and returns something useful?

It is already there : 
 https://github.com/oda-hub/dispatcher-app/blob/ce6f8ec994ba3da7a15106f47da66829fac1423e/cdci_data_analysis/analysis/renku_helper.py#L21-L104

volodymyrss:
> > But could nevertheless make a try-except block which catches any relevant exceptions and returns something useful?
> 
> It is already there, and it looks like the error is happening inside this try-catch, but apparently there is something wrong:
> 
> https://github.com/oda-hub/dispatcher-app/blob/ce6f8ec994ba3da7a15106f47da66829fac1423e/cdci_data_analysis/analysis/renku_helper.py#L21-L104

But the first error did not happen in there? Or why was it shown as internal error? Any insight?

burnout87:
> 
> But the first error did not happen in there? Or why was it shown as internal error? Any insight?

I couldn't see what the first error contained, and I didnt manage to reproduce it on production nor locally

volodymyrss:
> I couldn't see what the first error contained, and I didnt manage to reproduce it on production nor locally

It was probably because renku was down. Did you try a case when a fake unaccessible host is used instead of renkulab?

> the frontend could not parse what the dispatcher returned and so the whole HTML was displayed

I thought you referred to it here? Or I misunderstood?

burnout87:
> 
> It was probably because renku was down. Did you try a case when a fake unaccessible host is used instead of renkulab?
> 

then probably the content was correctly posted, but could not open the link since renku was down...

> 
> I thought you referred to it here? Or I misunderstood?

that is my interpretation of what happened and why the html was displayed

volodymyrss:
> > It was probably because renku was down. Did you try a case when a fake unaccessible host is used instead of renkulab?
> 
> then probably the content was correctly posted, but could not open the link since renku was down...
> 
> > I thought you referred to it here? Or I misunderstood?
> 
> that is my interpretation of what happened and why the html was displayed

The message with raw html was inside the box where the outcome of posting is written. Did the message come from the dispatcher?

volodymyrss:
> ```
> Traceback (most recent call last):
>   File "/pyenv/versions/3.8.5/lib/python3.8/site-packages/cdci_data_analysis/analysis/renku_helper.py", line 52, in push_api_code
>     notebook_hash = generate_nb_hash(nb_obj)
>   File "/pyenv/versions/3.8.5/lib/python3.8/site-packages/cdci_data_analysis/analysis/renku_helper.py", line 214, in generate_nb_hash
>     copied_nb['cells'][0].pop('id')
> KeyError: 'id'
> ```

This exception indicates that the notebook was incomplete. Could you make a catch in this function and throw some exception with details of what is wrong with the notebook?

burnout87:
> 
> The message with raw html was inside the box where the outcome of posting is written. Did the message come from the dispatcher?

Probably yes, but I am doing some testing trying to reproduce the error, of course, all this is in the frontend so some changes will be necessary there. In any case, it's tricky to guess what caused it at the moment.

>
> This exception indicates that the notebook was incomplete. Could you make a catch in this function and throw some exception with details of what is wrong with the notebook?

yes that is something I was planning on doing

burnout87:
> 
> This exception indicates that the notebook was incomplete. Could you make a catch in this function and throw some exception with details of what is wrong with the notebook?

It now catches and throws exception when generating the hash

volodymyrss (approved):

PR #488, opened 2022-12-12
binary parameter

dsavchenko:


burnout87:
So...nothing happens?

dsavchenko:
> So...nothing happens?

I already copied this code to another branch, so no need in this PR anymore

PR #489, opened 2022-12-14
freezing nbformat version

burnout87:


requirements.txt
```diff
@@ -26,7 +26,7 @@ aiosmtpd
 marshmallow
 requests
 GitPython
-nbformat
+nbformat==5.7.0
```
volodymyrss, line 29: it might end up not working in some environments. It is sufficient for starting renku session.
Is it possible to support both versions?

burnout87: ok I will revert this change

for the `id`, it will simply try to pop it, trying to avoid the keyerror that would otherwise happen

cdci_data_analysis/analysis/renku_helper.py
```diff
@@ -213,7 +213,7 @@ def generate_nb_hash(nb_obj):
 
     try:
         copied_notebook_obj['cells'][0].pop('id', None)
-        copied_notebook_obj['cells'].pop('id', None)
+        copied_notebook_obj['cells'][1].pop('id', None)
```
volodymyrss, line 216: you'd need to remove it from all cells, right? Right now it's two, but why not remove from all, in a loop?

cdci_data_analysis/analysis/renku_helper.py
```diff
@@ -212,8 +212,8 @@ def generate_nb_hash(nb_obj):
     copied_notebook_obj = copy.deepcopy(nb_obj)
 
     try:
-        del copied_notebook_obj['cells'][0]['id']
-        del copied_notebook_obj['cells'][1]['id']
+        copied_notebook_obj['cells'][0].pop('id', None)
+        copied_notebook_obj['cells'].pop('id', None)
```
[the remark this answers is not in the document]
burnout87: > not "[1]" ? just a guess

yes, noticed and corrected 
> 
> also, does it make sense to just loop over all of them? We might add more cells in the near future.

ok, didnt consider that eventuality but yeah makes sense

cdci_data_analysis/analysis/renku_helper.py
burnout87: now loops over all the cells and pop the id out if present

volodymyrss:
it's deployed in prod with the latest version https://gitlab.astro.unige.ch/oda/dispatcher/dispatcher-chart/-/pipelines and the fixes the title issue.
but there is  an issue: https://github.com/oda-hub/dispatcher-app/issues/491

PR #493, opened 2022-12-14
Parameter restrictions in metadata

dsavchenko:
Added min_value and max_value to the parameters

min_value, max_value and allowed_values are shown in metadata

also added boolean parameter type

### DIFF
--- a/cdci_data_analysis/analysis/parameters.py
+++ b/cdci_data_analysis/analysis/parameters.py
@@ -431,7 +431,7 @@ class Parameter:
 
     @classmethod
     def matches_owl_uri(cls, owl_uri: str) -> bool:
-        return owl_uri in getattr(cls, "owl_uris", [])
+        return owl_uri in getattr(cls, "owl_uris", ())
 
     @classmethod
     def from_owl_uri(cls,
@@ -478,7 +478,7 @@ class Parameter:
         return possible_parameter_interpretations[0]
 
 class String(Parameter):
-    owl_uris = ["http://www.w3.org/2001/XMLSchema#str"]
+    owl_uris = ("http://www.w3.org/2001/XMLSchema#str")
     
     def __init__(self, value=None, name_format='str', name=None, allowed_values = None):
 
@@ -495,7 +495,7 @@ class String(Parameter):
         pass
 
 class Name(String):
-    owl_uris = ["http://odahub.io/ontology#AstrophysicalObject"]
+    owl_uris = ("http://odahub.io/ontology#AstrophysicalObject")
 
 class NumericParameter(Parameter):
     @staticmethod
@@ -508,7 +508,7 @@ class NumericParameter(Parameter):
                 raise ValueError(f'Parameter {name} wrong value {val}: should be less than {max_value}')
 
 class Float(NumericParameter):
-    owl_uris = ["http://www.w3.org/2001/XMLSchema#float"]
+    owl_uris = ("http://www.w3.org/2001/XMLSchema#float")
     def __init__(self, 
                  value=None, 
                  units=None, 
@@ -576,7 +576,7 @@ class Float(NumericParameter):
 
 
 class Integer(NumericParameter):
-    owl_uris = "http://www.w3.org/2001/XMLSchema#int"
+    owl_uris = ("http://www.w3.org/2001/XMLSchema#int")
 
     def __init__(self, value=None, units=None, name=None, check_value=None, min_value = None, max_value = None):
 
@@ -641,9 +641,9 @@ class Time(Parameter):
     # of this class.
     # reading the rdf should be done with thread-safe caching to avoid frequent requests
 
-    owl_uris = ["http://odahub.io/ontology#TimeInstant",
+    owl_uris = ("http://odahub.io/ontology#TimeInstant",
                 "http://odahub.io/ontology#StartTime",
-                "http://odahub.io/ontology#EndTime"]
+                "http://odahub.io/ontology#EndTime")
 
     def __init__(self, value=None, T_format='isot', name=None, Time_format_name=None, par_default_format='isot'):
 
@@ -690,7 +690,7 @@ class Time(Parameter):
 # it is confusing that TimeDelta derives from Time.  
 # https://github.com/astropy/astropy/blob/main/astropy/time/core.py#L379
 class TimeDelta(Time):
-    owl_uris = [] # to avoid unnecessary attempt to initialize it for time parameter
+    owl_uris = () # to avoid unnecessary attempt to initialize it for time parameter
     
     def __init__(self, value=None, delta_T_format='sec', name=None, delta_T_format_name=None, par_default_format='sec'):
 
@@ -781,7 +781,7 @@ class InputProdList(Parameter):
 
 
 class Angle(Float):
-    owl_uris = ["http://odahub.io/ontology#PointOfInterestRA", "http://odahub.io/ontology#PointOfInterestDEC"]
+    owl_uris = ("http://odahub.io/ontology#PointOfInterestRA", "http://odahub.io/ontology#PointOfInterestDEC")
     
     def __init__(self, value=None, units=None, default_units='deg', name=None):
 
@@ -867,7 +867,7 @@ class UserCatalog(Parameter):
         pass
 
 class Boolean(Parameter):
-    owl_uris = ['http://www.w3.org/2001/XMLSchema#bool']
+    owl_uris = ('http://www.w3.org/2001/XMLSchema#bool')
     
     def __init__(self, value=None, name=None):
 
--- a/cdci_data_analysis/analysis/parameters.py
+++ b/cdci_data_analysis/analysis/parameters.py
@@ -223,10 +223,7 @@ class Parameter:
                 par_default_format is not None and par_default_format != '':
             par_format = par_default_format
 
-        if check_value is not None:
-            logger.warning('The check_value argument of the %s parameter class is DEPRECATED', 
-                            self.__class__.__name__)
-            self.check_value = check_value
+        self.check_value = check_value
 
         if allowed_units is not None:
             # handles case of []
@@ -488,14 +485,11 @@ class String(Parameter):
         _allowed_units = ['str']
         super().__init__(value=value,
                          units=name_format,
+                         check_value=self.check_name_value,
                          name=name,
                          allowed_units=_allowed_units,
                          allowed_values=allowed_values)
-        
-    @staticmethod
-    def check_value(value, units=None, name=None, par_format=None):
-        String.check_name_value(value, units=units, name=name, par_format=par_format)
-        
+
     @staticmethod
     def check_name_value(value, units=None, name=None, par_format=None):
         pass
@@ -525,14 +519,12 @@ class Float(NumericParameter):
                  min_value= None,
                  max_value = None):
 
-        if check_value is not None:
-            self.check_value = check_value
-            logger.warning('The check_value argument of the %s parameter class is DEPRECATED', 
-                           self.__class__.__name__)
-            
+        if check_value is None:
+            check_value = self.check_float_value
         
         super().__init__(value=value,
                          units=units,
+                         check_value=check_value,
                          default_units=default_units,
                          name=name,
                          default_type=float,
@@ -581,10 +573,7 @@ class Float(NumericParameter):
             except:
                 raise RuntimeError(f'the Float parameter {name} cannot be assigned the value {value} '
                                    f'of type {type(value).__name__}')
-    
-    @staticmethod
-    def check_value(value, units=None, name=None):
-        Float.check_float_value(value, units=units, name=name)
+
 
 class Integer(NumericParameter):
     owl_uris = "http://www.w3.org/2001/XMLSchema#int"
@@ -593,13 +582,12 @@ class Integer(NumericParameter):
 
         _allowed_units = None
 
-        if check_value is not None:
-            self.check_value = check_value
-            logger.warning('The check_value argument of the %s parameter class is DEPRECATED', 
-                           self.__class__.__name__)
+        if check_value is None:
+            check_value = self.check_int_value
         
         super().__init__(value=value,
                          units=units,
+                         check_value=check_value,
                          default_type=int,
                          allowed_types=[int],
                          name=name,
@@ -645,9 +633,6 @@ class Integer(NumericParameter):
                 raise RuntimeError(f'the Integer parameter {name} cannot be assigned the value {value} '
                                    f'of type {type(value).__name__}')
 
-    @staticmethod
-    def check_value(val, units=None, name=None):
-        Integer.check_int_value(val, units, name)
 
 class Time(Parameter):
     # TODO:
@@ -742,6 +727,7 @@ class InputProdList(Parameter):
 
         super().__init__(value=value,
                          par_format=_format,
+                         check_value=self.check_list_value,
                          name=name,
                          allowed_units=_allowed_units)
 
@@ -793,9 +779,6 @@ class InputProdList(Parameter):
         else:
             raise RuntimeError(f'{name} units not valid {par_format}')
 
-    @staticmethod
-    def check_value(val, units=None, name=None, par_format=None):
-        InputProdList.check_list_value(val, units, name, par_format)
 
 class Angle(Float):
     owl_uris = ["http://odahub.io/ontology#PointOfInterestRA", "http://odahub.io/ontology#PointOfInterestDEC"]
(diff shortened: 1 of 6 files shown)


cdci_data_analysis/analysis/parameters.py
```diff
@@ -806,3 +865,28 @@ def __init__(self, value=None, name_format='str', name=None):
...
+        self._true_rep = ['True', 'true', 'yes', '1', True]
+        self._false_rep = ['False', 'false', 'no', '0', False]
+        super().__init__(value=value,
+                         name=name,
+                         allowed_values=self._true_rep+self._false_rep
+                         )
+
+    @property
+    def value(self):
+        return str(self._value).lower() #because passed in json
```
burnout87, line 883: always in json?

EDIT: you mean that it always will be used when building a json object?

dsavchenko: Good question. This came from gw plugin (I added boolean parameter locally there).
There, it's used this way. And it conforms to the way other parameters are passed.

On the other hand, the general logic is to return real boolean value. So I'm not sure on this.

Probably, it's better be postprocessed in custom encoder. I'll try

cdci_data_analysis/analysis/parameters.py
```diff
@@ -61,6 +61,14 @@ def subclasses_recursive(cls):
         indirect.extend(subclasses_recursive(subclass))
     return direct + indirect
 
+def basic_check_bounds(val, min_value = None, max_value = None, name=None):
```
burnout87, line 64: perhaps I would rename this as `basic_numeric_check_bounds`, as it the implementation is the most suitable for numeric values

dsavchenko: There is no implementation for non-numerics now. But ok, it could be added for e.g. time

burnout87: Indeed it's just a naming adjustment

burnout87 (commented):
a question

dsavchenko:
> a question

?

burnout87:
> > a question
> 
> ?

sorry, was a mistake

cdci_data_analysis/analysis/parameters.py
```diff
@@ -398,10 +416,23 @@ def check_type(par_type, allowed, name):
...
-
+    
+    @staticmethod
+    def check_bounds(val, min_value, max_value, name):
+        raise NotImplementedError(f"Parameter {name} doesn't support min_value/max_value check")
+        
     def reprJSONifiable(self):
         # produces json-serialisable list
         reprjson = [dict(name=self.name, units=self.units, value=self.value)]
+        restrictions = {}
```
volodymyrss, line 427: note that `check_value` is also a restriction. It can even be used by the plugin developer to implement this functionality. Could you please check what `check_value` does in plugins, and what should be the remaining function of `check_value`?

cdci_data_analysis/analysis/parameters.py
```diff
@@ -475,10 +507,20 @@ class Name(String):
...
+                 allowed_units=None, 
+                 default_units=None, 
+                 check_value=None, 
+                 min_value= None,
+                 max_value = None):
 
         if check_value is None:
             check_value = self.check_float_value
+        
+        self.check_bounds = basic_numeric_check_bounds
```
volodymyrss, line 523: This somehow mimics what would be more naturally done by an additional intermediate "BoundedParameter" (NumericParameter?)

cdci_data_analysis/analysis/parameters.py
dsavchenko: I don't see any uses of `check_value` in the plugins. 
In the derived Parameter classes, it's generally used to check for the proper python data type of the value.

cdci_data_analysis/analysis/parameters.py
dsavchenko: I don't like "BoundedParameter" as bounds are optional. NumericParameter is ok, it could be implemented like this, if you think it's more natural. 

About general comparison operation, we should be careful with this. Python strings, e.g., have comparison defined, but it's meaningless in most cases

volodymyrss: more like "boundable" then. But Numeric is a useful specialization of it which we can use.

cdci_data_analysis/analysis/parameters.py
dsavchenko: To be clear, there are no direct uses of it in plugins. 
There is one indirect use https://github.com/oda-hub/dispatcher-plugin-integral/blob/0f8d9754aa7c3da5a6aba14d8fb4c8acab9f60f7/cdci_osa_plugin/osa_isgri.py#L55
https://github.com/oda-hub/dispatcher-plugin-integral/blob/0f8d9754aa7c3da5a6aba14d8fb4c8acab9f60f7/cdci_osa_plugin/osa_isgri.py#L55
to check for the bounds.

So I propose to remove check_value from the specific parameters' signature. There are allowed_values and min/max for the most useful checks. If a plugin developer really needs some complex check, it could be accomplished with inheritance. 

Anyway, it's unclear how to report complex checks in parameter metadata.

volodymyrss: In principle, it makes sense. But to be safe, I'd make a deprecation warning when attempting to use it, instead of removing.

volodymyrss:
Not sure, it's the part @burnout87 made. I can look tomorrow maybe too.

```
2022-12-19T15:52:12.0823568Z FAILED tests/test_job_management.py::test_email_catalog[file] - RuntimeError: file format for catalog not valid
2022-12-19T15:52:12.0823818Z FAILED tests/test_job_management.py::test_email_catalog[params] - AssertionError
2022-12-19T15:52:12.0824125Z FAILED tests/test_server_basic.py::test_catalog_file[True] - RuntimeError: file format for catalog not valid
2022-12-19T15:52:12.0824369Z FAILED tests/test_server_basic.py::test_user_catalog[10-True] - AssertionError
2022-12-19T15:52:12.0824606Z FAILED tests/test_server_basic.py::test_user_catalog[11-True] - AssertionError
2022-12-19T15:52:12.0824853Z FAILED tests/test_server_basic.py::test_user_catalog[1,1,1-True] - AssertionError
2022-12-19T15:52:12.0825088Z FAILED tests/test_server_basic.py::test_user_catalog[None-True] - AssertionError
2022-12-19T15:52:12.0825628Z FAILED tests/test_server_basic.py::test_user_catalog_oda_api[True] - oda_api.api.RequestNotUnderstood: [ RequestNotUnderstood: RequestNotUnderstood():Error while setting catalog object , content of the temporary directory is [] ]
```

burnout87:
> Any idea on why some tests (catalog-related) have started to fail? They pass on my laptop, so I can't figure out

I am also noticing these errors, I am taking a look

burnout87:
> > Any idea on why some tests (catalog-related) have started to fail? They pass on my laptop, so I can't figure out
> 
> I am also noticing these errors, I am taking a look

edit same tests are failing also for me in #496

volodymyrss:
> > > Any idea on why some tests (catalog-related) have started to fail? They pass on my laptop, so I can't figure out
> > 
> > 
> > I am also noticing these errors, I am taking a look
> 
> edit same tests are failing also for me in #496

In particular this this to suggest there test content for the catalog is not parsed. Maybe astropy Table reading changed?..

```
FAILED tests/test_server_basic.py::test_catalog_file[True] - RuntimeError: file format for catalog not valid
```

burnout87:
> 
> In particular this this to suggest there test content for the catalog is not parsed. Maybe astropy Table reading changed?..
> 
> ```
> FAILED tests/test_server_basic.py::test_catalog_file[True] - RuntimeError: file format for catalog not valid
> ```

I am looking at that. Was there some recent change of version of some library we use?

volodymyrss:
> > In particular this this to suggest there test content for the catalog is not parsed. Maybe astropy Table reading changed?..
> > ```
> > FAILED tests/test_server_basic.py::test_catalog_file[True] - RuntimeError: file format for catalog not valid
> > ```
> 
> I am looking at that. Was there some recent change of version of some library we use?

it changes often enough. Last time it changed a week ago.  https://pypi.org/project/astropy/
Maybe you can set previous version `astropy==5.1`. But ultimately it's much better to adapt our test to new astropy (best if backward-compatible).

burnout87:
I think I managed to replicate the error locally, and this is the error I also found:

`module 'numpy' has no attribute 'bool'`

so I uninstalled and re-installed numpy locally and the same error appeared, and it appears when parsing the catalog.

Any ideas?

burnout87:
> I think I managed to replicate the error locally, and this is the error:
> 
> `module 'numpy' has no attribute 'bool'`
> 
> PS by uninstalling and re-installing numpy locally the same error appeared, and it appears when parsing the catalog.
> 
> Any ideas?

this is the point:

https://github.com/oda-hub/dispatcher-app/blob/17d2855747beccaa0f7aa9b3a3b687f198daa4aa/cdci_data_analysis/analysis/catalog.py#L55

By changing: `np.bool` to `np.bool_`, it worked

dsavchenko:
Great, thank you! 
It was actually deprecated in numpy 1.20 https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated
then probably removed in one of the following numpy versions (haven't found in which actually)
Will you fix it by separate PR?

burnout87:
> Will you fix it by separate PR?

#498

cdci_data_analysis/analysis/parameters.py
dsavchenko: Resolved by https://github.com/oda-hub/dispatcher-app/pull/493/commits/82f3cb5ed5e1ad34886970330efd84fb3080011f

volodymyrss (commented):
is there also a test somewhere on how this propagates to the client?

dsavchenko:
> is there also a test somewhere on how this propagates to the client?

I thought `test_parameter_meta_data` is enough, as these are then just concatenated into full meta-data output.

dsavchenko:
I just noticed one point on this check_value argument. Requires some changes

dsavchenko:
I give up! Let's not deprecate this check_value argument now.

volodymyrss (dismissed):

volodymyrss:
I suppose it is tested throughout. Could be useful to expand these tests for more options.

dsavchenko:
> I suppose it is tested throughout. Could be useful to expand these tests for more options.

What you mean by "more options"?

volodymyrss:
> > I suppose it is tested throughout. Could be useful to expand these tests for more options.
> 
> What you mean by "more options"?

More variations, and with verification that the bounds are applied when interacting with dispatcher fixture.

tests/test_parameters.py
```diff
@@ -342,3 +344,61 @@ def test_parameter_from_owl_uri_extra_param(caplog):
...
+    fl_param = Float(5., name = 'FL', min_value = 2.2, max_value = 7.7)
+    with pytest.raises(ValueError):
+        int_param.value = 1
+        int_param.value = 10
+        fl_param.value = 1.2
+        fl_param.value = 8.3
+        Integer(1, name = 'INT', min_value = 2, max_value = 8)
+        Float(1.1, name = 'FL', min_value = 2.2, max_value = 7.7)
+        Integer(10, name = 'INT', min_value = 2, max_value = 8)
+        Float(8.2, name = 'FL', min_value = 2.2, max_value = 7.7)
```
burnout87, line 360: are these ever reached in the test?

burnout87 (changes requested):

tests/test_parameters.py
dsavchenko: Thanks!

cdci_data_analysis/analysis/parameters.py
```diff
@@ -633,6 +690,8 @@ def _set_time(self, value, par_format):
 # it is confusing that TimeDelta derives from Time.  
 # https://github.com/astropy/astropy/blob/main/astropy/time/core.py#L379
 class TimeDelta(Time):
+    owl_uris = [] # to avoid unnecessary attempt to initialize it for time parameter
```
volodymyrss, line 693: could it be `()` tuple to avoid possible issues with mutations?

tests/test_parameters.py
```diff
@@ -342,3 +344,68 @@ def test_parameter_from_owl_uri_extra_param(caplog):
                            units='d', # wrong parameter
                            name='example')
     assert "parameter units with value d not used to construct <class 'cdci_data_analysis.analysis.parameters.Time'>" in caplog.text
+
+@pytest.mark.fast
+def test_parameter_bounds():
```
volodymyrss, line 349: this is good, but maybe it'd really help to ask the live dispatcher fixture, to check what sort of response we are getting out all the way.

A test for oda_api could be helpful too.

dsavchenko:
Integral and oda_api tests fail due to https://github.com/oda-hub/dispatcher-plugin-integral/pull/114 and https://github.com/oda-hub/oda_api/pull/176 are merged but not #504 
Unrelated to this PR

burnout87 (dismissed):

tests/test_server_basic.py
```diff
@@ -2849,4 +2849,25 @@ def test_catalog_selected_objects_accepted(dispatcher_live_fixture):
...
+    jdata=c.json()
+    
+    metadata = [json.loads(x) for x in jdata[0] if isinstance(x, str)]
+    restricted_meta = [x for x in metadata if x[0]['query_name'] == 'restricted_parameters_dummy_query'][0]
+    def meta_for_par(parname):
+        return [x for x in restricted_meta if x.get('name', None) == parname][0]
+    
+    assert meta_for_par('bounded_int_par')['restrictions'] == {'min_value': 2, 'max_value': 8}
+    assert meta_for_par('bounded_float_par')['restrictions'] == {'min_value': 2.2, 'max_value': 7.7}
+    assert meta_for_par('string_select_par')['restrictions'] == {'allowed_values': ['spam', 'eggs', 'ham']}
```
volodymyrss, line 2873: ok! could you just add one more which requests it with in-bounds and out-of-bounds values? to see that the response makes sense. To keep in mind how will it be shown in api, dispatcher?

volodymyrss: yes

cdci_data_analysis/analysis/parameters.py
```diff
@@ -701,7 +762,7 @@ def value(self, v):
                 self.check_value(v, par_format=self.par_format, name=self.name)
             if self._allowed_values is not None:
                 if v not in self._allowed_values:
-                    raise RuntimeError(f'value {v} not allowed, allowed= {self._allowed_values}')
+                    raise RuntimeError(f'Parameter {self.name} wrong value {v}: not in allowed {self._allowed_values}')
```
volodymyrss, line 765: just wondering, why is it `RuntimeError` and not something like [`RequestNotUnderstood`](https://github.com/oda-hub/dispatcher-app/blob/master/cdci_data_analysis/analysis/exceptions.py#L37)?

`RuntimeError` can be many things, and we unsurprisingly had issues when internal errors are sent to users unnecessarily. Could be better to distinguish right here that this is specifically an issue with request parameters.

dsavchenko: It was RuntimeError before. And in most similar checks of this module it is.
In check_bounds I've made it to be ValueError though. 
The error message describes that the issue is with parameter value.
So what do you propose?

volodymyrss: I propose to change to more specific exception class.

Also, just for clarify, to check how `RuntimeError` is caught so it converts to 400 error - if this happens there may be false 400 messages.

dsavchenko: And in the integral plugin, the realization of bounds checking actually raises RequestNotUnderstood. I can change all exceptions here to it. But what "it is not user error" means in the exception comment?

volodymyrss: it's just to be polite. some people just react better when we are saying "we can not understand this, you probably meant something else" instead of "you are wrong because you made an error"

dsavchenko: Ok, got it

Concerning exception class, I see now that it definitely must be RequestNotUnderstood. It's a subclass of APIError and it's treated specifically to not clutter the output with unnecessary info. Will change it throughout the module

dsavchenko: I changed the exception class for those occurrences where I am sure the wrong value could come from user input. Not sure about others so left them as-is

volodymyrss (approved):

burnout87 (approved):

volodymyrss (approved):

burnout87 (approved):

dsavchenko:
Merge it please

PR #496, opened 2022-12-15
reformatting code for inspect state and adapted test,  free_up space …

burnout87:
…function with number of scratch folders to delete

### DIFF
--- a/cdci_data_analysis/configurer.py
+++ b/cdci_data_analysis/configurer.py
@@ -226,6 +226,8 @@ class ConfigEnv(object):
                                      disp_dict['dispatcher_callback_url_base'],
                                      disp_dict['secret_key'],
                                      disp_dict.get('token_max_refresh_interval', 604800),
+                                     disp_dict.get('soft_minimum_folder_age_days', 1),
+                                     disp_dict.get('hard_minimum_folder_age_days', 30),
                                      disp_dict['email_options']['smtp_server'],
                                      disp_dict['email_options']['sender_email_address'],
                                      disp_dict['email_options']['cc_receivers_email_addresses'],
@@ -288,6 +290,8 @@ class ConfigEnv(object):
                             dispatcher_callback_url_base,
                             secret_key,
                             token_max_refresh_interval,
+                            soft_minimum_folder_age_days,
+                            hard_minimum_folder_age_days,
                             smtp_server,
                             sender_email_address,
                             cc_receivers_email_addresses,
@@ -322,6 +326,8 @@ class ConfigEnv(object):
         self.dispatcher_callback_url_base = dispatcher_callback_url_base
         self.secret_key = secret_key
         self.token_max_refresh_interval = token_max_refresh_interval
+        self.soft_minimum_folder_age_days = soft_minimum_folder_age_days
+        self.hard_minimum_folder_age_days = hard_minimum_folder_age_days
         self.smtp_server = smtp_server
         self.sender_email_address = sender_email_address
         self.cc_receivers_email_addresses = cc_receivers_email_addresses
--- a/cdci_data_analysis/configurer.py
+++ b/cdci_data_analysis/configurer.py
@@ -226,7 +226,7 @@ class ConfigEnv(object):
                                      disp_dict['dispatcher_callback_url_base'],
                                      disp_dict['secret_key'],
                                      disp_dict.get('token_max_refresh_interval', 604800),
-                                     disp_dict.get('soft_minimum_folder_age_days', 1),
+                                     disp_dict.get('soft_minimum_folder_age_days', 5),
                                      disp_dict.get('hard_minimum_folder_age_days', 30),
                                      disp_dict['email_options']['smtp_server'],
                                      disp_dict['email_options']['sender_email_address'],
--- a/cdci_data_analysis/flask_app/dispatcher_query.py
+++ b/cdci_data_analysis/flask_app/dispatcher_query.py
@@ -348,9 +348,11 @@ class InstrumentQueryBackEnd:
         current_time_secs = time.time()
         hard_minimum_folder_age_days = app_config.hard_minimum_folder_age_days
         # let's pass the minimum age the folders to be deleted should have
-        soft_minimum_folder_age_days = request.args.get('minimum_age_days', None)
+        soft_minimum_folder_age_days = request.args.get('soft_minimum_age_days', None)
         if soft_minimum_folder_age_days is None or isinstance(soft_minimum_folder_age_days, int):
             soft_minimum_folder_age_days = app_config.soft_minimum_folder_age_days
+        else:
+            soft_minimum_folder_age_days = int(soft_minimum_folder_age_days)
 
         list_scratch_dir = sorted(glob.glob("scratch_sid_*_jid_*"), key = os.path.getmtime)
         list_scratch_dir_to_delete = []
--- a/cdci_data_analysis/flask_app/dispatcher_query.py
+++ b/cdci_data_analysis/flask_app/dispatcher_query.py
@@ -375,7 +375,7 @@ class InstrumentQueryBackEnd:
                     job_id = monitor['job_id']
                 if job_status == 'done' and (token is None or token_expired):
                     list_scratch_dir_to_delete.append(scratch_dir)
-                if job_status != 'done' and token_expired:
+                elif job_status != 'done' and token_expired:
                     incomplete_job_alert_message = f"The job {job_id} is yet to complete despite being older " \
                                                    f"than {soft_minimum_folder_age_days} days. This has been detected " \
                                                    f"while checking for deletion the folder {scratch_dir}."
--- a/cdci_data_analysis/flask_app/dispatcher_query.py
+++ b/cdci_data_analysis/flask_app/dispatcher_query.py
@@ -344,7 +344,10 @@ class InstrumentQueryBackEnd:
         for d in list_scratch_dir_to_delete:
             shutil.rmtree(d)
 
-        logger.info(f"Removed {numb_folders_to_delete} scratch directories")
+        result_scratch_dir_deletion = f"Removed {numb_folders_to_delete} scratch directories"
+        logger.info(result_scratch_dir_deletion)
+
+        return jsonify(dict(output_status=result_scratch_dir_deletion))
 
     @staticmethod
     def inspect_state(app):
--- a/cdci_data_analysis/flask_app/dispatcher_query.py
+++ b/cdci_data_analysis/flask_app/dispatcher_query.py
@@ -349,7 +349,7 @@ class InstrumentQueryBackEnd:
         hard_minimum_folder_age_days = app_config.hard_minimum_folder_age_days
         # let's pass the minimum age the folders to be deleted should have
         soft_minimum_folder_age_days = request.args.get('minimum_age_days', None)
-        if soft_minimum_folder_age_days is None:
+        if soft_minimum_folder_age_days is None or isinstance(soft_minimum_folder_age_days, int):
             soft_minimum_folder_age_days = app_config.soft_minimum_folder_age_days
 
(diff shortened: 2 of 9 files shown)


burnout87:
In the current implementation:
- as an argument, the number of the oldest scratch dirs to be deleted is passed, defaults to 5
- those oldest folders are deleted

this is a very first basic implementation, something else to be considered?

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -322,39 +322,48 @@ def __init__(self, app,
...
+        numb_folders_to_delete = int(request.args.get('folder_to_delete', 5))
+
+        list_scratch_dir = sorted(glob.glob("scratch_sid_*_jid_*"), key = os.path.getmtime)
+
+        list_scratch_dir_to_delete = list_scratch_dir[0:numb_folders_to_delete] if len(list_scratch_dir) >= 5 else list_scratch_dir
+
+        for d in list_scratch_dir_to_delete:
+            shutil.rmtree(d)
+
+        result_scratch_dir_deletion = f"Removed {len(list_scratch_dir_to_delete)} scratch directories"
```
volodymyrss, line 347: could you make it json with some info: pre-clean space; after-clean space, number of scratch removed and left, time spent cleaning, etc

burnout87: more information to be logged?

burnout87: >  pre-clean space; after-clean space,

BTW do you have some suggestions regarding retrieving this information? Have you already done it somewhere else?

volodymyrss: os.statvfs?

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -322,39 +320,62 @@ def __init__(self, app,
...
-            # raise RequestNotAuthorized("The token provided is not valid.")
-            return make_response('The token provided is not valid.'), 403
 
-        recent_days = request.args.get('recent_days', 3, type=float)
-        job_id = request.args.get('job_id', None)
+        app_config = app.config.get('conf')
+        secret_key = app_config.secret_key
 
-        roles = tokenHelper.get_token_roles(decoded_token)
+        output, output_code = tokenHelper.validate_token_from_request(token=token, secret_key=secret_key,
```
volodymyrss, line 329: there should be a special role for this, like mmoda-ops

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -322,39 +320,62 @@ def __init__(self, app,
...
-            message = (
-                f"Unfortunately, your privileges are not sufficient for this type of request.\n"
-                f"Your privilege roles include {roles}, but the following roles are missing: {lacking_roles}."
-            )
-            # raise RequestNotAuthorized(message=message)
-            return make_response(message), 403
+        if output_code is not None:
+            return make_response(output, output_code)
+
+        numb_folders_to_delete = int(request.args.get('folder_to_delete', 5))
```
volodymyrss, line 336: this is not a very useful input, since it can only be provided meaningfully if the existing number of folders is known, and their usage is considered. I'd rather suggest "max_age" - to protect very recent folders. Also, we can go in the folders and check if their job is finished.

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: Thanks, I will try it, for the moment I used `shutil.disk_usage`, it seems to be ok

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: I was planning on using the already existing role `job-manager` already used for the inspect state endpoint, why do you think another one, specific for this, is necessary? We are having more and more roles, which many of which are specific for single operations.

cdci_data_analysis/flask_app/dispatcher_query.py
volodymyrss: It does not matter, there are many ways, shutil is a wrapper around lower level functions.

cdci_data_analysis/flask_app/dispatcher_query.py
volodymyrss: The roles first of all are meant to group people, nothing wrong with them being specific to operations if the operations are meaningfully distinct.
Here, removing directories is distinct in that it is invasive. In fact, nobody should be allowed to do this except for the bot.

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: i considered using the number of weeks to refer t the minimum age a certain folder should have in order to be deleted.

If it's "old enough" it will also check the status of its job, and if its done then it will be deleted.

Does it make sense? Is the number of weeks a good enough way to decide on the deletion ?

volodymyrss: in principle this is good, just one more element: if there is a token, we could check if it has expired, and only then delete.

note however that number of weeks is not the right scale since we are rather talking about days here.

burnout87: > if it has expired, and only then delete.
> 

Which advantages does it bring?

> note however that number of weeks is not the right scale since we are rather talking about days here.
>
That is why I was asking

volodymyrss: > > if it has expired, and only then delete. 
> Which advantages does it bring?

We are using these records to reconstruct request parameters by jobid. These records should be available as long as the token provided with them is valid.

volodymyrss: by the way, this way if we detect incomplete jobs older than the soft limit, we can send ourselves an alert. since this is not right

burnout87: good idea, I just introduced those additional checks

BTW testing will be tricky

volodymyrss: > good idea, I just introduced those additional checks
> 
> BTW testing will be tricky

you can set mtime/ctime in the past

https://docs.python.org/3/library/os.html#os.utime

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -322,39 +321,110 @@
...
+                    monitor = json.load(jm_file)
+                    job_status = monitor['status']
+                    job_id = monitor['job_id']
+                if job_status == 'done' and (token is None or token_expired):
+                    list_scratch_dir_to_delete.append(scratch_dir)
+                if job_status != 'done':
+                    incomplete_job_alert_message = f"The job {job_id} is yet to complete despite being older " \
+                                                   f"than {minimum_folder_age_days} days. This has been detected " \
+                                                   f"while checking for deletion the folder {scratch_dir}."
+                    logger.info(incomplete_job_alert_message)
```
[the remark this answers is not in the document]
volodymyrss: maybe good to check it has the right type/form before logging

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -322,39 +321,108 @@ def __init__(self, app,
...
+                    token_expired = True
+
+                job_monitor_path = os.path.join(scratch_dir, 'job_monitor.json')
+                with open(job_monitor_path, 'r') as jm_file:
+                    monitor = json.load(jm_file)
+                    job_status = monitor['status']
+                    job_id = monitor['job_id']
+                if job_status == 'done' and (token is None or token_expired):
+                    list_scratch_dir_to_delete.append(scratch_dir)
+                if job_status != 'done' and token_expired:
```
volodymyrss, line 378: why not just `else:`?

cdci_data_analysis/configurer.py
```diff
@@ -226,6 +226,8 @@ def __init__(self, cfg_dict, origin=None):
                                      disp_dict['dispatcher_callback_url_base'],
                                      disp_dict['secret_key'],
                                      disp_dict.get('token_max_refresh_interval', 604800),
+                                     disp_dict.get('soft_minimum_folder_age_days', 1),
```
volodymyrss, line 229: let's set a default of 5 days

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -322,39 +321,108 @@ def __init__(self, app,
...
+                    token_expired = True
+
+                job_monitor_path = os.path.join(scratch_dir, 'job_monitor.json')
+                with open(job_monitor_path, 'r') as jm_file:
+                    monitor = json.load(jm_file)
+                    job_status = monitor['status']
+                    job_id = monitor['job_id']
+                if job_status == 'done' and (token is None or token_expired):
+                    list_scratch_dir_to_delete.append(scratch_dir)
+                elif job_status != 'done' and token_expired:
```
volodymyrss, line 378: ok, but you still need an "else", to account for the other options.
Right now, there is a path where it's not deleted and there is no warning

burnout87: that is a case of an unfinished job, with still valid token, nothing to do in that case

volodymyrss: we need to complain since it means that the job is likely stuck, as it has been running for 5 days.
it's not as bad as stuck and token expired, but still unfortunate.

tests/test_job_management.py
```diff
@@ -2409,6 +2409,83 @@ def test_email_t1_t2(dispatcher_long_living_fixture,
         assert jdata["error_message"] == error_message
 
 
+@pytest.mark.parametrize("number_folders_to_delete", [1, 8])
+@pytest.mark.parametrize("soft_minimum_age_days", ["not_provided", 1, 5])
```
volodymyrss, line 2413: you do not test hard limit?

burnout87: It's actually tested afterwards, simply because I thought of having it only as part of the configuration of the dispatcher (default to 30 days). Shall we have it as an argument for the endpoint?

cdci_data_analysis/pytest_fixtures.py
```diff
@@ -694,6 +696,39 @@ def empty_products_user_files_fixture(default_params_dict, default_token_payload
 
     yield scratch_params
 
+@pytest.fixture
+def dispatcher_parameterized_hard_minimum_age_days(pytestconfig, dispatcher_test_conf_fn, dispatcher_debug, request):
```
volodymyrss, line 700: could we doing without duplicating code?

burnout87: That's what I want. But I am a bit confused on how to pass the parametrized configuration (hard limit parameter), to the dispatcher_live_fixture.

Any ideas?

volodymyrss: I would try to make a parametrized variation of `dispatcher_test_conf_fn`, perhaps inheriting from this one, or straight modifying, with an empty default. It takes some attempts.
Maybe to save time now we could have a way without changing config, just changing the files.

burnout87: I parametrized `dispatcher_live_fixture`. but It's not the best solution for sure,

volodymyrss: I imagine it can cause issues. But there is a todo. If it works it's fine

volodymyrss (approved):

cdci_data_analysis/pytest_fixtures.py
burnout87: what I was trying to achieve is a bit convoluted and it probably requires quite more time, but it would be cool to have it working in the future

PR #498, opened 2022-12-20
using bool instead of the deprecated numpy type

burnout87:
If more numpy-related changes are necessary, this is a good place where to implement those

### DIFF
--- a/cdci_data_analysis/plugins/dummy_plugin/instr_catalog.py
+++ b/cdci_data_analysis/plugins/dummy_plugin/instr_catalog.py
@@ -66,9 +66,9 @@ class MyInstrCatalog(BasicCatalog):
 
 
         self.add_column(data=NEW_SOURCE, name='NEW_SOURCE')
-        self.add_column(data=ISGRI_FLAG, name='ISGRI_FLAG', dtype=np.int32)
-        self.add_column(data=FLAG, name='FLAG', dtype=np.int32)
-        self.add_column(data=ERR_RAD, name='ERR_RAD', dtype=np.float64)
+        self.add_column(data=ISGRI_FLAG, name='ISGRI_FLAG', dtype=int)
+        self.add_column(data=FLAG, name='FLAG', dtype=int)
+        self.add_column(data=ERR_RAD, name='ERR_RAD', dtype=float)
 
     @classmethod
     def build_from_dict_list(cls, distlist):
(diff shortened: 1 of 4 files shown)


cdci_data_analysis/plugins/dummy_plugin/instr_catalog.py
```diff
@@ -66,9 +66,9 @@ def __init__(self,
 
 
         self.add_column(data=NEW_SOURCE, name='NEW_SOURCE')
-        self.add_column(data=ISGRI_FLAG, name='ISGRI_FLAG', dtype=np.int)
-        self.add_column(data=FLAG, name='FLAG', dtype=np.int)
-        self.add_column(data=ERR_RAD, name='ERR_RAD', dtype=np.float)
+        self.add_column(data=ISGRI_FLAG, name='ISGRI_FLAG', dtype=np.int32)
+        self.add_column(data=FLAG, name='FLAG', dtype=np.int32)
+        self.add_column(data=ERR_RAD, name='ERR_RAD', dtype=np.float64)
```
volodymyrss, line 71: how do you choose the precision (32 or 64)? It's in the table?

dsavchenko: I've chosen it rather arbitrarily, it's just a dummy plugin for tests. I decided that int32 is enough for flags.
Or we could use python types here, this corresponds to what was previously meant by deprecated types, as far as I understand from the deprecation notice.

volodymyrss: > I've chosen it rather arbitrarily, it's just a dummy plugin for tests.

right, indeed

> I decided that int32 is enough for flags. 

> Or we could use python types here, this corresponds to what was previously meant by deprecated types, as far as I understand from the deprecation notice.

This seems to be the most straightforward way, if it works

burnout87:
Seems like the `all-sky` plugin needs some adaptions

burnout87:
All tests are passing now, I suppose we can merge, unless we forgot something

cdci_data_analysis/plugins/dummy_plugin/instr_catalog.py
volodymyrss, line 71:
```suggestion
        self.add_column(data=ISGRI_FLAG, name='ISGRI_FLAG', dtype=int)
        self.add_column(data=FLAG, name='FLAG', dtype=int)
        self.add_column(data=ERR_RAD, name='ERR_RAD', dtype=float)
```

volodymyrss:
> All tests are passing now, I suppose we can merge, unless we forgot something

there is an unresolved conversation here  https://github.com/oda-hub/dispatcher-app/pull/498/files#r1053537555

volodymyrss (approved):

PR #503, opened 2023-01-02
Renku ini config file and add it to the main commit

burnout87:


burnout87:
Implementing suggestion https://github.com/SwissDataScienceCenter/renku-ui/issues/2217#issuecomment-1357848759

burnout87:
It seems to be working, my only concern is related to the already existing branches. So, for backward-compatibility we could push the config change if necessary. What do you think? 

[Screencast 2023-01-02 15:52:42.webm](https://user-images.githubusercontent.com/3757833/210248224-64e70a9c-7fbe-4d5a-a853-47e32ac8450b.webm)

volodymyrss:
> It seems to be working, my only concern is related to the already existing branches. So, for backward-compatibility we could push the config change if necessary. What do you think? 

Ok, cool. 
For existing branches: since this is new version of the repository generation, the branch name should be different. Then both the old branches still work (except for this issue of renku) and new ones can co-exist.
Branch name can contain also `renku.ini` hash.
It could also just container "repository generator version" but it's difficult to track well. Instead, adding renku.ini "version" is easy.

burnout87:
> 
> Ok, cool. For existing branches: since this is new version of the repository generation, the branch name should be different. Then both the old branches still work (except for this issue of renku) and new ones can co-exist. Branch name can contain also `renku.ini` hash. It could also just container "repository generator version" but it's difficult to track well. Instead, adding renku.ini "version" is easy.

Ok. But by adding this hash to the branch name we will start creating completely new branches, thus no longer using the already existing ones. Which I guess is fine, just that those previously created branches will not be accessed anymore.

volodymyrss:
> > Ok, cool. For existing branches: since this is new version of the repository generation, the branch name should be different. Then both the old branches still work (except for this issue of renku) and new ones can co-exist. Branch name can contain also `renku.ini` hash. It could also just container "repository generator version" but it's difficult to track well. Instead, adding renku.ini "version" is easy.
> 
> Ok. But by adding this hash to the branch name we will start creating completely new branches, thus no longer using the already existing ones. Which I guess is fine, just that those previously created branches will not be accessed anymore.

Previously used branches will be used in links which were sent before (we can also define an expiration date for those).
New ones are going to be created and used in all new requests.
This way it is backward-compatible.

burnout87:
still not 100% convinced the link is working, I described it in https://github.com/SwissDataScienceCenter/renku-ui/issues/2217#issuecomment-1370700382

burnout87:
> still not 100% convinced the link is working, I described it in [SwissDataScienceCenter/renku-ui#2217 (comment)](https://github.com/SwissDataScienceCenter/renku-ui/issues/2217#issuecomment-1370700382)

Following the discussion with the renku team, the image is now fixed within the renku configuration file.

cdci_data_analysis/analysis/renku_helper.py
```diff
@@ -219,6 +282,17 @@ def generate_nb_hash(nb_obj):
...
 
 
+def create_renku_ini_config_obj(repo, default_url_file_name):
+    repo_dir = repo.working_dir
+
+    renku_ini_path = os.path.join(repo_dir, '.renku', 'renku.ini')
+
+    renku_config = ConfigParser()
+    renku_config.read(renku_ini_path)
+    renku_config['renku "interactive"']['default_url'] = f'/lab/tree/{default_url_file_name}'
```
volodymyrss, line 292: just wondering, as we now have unique branch, is it possible to set the file name (and the default url name) to something simple, like `api.ipynb`? It looks a bit less nice when it's shown like it is now.

burnout87: I agree it'd look much cleaner, but by doing so the content of the renku ini file will always be the same, and so the resulting hash will. I suppose is no gonna be a major issue, but we should keep it in mind.

volodymyrss: hopefully it will be ok

volodymyrss (dismissed):

cdci_data_analysis/analysis/renku_helper.py
burnout87: It is now fixed to `api_copde.ipynb`

volodymyrss (dismissed):

volodymyrss (approved):

PR #504, opened 2023-01-04
Development instrument list

burnout87:


burnout87:
Within the approach I am implementing:

- there now is an attribute specific to the instrument telling whether the instrument is a development one
- an endpoint that returns the list of instruments based on the roles of the user

- does this approach make sense? 
- can the endpoint be used to address https://github.com/oda-hub/mmoda-frontend-module/issues/187 ?

dsavchenko:
oda_api uses "/api/instr-list" endpoint to list the available instruments, and it would probably be better if only instruments, available to the user are reported there

burnout87:
> oda_api uses "/api/instr-list" endpoint to list the available instruments, and it would probably be better if only instruments, available to the user are reported there

Once this is complete we will change that, so, potentially we can deprecate `/api/instr-list`

burnout87:
Need [to merge](https://github.com/oda-hub/dispatcher-plugin-integral/pull/114) and https://github.com/oda-hub/oda_api/pull/176 to have all the tests pass

dsavchenko:
Do we also want to restrict the possibility to actually query the instrument? Could be done by separate PR of course.

tests/test_server_basic.py
```diff
@@ -132,14 +132,14 @@ def test_empty_request(dispatcher_live_fixture):
...
-    print("content:", c.text)
+    print("test_empty_request content output:", c.text)
 
     jdata=c.json()
 
     assert c.status_code == 400
 
     # parameterize this
-    assert sorted(jdata['installed_instruments']) == sorted(['empty', 'empty-async', 'empty-semi-async']) or \
+    assert sorted(jdata['installed_instruments']) == sorted(['empty', 'empty-async', 'empty-semi-async', 'empty-development']) or \
```
dsavchenko, line 142: This is not generally presented to the user, but should it really list all the instruments?
It could be not so easy to change, so not critical.

tests/test_server_basic.py
```diff
@@ -400,6 +400,36 @@ def test_download_products_public(dispatcher_long_living_fixture, empty_products
...
+    encoded_token = jwt.encode(token_payload, secret_key, algorithm='HS256')
+
+    c = requests.get(server + "/get-user-specific-instrument-list",
+                     params={"token": encoded_token})
+
+    jdata = c.json()
+
+    assert isinstance(jdata, list)
+    assert 'empty-development' in jdata
+
```
dsavchenko, line 431: Would be also good to test that it's hidden for non-developers

dsavchenko (changes requested):

cdci_data_analysis/flask_app/app.py
```diff
@@ -259,6 +259,13 @@ def inspect_state():
     return state_data_obj
 
 
+# TODO perhaps the way we name the functionality should be better phrased
+@app.route('/get-user-specific-instrument-list')
```
volodymyrss, line 263: it should rather be user-specific by default. We also want to actively prevent users from accessing other instruments.

I.e. the regular old endpoint name would work

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -411,6 +411,34 @@ def free_up_space(app):
...
+            secret_key = app_config.secret_key
+            output, output_code = tokenHelper.validate_token_from_request(token=token, secret_key=secret_key,
+                                                                          action="getting the list of instrument")
+            decoded_token = tokenHelper.get_decoded_token(token, secret_key)
+            roles = tokenHelper.get_token_roles(decoded_token)
+            if output_code is not None:
+                return make_response(output, output_code)
+
+        out_instrument_list = []
+        for instrument_factory in importer.instrument_factory_list:
```
volodymyrss, line 430: we can put the method to check if the user email and roles are sufficient for the instrument inside the instrument.

in the future we might change how the selection is done. E.g. some instruments will belong to some users.

cdci_data_analysis/flask_app/app.py
burnout87: renamed. Also once this will be merged we will adapt oda_api to use this endpoint so we can safely remove the old endpoint

cdci_data_analysis/analysis/queries.py
```diff
@@ -316,6 +316,8 @@ def __init__(self,
 
         super(InstrumentQuery, self).__init__(name,parameters_list)
 
+    def check_instrument_roles(self, roles):
+        raise RuntimeError(f'{self}: check_instrument_access needs to be implemented for the {self.name} instrument')
```
volodymyrss, line 320: this should check for development role in the `roles`, for the `development` attribute in the instance, and return accordingly. This is the useful default, no?

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -411,6 +411,35 @@ def free_up_space(app):
...
+            instrument = instrument_factory()
+
+            if instrument.development:
```
volodymyrss, line 433: this logic can all go in the method `check_access`

cdci_data_analysis/analysis/queries.py
burnout87: actually only checking the development role is within `roles`

development attribute is checked outside

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: now inside the `check_instrument_access`

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: It is now like that: in case a certain restrictive access rule, needs to be implemented for an instrument, it is now possible

cdci_data_analysis/analysis/instrument.py
```diff
@@ -93,6 +93,7 @@ def __init__(self,
                  instrumet_query,
                  input_product_query=None,
                  asynch=True,
+                 development=False,
```
dsavchenko, line 96: As far as the behaviour can be more general, not only checking for 'development' role, this should probably be renamed to smth like "restricted_access"

cdci_data_analysis/analysis/queries.py
dsavchenko: if error raising will stay the default here, error message is misleading: `check_instrument_access` actually is implemented and it checks for the attribute. One must implement this actual method

cdci_data_analysis/flask_app/app.py
```diff
@@ -259,6 +259,12 @@ def inspect_state():
     return state_data_obj
 
 
+@app.route('/instr-list')
```
dsavchenko, line 262: There are now two endpoints: `/api/instr-list` and `/instr-list` which could be misleading
If you intend to deprecate the one with 'api', probably add deprecation warning?

cdci_data_analysis/analysis/queries.py
volodymyrss: Default should be compatible with what we had before, i.e.:

```python
(self.development and ('oda workflow developer' in roles)) or (not self.development)
```

By the way, maybe good to already now pass also the tocket subject (email) to the method. We will use it in the future.

dsavchenko (changes requested):

cdci_data_analysis/analysis/queries.py
burnout87: > if error raising will stay the default here, error message is misleading: `check_instrument_access` actually is implemented and it checks for the attribute. One must implement this actual method

you're right I forgot to change the message once renamed the method

volodymyrss:
> Do we also want to restrict the possibility to actually query the instrument? Could be done by separate PR of course.

We should do it, and why not here.

cdci_data_analysis/analysis/instrument.py
burnout87: yeah makes sense

cdci_data_analysis/flask_app/app.py
burnout87: probably we should directly remove `/api/instr-list`, and adapt oda_api in turn

burnout87: > There are now two endpoints: `/api/instr-list` and `/instr-list` which could be misleading If you intend to deprecate the one with 'api', probably add deprecation warning?

https://github.com/oda-hub/oda_api/pull/178

burnout87: do we want to keep the endpoint as `/api/instr-list` od we can remove the `/api` ?

volodymyrss: what do you think, what are the reasons in favor of each choice? Which one follows the approach we used in the rest of the application?

If you do https://github.com/oda-hub/oda_api/pull/178 it breaks compatibility. It absolutely has to have deprecation message and a redirection.

burnout87: redirection now takes place

volodymyrss: > redirection now takes place

wait, but concerning this:

> what do you think, what are the reasons in favor of each choice? Which one follows the approach we used in the rest of the application?

why did we need to change the endpoint removing the "/api/" ?

cdci_data_analysis/plugins/dummy_plugin/empty_development_instrument.py
```diff
@@ -0,0 +1,96 @@
...
+from .empty_instrument import BoundaryFloat
+
+# duplicated with jemx, but this staticmethod makes it complex.
+# this all should be done commonly, for all parameters - limits are common thing
+from ...analysis.parameters import SpectralBoundary, Angle, Energy
+
+
+# class InstrumentWithCheckQuery(InstrumentQuery):
+#     def check_instrument_roles(self, roles, email):
+#         return True
```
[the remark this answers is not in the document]
volodymyrss: maybe we can remove the commented out code?.. @burnout87 
overall, would be better not to copy the section but reuse. but ok

tests/test_server_basic.py
volodymyrss: I am not sure exactly what you mean so it's best that you try to implement the suggestion so we can see the details

cdci_data_analysis/analysis/instrument.py
```diff
@@ -486,6 +489,14 @@ def get_product_query_name(self, product_type):
         else:
             return self.query_dictionary[product_type]
 
+
+    def check_instrument_access(self, roles, email):
+        if self.restricted_access:
+            return self.instrumet_query.check_instrument_roles(roles, email)
```
volodymyrss, line 495: `check_instrument_roles` also can be renamed

cdci_data_analysis/analysis/instrument.py
```diff
@@ -486,6 +489,14 @@ def get_product_query_name(self, product_type):
         else:
             return self.query_dictionary[product_type]
 
+
+    def check_instrument_access(self, roles, email):
+        if self.restricted_access:
```
volodymyrss, line 494: this kind of two-level access specification is not quite what I meant. Rather "development" can be an attribute of InstrumetQuery and all logic done in `instrumet_query.check_instrument_roles` - it's easier to see access logic then. Any disadvantage of this approach?

cdci_data_analysis/analysis/instrument.py
burnout87: any ideas on how ?

cdci_data_analysis/analysis/instrument.py
burnout87: But my idea is to already have the access login within `instrument_query.check_instrument_roles`, as I am doing. But I had never seen the `development`/ `restricted_access` as an attribute of the InstrumentQuery, rather of the Instrument itself. 

I will try what you propose

cdci_data_analysis/flask_app/app.py
burnout87: To actually have it more aligned with the majority of endpoint names we implemented. But I don't really now the inital reasons why there was `/api`

cdci_data_analysis/analysis/instrument.py
volodymyrss: You mean you never seen it where? We are just adding it here, no?

cdci_data_analysis/analysis/instrument.py
volodymyrss: `check_instrument_access`

cdci_data_analysis/analysis/instrument.py
burnout87: Simply I meant that I was never considering the option of having it as an attribute of `InstrumetQuery`

And yes, I added it only here

cdci_data_analysis/flask_app/app.py
volodymyrss: all of the analogous ones are still with `/api/`. But indeed the newer ones do not have it. We should change them all at once at some point. For here, what you added now is ok.

tests/test_server_basic.py
burnout87: I double-checked : this list is built using the `importer.instrument_factory_list`, so indeed it would not be so simple to change the logic for building such a list

burnout87:
> > Do we also want to restrict the possibility to actually query the instrument? Could be done by separate PR of course.
> 
> We should do it, and why not here.

Will also implement this. If I am not mistaken, it should not take too long

burnout87:
> 
> Will also implement this. If I am not mistaken, it should not take too long

It now checks if an instrument can actually be set, based on it access rules

burnout87:
I noticed tests for `LegacySurvey` are failing because within `LegacySurvey` implements the instrument query by not extending the class `Intrument_Query`. Is this intended? Is it possible to change and how long would it take?

volodymyrss:
> I noticed tests for `LegacySurvey` are failing because within `LegacySurvey` implements the instrument query by not extending the class `Instrument_Query`. Is this intended? Is it possible to change and how long would it take?

@dsavchenko what do you think?
[BaseQuery](https://github.com/oda-hub/dispatcher-plugin-legacysurvey/blob/755b995e973d12c067f8b03b5b91696774870638/dispatcher_plugin_legacysurvey/queries.py#L7)

burnout87:
> I noticed tests for `LegacySurvey` are failing because within `LegacySurvey` implements the instrument query by not extending the class `Instrument_Query`. Is this intended? Is it possible to change and how long would it take?

I also wonder whether the same approach has been used elsewhere

volodymyrss:
> > I noticed tests for `LegacySurvey` are failing because within `LegacySurvey` implements the instrument query by not extending the class `Instrument_Query`. Is this intended? Is it possible to change and how long would it take?
> 
> I also wonder whether the same approach has been used also elsewhere

if yes, maybe we could move the `check_access` method to `BaseQuery`.

burnout87:
> 
> if yes, maybe we could move the `check_access` method to `BaseQuery`.

it could work but it's not ideal since the `check_instrument_access` is something specific to the instrument, and `BaseQuery` is used more broadly, ie  to define `SourceQuery` and `ProductQuery`

another way would be to check if the `check_instrument_access` is an attribute of the `instrumet_query` object, but this  requires multiple changes

dsavchenko:
> I noticed tests for `LegacySurvey` are failing because within `LegacySurvey` implements the instrument query by not extending the class `Instrument_Query`. Is this intended? Is it possible to change and how long would it take?

I made it this way in all plugins I created. This was done because several parameters are hard-coded in the InstrumentQuery class, and not every instrument needs them. So it's kind of workaround. 


This kind of workaround is done differently in spi_acs plugin, but it produces a weird piece of metadata for instrument query with `None` parameter and catalog parameters, actually unusable
```
  '[{"query_name": "spiacs_parameters"}, {"name": "user_catalog", "units": None, "value": None}, {"name": None, "units": None, "value": []}, {"name": "selected_catalog", "units": None, "value": None}]',
```

burnout87:
> Like this [oda-hub/dispatcher-plugin-legacysurvey#9](https://github.com/oda-hub/dispatcher-plugin-legacysurvey/pull/9)

It looks like that also `nb2workflow` fails for the same reason

volodymyrss:
> > Like this [oda-hub/dispatcher-plugin-legacysurvey#9](https://github.com/oda-hub/dispatcher-plugin-legacysurvey/pull/9)
> 
> `nb2workflow` fails for the same reason

you mean [dispatcher-plugin-nb2workflow ](https://github.com/oda-hub/dispatcher-plugin-nb2workflow/blob/16480002217721cb5b1fa0133f221afd8f3b834f/dispatcher_plugin_nb2workflow/queries.py#L136) (and also gw plugin), not nb2workflow itself

burnout87:
> > > Like this [oda-hub/dispatcher-plugin-legacysurvey#9](https://github.com/oda-hub/dispatcher-plugin-legacysurvey/pull/9)
> > 
> > 
> > `nb2workflow` fails for the same reason
> 
> you mean [dispatcher-plugin-nb2workflow ](https://github.com/oda-hub/dispatcher-plugin-nb2workflow/blob/16480002217721cb5b1fa0133f221afd8f3b834f/dispatcher_plugin_nb2workflow/queries.py#L136) (and also gw plugin), not nb2workflow itself

exactly

dsavchenko:
The last conflicting plugin https://github.com/oda-hub/dispatcher-plugin-gw/pull/14
Sorry for the delay, I used there feature which depends on other PR. Now I've decided to revert this change, not to delay your work

volodymyrss (approved):

dsavchenko (approved):

PR #510, opened 2023-01-25
missing sentry dsn after url call fails and as arg to a number of calls

burnout87:

PR #511, opened 2023-01-25
invalid instr list

volodymyrss:
this is critical since it prevents listing instruments with old-style url, still in production


volodymyrss:
please have a look @burnout87 but I will merge as soon as most of the tests except for the one there https://github.com/oda-hub/dispatcher-app/pull/510 pass

PR #513, opened 2023-01-25
Fix redirection instr list

burnout87:


tests/test_server_basic.py
```diff
@@ -471,14 +471,101 @@ def test_query_restricted_instrument(dispatcher_live_fixture):
...
+    server = dispatcher_live_fixture_with_local_products_url
+
+    logger.info("constructed server: %s", server)
+
+    c = requests.get(os.path.join(server, "api/instr-list"), allow_redirects=allow_redirect)
+
+    if not allow_redirect:
+        assert c.status_code == 302
+        redirection_url = c.headers["Location"]
+        assert "instr-list" in redirection_url
```
volodymyrss, line 486: dont you want to check that it uses the config url and not the bind host?

burnout87: This test was indeed not really exhaustive, I am adding more

burnout87: now it's tested

cdci_data_analysis/flask_app/app.py
```diff
@@ -90,8 +84,24 @@ def run_api_parameters():
...
+                   '\'/instr-list\' endpoint. Please use this one in the future.\n')
+
+    bind_host = app.config['conf'].bind_host
+    bind_port = app.config['conf'].bind_port
+
+    parsed_product_url = urlparse(app.config['conf'].products_url)
+
+    if (parsed_product_url.hostname is None and parsed_product_url.port is None) or \
+            (parsed_product_url.hostname == bind_host and parsed_product_url.port == bind_port):
+        redirection_url = url_for('instr_list')
```
volodymyrss, line 97: why is this needed? will the second branch always be enough?

cdci_data_analysis/flask_app/app.py
```diff
@@ -90,8 +84,24 @@ def run_api_parameters():
...
+    else:
+        redirection_url = os.path.join(app.config['conf'].products_url, 'dispatch-data/instr-list')
+
+    if request.args:
+        redirection_url += f'?{urlencode(request.args)}'
```
volodymyrss, line 102: do we only have `get` here?

cdci_data_analysis/flask_app/app.py
burnout87: The `dispatch-data` part needs to be added, that is something we do also when building the url for the email, we add it  to the `products_url` from the config

cdci_data_analysis/flask_app/app.py
burnout87: yes

volodymyrss: where is it restricted?

burnout87: are we talking about the method for this endpoint?

volodymyrss: yes

burnout87: default to GET

cdci_data_analysis/flask_app/app.py
```diff
@@ -90,8 +84,24 @@ def run_api_parameters():
...
-    return instrument_list
+    logger.warning('\nThe endpoint \'/api/instr-list\' is deprecated and you will be automatically redirected to the '
+                   '\'/instr-list\' endpoint. Please use this one in the future.\n')
+
+    bind_host = app.config['conf'].bind_host
+    bind_port = app.config['conf'].bind_port
+
+    parsed_product_url = urlparse(app.config['conf'].products_url)
+
+    if (parsed_product_url.hostname is None and parsed_product_url.port is None) or \
```
volodymyrss, line 95: all is this quite convoluted. Maybe it is possible to use `request.referrer` ?

burnout87: Tried a bit, but it always returns an empty value. Is there something I am missing?

volodymyrss: maybe `request.url`?

burnout87: `request.url` does not return the bind_host and bind_port, but the url that flask assigns when starting the app.

How could use it in this case?

volodymyrss: you do not need bind, you need to know URL used to call the service. usually, it can be found. can you investigate how to discover this url?

volodymyrss: you could use it as a base for redirect. Do make sure it makes sense.
But, what you want, is to return to the user the same URL they used, just replacing the "/api/" part, right? 
It's good to check for clarity that it matches that in the config

burnout87: Yes, basically, replacing the `/api` part

burnout87: It now uses `request.url`, but I am a bit stuck on how to test that `products_url` within the config matches the one used for building `redirection_url`

volodymyrss: products url is a frontend url. if it's set (like in production and some tests) it should match the redirect with dispatch-data. if not, it should be ignored

burnout87: Checks now if products_url from config is set and is a valid url. Done by using the `validators` library

volodymyrss (approved):

PR #514, opened 2023-01-26
adjusted sentry trace rate

volodymyrss:
we get too many  transations, alerts (which are actually not errors). This becomes expensive, prevents free tier, etc.


volodymyrss:
I will claim this to be a hotfix since it really affects operations. And merge

PR #519, opened 2023-01-27
Support new version renku ini file

burnout87:


burnout87:
I updated my private test repo to the newer version of the renku.ini file, and I also created another private test repository to test both renku ini version.

volodymyrss:
I do not know if we want to keep old-style keys, just make sure we do not crash with them, and create new key always.

burnout87:
> I do not know if we want to keep old-style keys, just make sure we do not crash with them, and create new key always.

Absolutely, this makes sure that no crash (hopefully) happens. Currently it also supports the older approach and in principle it does not create anything new

cdci_data_analysis/analysis/renku_helper.py
```diff
@@ -259,6 +259,10 @@ def update_and_commit_default_url_renku_ini(repo, config_obj, user_name=None, us
 def generate_ini_file_hash(config_ini_obj):
     try:
         ini_config_dict = { s:dict(config_ini_obj.items(s)) for s in config_ini_obj.sections() }
+    except:
```
volodymyrss, line 262: please catch an exception and report it.

cdci_data_analysis/analysis/renku_helper.py
```diff
@@ -259,6 +259,10 @@ def update_and_commit_default_url_renku_ini(repo, config_obj, user_name=None, us
...
+        logger.error(f'Unable to generate a dictionary starting from the ini config file: {config_ini_obj}')
+        raise Exception(f'Unable to generate a dictionary starting from the ini config file: {config_ini_obj}')
+    try:
         ini_hash = make_hash(ini_config_dict)
     except:
```
volodymyrss, line 267: also here, please catch and report

cdci_data_analysis/analysis/renku_helper.py
```diff
@@ -287,15 +291,20 @@ def create_renku_ini_config_obj(repo, default_url_file_name):
...
 
     try:
-        renku_config['renku "interactive"']['default_url'] = f'/lab/tree/{default_url_file_name}'
-    except KeyError:
-        renku_config['interactive']['default_url'] = f'/lab/tree/{default_url_file_name}'
+        if 'renku "interactive"' in config_ini_obj:
+            config_ini_obj['renku "interactive"']['default_url'] = f'/lab/tree/{default_url_file_name}'
+        elif 'interactive' in config_ini_obj:
+            config_ini_obj['interactive']['default_url'] = f'/lab/tree/{default_url_file_name}'
+    except:
```
volodymyrss, line 302: please catch

cdci_data_analysis/analysis/renku_helper.py
```diff
@@ -287,15 +291,20 @@ def create_renku_ini_config_obj(repo, default_url_file_name):
...
-    renku_config = ConfigParser()
-    renku_config.read(renku_ini_path)
+    config_ini_obj = ConfigParser()
+    config_ini_obj.read(renku_ini_path)
 
     try:
-        renku_config['renku "interactive"']['default_url'] = f'/lab/tree/{default_url_file_name}'
-    except KeyError:
-        renku_config['interactive']['default_url'] = f'/lab/tree/{default_url_file_name}'
+        if 'renku "interactive"' in config_ini_obj:
```
volodymyrss, line 298: but you are still keeping the old-style key, why? We want to discard them

burnout87: Because I don't create any new file, I just change the content by modifying the `default_url` entry, and I made it in a way that it still supports the old approach without modifying its content.

volodymyrss: You are "modifying the default_url entry" and later write it in a file, right? How is this "without modifying its content"?

volodymyrss: You do not just modify, you actually write a file from scratch from the ini object which you constructed from the previous file. You can as well remove the old-style field and add the new-style one. And show a warning that the repository is updated.

volodymyrss: By the way, how does renku treat this field? Could you have a look in renku-python code to see if the old value is still supported? If no - it's imperative to update to new. If still supported - still good to update, why not? Any disadvantage of upgrading this feature of the repository in this push branch?

cdci_data_analysis/pytest_fixtures.py
```diff
@@ -480,7 +480,7 @@ def dispatcher_test_conf_with_renku_options_fn(dispatcher_test_conf_fn):
             f.write(f_default.read())
 
         f.write('\n    renku_options:'
-                '\n        renku_gitlab_repository_url: "git@renkulab.io:gabriele.barni/test-dispatcher-endpoint.git"'
+                '\n        renku_gitlab_repository_url: "git@renkulab.io:gabriele.barni/old-test-dispatcher-endpoint.git"'
```
volodymyrss, line 483: did you test with both new and old repositories? maybe no need to add since it can not be now done in CI anyway, just to be sure?

cdci_data_analysis/analysis/renku_helper.py
volodymyrss: Just to make sure, did you manage to have a look on this question? @burnout87

volodymyrss (approved):

PR #520, opened 2023-02-13
Ontology-aware parameter creation

dsavchenko:


dsavchenko:
Test failing for nb2workflow plugin is expected, because its master brunch is tested. 
This test passes in https://github.com/oda-hub/dispatcher-plugin-nb2workflow/pull/40

cdci_data_analysis/configurer.py
```diff
@@ -250,7 +250,7 @@ def __init__(self, cfg_dict, origin=None):
                                      disp_dict.get('product_gallery_options', {}).get('converttime_revnum_service_url', 'https://www.astro.unige.ch/mmoda/dispatch-data/gw/timesystem/api/v1.0/converttime/UTC/{}/REVNUM'),
                                      disp_dict.get('renku_options', {}).get('renku_gitlab_repository_url', None),
                                      disp_dict.get('renku_options', {}).get('renku_base_project_url', None),
-                                     disp_dict.get('renku_options', {}).get('ssh_key_path', None)
+                                     disp_dict.get('renku_options', {}).get('ssh_key_path', None),
```
volodymyrss, line 253: did you try to add option for ontology? we should probably.

volodymyrss:
Looks ok, cool. I want to look just a bit deeper. Meanwhile, is it possible to have just a couple of representative tests with dispatcher live fixture? To check the full chain.

cdci_data_analysis/configurer.py
dsavchenko: I tried, then reverted.  
This functionality is used only in nb2workflow plugin, so I added the option there. This requires less changes than passing this option all the way from flask app up to Queries (which are anyway initialized in the plugin code).

dsavchenko:
> is it possible to have just a couple of representative tests with dispatcher live fixture?

It will anyway be in the nb2workflow-plugin. I will try to add as separate PR there.

volodymyrss:
> > is it possible to have just a couple of representative tests with dispatcher live fixture?
> 
> It will anyway be in the nb2workflow-plugin. 

It's good that something will be there of course, but since the code is here, something is good to have here too. 

> I will try to add as separate PR there.

Ok! Just please create it, or an issue, or TODO, before merging this, so we can mention it here and track the reference.

volodymyrss (dismissed):

dsavchenko:
Here is the test for full stack nb2workflow -> dispatcher with nb2workflow-plugin -> oda_api request
https://github.com/oda-hub/dispatcher-plugin-nb2workflow/pull/41

volodymyrss (approved):

dsavchenko:
@burnout87 do you want/have time to see on it?

cdci_data_analysis/analysis/ontology.py
```diff
@@ -0,0 +1,364 @@
...
+    
+    def parse_extra_ttl(self, extra_ttl, parse_oda_annotations = True):
+        if parse_oda_annotations:
+            tmpg = rdf.Graph()
+            tmpg.parse(data = extra_ttl)       
+            try:
+                self.parse_oda_annotations(tmpg)
+            except RuntimeError as e:
+                raise RequestNotUnderstood(e.message)
+            extra_ttl = tmpg.serialize(format='turtle')
```
burnout87, line 231: In the plugin I have till now used `n3` which is basically a superset of ttl, perhaps we want to have them aligned, even though nothing should change

burnout87 (commented):
It looks good overall, I have a couple of questions, but they're minor

cdci_data_analysis/analysis/ontology.py
dsavchenko: here, the serialisation is used only internally, to feed it to `self.g.parse()` (in the next line), so any format would generally work, I use turtle as it is the default expected format for `parse` 
Also, this method is used to parse `extra_ttl` from nb2workflow, which is turtle.

(But I will extend this method a bit, so that any valid triple format can be used)

cdci_data_analysis/analysis/ontology.py
```diff
@@ -0,0 +1,364 @@
...
+    
+    @property
+    def graph(self):
+        return deepcopy(self._g)
+    
+    def reset(self, ontology_path, version):
+        if version != self._ver or ontology_path != self._path:
+            self._g = rdf.Graph()
+            self._g.parse(ontology_path)
+            self._g.bind('oda', ODA)
```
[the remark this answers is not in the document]
dsavchenko: I believe we opted to only use `http://` terms everywhere for consistency. But the prefix could be safely added, just in case `https://` is ocasionaly used somehere

burnout87 (dismissed):

volodymyrss (approved):

PR #532, opened 2023-04-26
Endpoint astro entity info

burnout87:


burnout87:
Endpoint to retrieve from the product gallery a number of info of a given source. To be used for https://github.com/oda-hub/mmoda-frontend-module/issues/194

volodymyrss:
> No checks on the token is performed, so the endpoint is actually open.

I am not quite sure what associates source/astroentity with an observation. Is it when the source was in the FoV or when it was detected? @ferrigno knows probably.

If it's when it is detected, this can be private information.

ferrigno:
It is detection, as we ingest products generally only if a source is detected, sometimes when it is explicitly asked.
However, the policy of INTEGRAL private data is quite loose now, as all data can be downloaded and nalyzedd in the CONS format. The responsibility lies on the publication and for this, you need more than just a detection.

volodymyrss:
> It is detection, as we ingest products generally only if a source is detected, sometimes when it is explicitly asked. However, the policy of INTEGRAL private data is quite loose now, as all data can be downloaded and nalyzedd in the CONS format. The responsibility lies on the publication and for this, you need more than just a detection.

Ok, just that we are clear on what's being public (not much of course, but still something) and there will be no surprises

volodymyrss (approved):

PR #541, opened 2023-06-02
load oda_site parameters for email from config

burnout87:


dsavchenko (approved):
Thank you!

volodymyrss (approved):

volodymyrss:
could we add a default option here, @burnout87 ? good to keep as we do in many cases

PR #545, opened 2023-06-20
bug fix token validation in `get_user_specific_instrument_list`

burnout87:


burnout87:
Discovered when working on https://github.com/oda-hub/oda_api/pull/199

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -413,11 +413,12 @@ def get_user_specific_instrument_list(app):
             secret_key = app_config.secret_key
             output, output_code = tokenHelper.validate_token_from_request(token=token, secret_key=secret_key,
                                                                           action="getting the list of instrument")
-            decoded_token = tokenHelper.get_decoded_token(token, secret_key)
-            roles = tokenHelper.get_token_roles(decoded_token)
-            email = tokenHelper.get_token_user_email_address(decoded_token)
             if output_code is not None:
                 return make_response(output, output_code)
+            else:
+                decoded_token = tokenHelper.get_decoded_token(token, secret_key)
```
volodymyrss, line 419: hm, why is this needed?

burnout87: In the case the token was expired or not valid, the function has to return with an error message. 

The way it was before it would crash when attempting to decode a not valid token

volodymyrss: ah, those functions do not parse it. ok

volodymyrss (approved):

PR #568, opened 2023-08-28
Update sentry.py

volodymyrss:


cdci_data_analysis/flask_app/sentry.py
```diff
@@ -38,6 +39,7 @@ def have_sentry(self):
                     traces_sample_rate=0.1,
                     debug=False,
                     max_breadcrumbs=10,
+                    environment=os.getenv("ODA_SENTRY_ENV", "production")
```
burnout87, line 42: should not it be included i the configuration?

volodymyrss: could be. Ok I can not do it now, could you please?

PR #575, opened 2023-08-29
Catch not authorized oda api status details request

burnout87:


cdci_data_analysis/analysis/instrument.py
```diff
@@ -334,6 +334,22 @@ def get_status_details(self,
             sentry.capture_message((f'ConnectionError detected when retrieving additional '
                                     f'information from a completed job '
                                     f'{ce}'))
+        except Unauthorized as ue:
+            detail_message = ""
```
volodymyrss, line 338: the job was submitted correctly so the error response should be different.

What seemed to have happened is that token has expired before the job completed. We should say this and that your result might have been partially or mostly ready, please resubmit the job. Also please consider to submit a job with longer token validity (i.e. login recently before the job submission)

volodymyrss: Also please add that if unclear they should contact us. Maybe even put URL to contact form?

burnout87: inserted the email address as well as a hint at using the "contact us" form on MModa

PR #587, opened 2023-09-11
removing encoding arg json load

burnout87:


burnout87:
tests are all passing, but I am not sure it will solve the issue...let's see

volodymyrss (approved):

PR #590, opened 2023-09-18
Job id included in exception message for sentry

burnout87:


volodymyrss:
are you sure this is going to add to the message to sentry?

it looks like `AnalysisException:ProduceDisabledException` is caught by sentry. If so, it needs to be wrapped and re-raised.

volodymyrss (changes requested):
tests fail

burnout87:
> are you sure this is going to add to the message to sentry?
> 
> it looks like `AnalysisException:ProduceDisabledException` is caught by sentry. If so, it needs to be wrapped and re-raised.

`job_id` is included in the `e_message` here: https://github.com/oda-hub/dispatcher-app/blob/1a582a860c1a0b4f85fb65e39242114748f76b99/cdci_data_analysis/analysis/queries.py#L653-L664

which in turn is sent as a sentry message here: https://github.com/oda-hub/dispatcher-app/blob/1a582a860c1a0b4f85fb65e39242114748f76b99/cdci_data_analysis/analysis/products.py#L196

volodymyrss (approved):

PR #604, opened 2023-10-12
Cc receivers matrix no sentry

burnout87:


burnout87:
the idea is to:

* the list of cc-room_ids can be configured (more tests is still needed)
* in case no user room_id is present in the token, no sentry is sent, but rather a logging message, and execution continues with the sending of the message to the cc-room_ids
* in case no sufficient config is present (no matrix server url or matrix sender access token) matrix cannot be sent, this is checked beforehand sending the message

What do you think of this approach?

Needed configuration would be:
* matrix server url
* sender access token
* cc-room_ids list

cdci_data_analysis/config_dir/conf_env.yml.example
```diff
@@ -80,6 +80,16 @@ dispatcher:
...
+        # list of additional room ids receivers towards which the message will be sent, besides the room id extracted from the token
+        matrix_cc_receivers_room_ids: ['room_id']
+        # incident report related options, for the messages sent via matrix
+        incident_report_matrix_options:
+            # list of room ids receivers towards which the incident report message will be sent
+            matrix_incident_report_receivers_room_ids: ['room_id']
+            # personal access token of the sender of the incident report
+            matrix_incident_report_sender_personal_access_token: MATRIX_INCIDENT_REPORT_SENDER_ACCESS_TOKEN
+        # list of room ids to which the matrix message should be sent
+        matrix_cc_receivers_room_id: []
```
volodymyrss, line 92:
```suggestion
        matrix_bcc_receivers_room_id: []
```

volodymyrss (approved):

volodymyrss (approved):

PR #618, opened 2023-10-27
Matrix msgs readme

burnout87:


volodymyrss (approved):

burnout87:
> Sorry for the late review
> 
> Probably it's better to move it to a separate file and have a link in the main README

Ok thanks, it makes sense

> 
> Do we have a way to add matrix to the token, generated in frontend?

yes, it will be added you can check https://github.com/oda-hub/mmoda-frontend-module/issues/212

PR #636, opened 2023-12-07
Inspect jobs

burnout87:


cdci_data_analysis/flask_app/app.py
```diff
@@ -277,9 +277,34 @@ def free_up_space():
...
+                                                                  action="inspect the state for a given job_id")
+
+    if output_code is not None:
+        return make_response(output, output_code)
     state_data_obj = InstrumentQueryBackEnd.inspect_state(app)
-    return state_data_obj
+    return jsonify(dict(records=state_data_obj['records']))
+
 
+@app.route('/inspect-jobs', methods=['POST', 'GET'])
```
volodymyrss, line 294: could you just make this part of the `inspect-state` with some extra argument?

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -481,89 +472,147 @@ def inspect_state(app):
...
+
+        query_output_fn = os.path.join(scratch_dir, 'query_output.json')
+        try:
+            with open(query_output_fn) as query_output_file:
+                query_output_content = json.load(query_output_file)
+            query_output_status_dict = query_output_content.get('status_dictionary', None)
+            if query_output_status_dict is not None:
+                result_job_status['query_output'] = query_output_status_dict
+        except Exception as e:
+            # write something
```
volodymyrss, line 528: ?

tests/test_job_management.py
```diff
@@ -2639,6 +2640,383 @@ def test_inspect_status(dispatcher_live_fixture, request_cred, roles, include_se
...
+                f'Your privilege roles include {roles}, but the following roles are missing: {lacking_roles}.'
+            )
+
+    inspect_params = dict(
+                         token=encoded_token
+                     )
+    if pass_job_id:
+        inspect_params['job_id'] = job_id_done[:8]
+
+    c = requests.get(server + "/inspect-jobs",
```
volodymyrss, line 2729: I am concerned this will become too slow becasue of all the additional reads. Could you check timing here, and maybe also number of openned files.

tests/test_job_management.py
```diff
@@ -2639,6 +2640,383 @@ def test_inspect_status(dispatcher_live_fixture, request_cred, roles, include_se
...
+                     params=inspect_params)
+
+    if request_cred != 'private' or ('job manager' not in roles):
+        # email not supposed to be sent for public request
+        assert c.status_code == status_code
+        assert c.text == error_message
+    else:
+        jdata= c.json()
+        assert 'jobs' in jdata
```
volodymyrss, line 2738: that's a lot of asserts! Could you maybe use a json schema instead, or something like that?

tests/test_job_management.py
```diff
@@ -2639,6 +2640,383 @@ def test_inspect_status(dispatcher_live_fixture, request_cred, roles, include_se
...
+                    f'scratch_sid_{session_id_failed}_jid_{job_id_failed}')
+            assert (jdata['jobs'][1]['job_status_data'][0]['job_statuses_fn'] ==
+                    f'scratch_sid_{session_id_done}_jid_{job_id_done}' or
+                   jdata['jobs'][1]['job_status_data'][0]['job_statuses_fn'] ==
+                    f'scratch_sid_{session_id_failed}_jid_{job_id_failed}')
+        else:
+            assert jdata['jobs'][0]['job_status_data'][0]['job_statuses_fn'] == f'scratch_sid_{session_id_done}_jid_{job_id_done}'
+
+        assert jdata['jobs'][0]['job_status_data'][0]['job_statuses'][0]['job_status_file'] == 'job_monitor.json'
+        if not pass_job_id:
```
volodymyrss, line 2779: this condition is repeated many times. Anyway, validation could be much more readable

cdci_data_analysis/flask_app/app.py
burnout87: I thought about, but they serve two purposes that are quite different, and the relative outputs are quite different. 

Reason why I preferred to have two separate endpoints...

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: I suppose you refer to the comment...?

tests/test_job_management.py
burnout87: json schema is there, and the in the tests the output will be validated against it

tests/test_job_management.py
burnout87: introducing the schema validation made allowed to remove many checks and made the whole more compact

tests/test_job_management.py
burnout87: indeed there were several opened files that were just read but the content was not included within the output

I reformatted the code so that the two functionalities are now separated

cdci_data_analysis/flask_app/app.py
volodymyrss: Why different, what are the purposes?

They are not called in meaningfully different ways, both return dispatcher state which consists of set of jobs.

cdci_data_analysis/flask_app/dispatcher_query.py
volodymyrss: yes. The comment is not clear.

burnout87: removed

tests/test_job_management.py
volodymyrss: Yes, but check how long it takes in the test. And I suspect you might be able to trace how many files were opened.

cdci_data_analysis/flask_app/app.py
burnout87: `inspect-jobs` returns a list of objects, each related to a specific job. It serves the purpose of reporting the status of a job, by reading the content of the various scratch_dir related to that job, and understand whether or not that job has completed

`inspect-state`returns a list of objects, each containing information about the content of a single scratch_dir, and multiple scratch_dir might be related to the same job

perhaps more appropriate names should be assigned...

tests/test_job_management.py
burnout87: execution times I read (in the tests) range from 0.003 sec to 0.006 sec

cdci_data_analysis/flask_app/app.py
volodymyrss: They both return information about jobs (derived from scratch directories), just grouped differently, and with more details in the new case. `inspect-jobs` in just further development on `inspect-state` and I still do not see why it has to be separate.

To maintain previous interface, we can have an option to return previous-style structure.

tests/test_job_management.py
volodymyrss: Yeah it's hard to check timing in test. 
For the files, can you find a way?

burnout87: 22 files where read when running the `test_inspect_jobs_with_callbacks` test
3 files where read when running `test_inspect_jobs_failed` test

for each inspect-jobs call the following files are opened (and closed):
- `query_output.json`
- `job_monitor*.json` (depending on the number of nodes per job)
- `analysis_parameters.json`

cdci_data_analysis/flask_app/app.py
burnout87: but they both contain information that serve different purposes, and they should be both maintained (perhaps with a more appropriate naming):

- `inspect-state` contains info regarding the email_history, matrix_message_history, session_log, fits_files etc for a single scratch_dir
- `inspect-jobs` contains info "only" related to the job itself: in particular, its status and it also includes the query_output (extracted from the `query_output.json` file), plus, a logic is implemented to understand whether or not that job has completed

So effectively, the information returned from one are not contained into the other (and viceversa).

That is why I think inspect-jobs cannot be seen as an extension of inspect-state

volodymyrss: They contain information serving the same purpose: inspecting state record of the job history, for the purpose of assessing what messages were received and sent in relation to each job. Do they not?

burnout87: With inspect-state the focus is on the content of the single scratch folder.

With inspect-jobs the focus is on the single job.

I see this as the main difference, besides the returned content itself.

In any case, they can be "merged" so that inspect-jobs also includes  what inspect-state returns

tests/test_job_management.py
burnout87: if the output of inspect-state is to be included within the output of inspect-jobs, than also the following files are open:
- `session.log`

cdci_data_analysis/flask_app/app.py
volodymyrss: This sort of focus difference is just difference in rearranging output.

Job info record is state record. So I don't see why it's needed to renamed the endpoint.
Just add a request argument to `inspect-state` something like `group_by_job`. Also maybe other arguments would be useful to reduce the amount of files open, given the test it might end up being some 10x1000 files in production.

tests/test_job_management.py
burnout87: I indirectly do it by inspecting the content of the returned object.

What would you like to see?

cdci_data_analysis/flask_app/app.py
burnout87: but then we should include also what inspect-state already does (email_history etc etc)...

burnout87: included now, and json schema adapted

burnout87: to reduce the number of opened files, I added also the arg `exclude_analysis_parameters` that if set, would exclude the content from the `analysis_parameters.json` file

by adding this the `token_expired` won't be present as it cannot be evaluated

tests/test_job_management.py
burnout87: > if the output of inspect-state is to be included within the output of inspect-jobs, than also the following files are open:
> 
>     * `session.log`

with the args `exclude_analysis_parameters` the content of the `analysis_parameters.json` file can be excluded thusr educing the number of files opened

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -481,89 +473,165 @@ def inspect_state(app):
...
+        result_job_status['scratch_dir_content'], result_job_status['request_completed'] = (
+            InstrumentQueryBackEnd.read_content_scratch_dir(scratch_dir,
+                                                            include_session_log=include_session_log,
+                                                            include_status_query_output=include_status_query_output,
+                                                            exclude_analysis_parameters=exclude_analysis_parameters))
+
+        if (not exclude_analysis_parameters and
+                (isinstance(result_job_status['scratch_dir_content']['analysis_parameters'], dict) and
+                 'token' in result_job_status['scratch_dir_content']['analysis_parameters'])):
+            # TODO I am not 100% sure this is enough, perhaps it's not even needed
```
volodymyrss, line 551: I am not sure what do you mean?
We need a check for expired token. This might be ok here.

volodymyrss (approved):

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: I wasn't sure whether checking against the current time would be enough, and instead using another "time", like the time of the last folder access.

I did try that it wasn't really reliable

That's why I was asking if you had some better suggestions

volodymyrss: We want to know if there is a chance the job can still finish, as of at the time of the request.
So current time is the right one.

PR #640, opened 2024-01-03
StructuredParameter and PhosphorosTableParameter

dsavchenko:


burnout87 (approved):

cdci_data_analysis/analysis/parameters.py
```diff
@@ -964,4 +968,76 @@ def value(self, v):
...
+        except (AssertionError, ValidationError):
+            raise RequestNotUnderstood('Wrong value of structured parameter %s', self.name)
+            
+
+class PhosphorosFiltersTable(StructuredParameter):
+    owl_uris = ('http://odahub.io/ontology#PhosphorosFiltersTable')
+    
+    def __init__(self, value=None, name=None):
+        
+        # TODO: this list should come from outside
```
volodymyrss, line 1007: from where for example?

cdci_data_analysis/analysis/parameters.py
```diff
@@ -964,4 +968,76 @@ def value(self, v):
...
+                       "GALEX|GALEX.FUV", "GALEX|GALEX.NUV",
+                       "LSST|LSST.g", "LSST|LSST.i", "LSST|LSST.r", "LSST|LSST.u", "LSST|LSST.y", "LSST|LSST.z",
+                       "PAN-STARRS|PS1.g", "PAN-STARRS|PS1.i", "PAN-STARRS|PS1.open", "PAN-STARRS|PS1.r", "PAN-STARRS|PS1.w", "PAN-STARRS|PS1.y", "PAN-STARRS|PS1.z",
+                       "VIRCAM|VISTA.H", "VIRCAM|VISTA.J", "VIRCAM|VISTA.Ks", "VIRCAM|VISTA.NB118", "VIRCAM|VISTA.NB980", "VIRCAM|VISTA.NB990", "VIRCAM|VISTA.Y", "VIRCAM|VISTA.Z",
+                       "SLOAN|SDSS.g", "SLOAN|SDSS.i", "SLOAN|SDSS.r", "SLOAN|SDSS.u", "SLOAN|SDSS.z",
+                       "Subaru|HSC.Y", "Subaru|HSC.g", "Subaru|HSC.i", "Subaru|HSC.r", "Subaru|HSC.z",
+                       "UKIRT|WFCAM.H", "UKIRT|WFCAM.J", "UKIRT|WFCAM.K", "UKIRT|WFCAM.Y", "UKIRT|WFCAM.Z",
+                       "WISE|WISE.W1", "WISE|WISE.W2", "WISE|WISE.W3", "WISE|WISE.W4"]
+        
+        schema = {"type": "object",
```
volodymyrss, line 1021: ultimately we could even load this schema from a JSON document somewhere, purely based on the URI.
If there is no additional check, this would allow to avoid even having the class.
But it's ok at this stage in the code like that I think.

volodymyrss (dismissed):
Thanks, cool!

Couple of comments, maybe they can be adapted in code comments for the future.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -461,7 +461,12 @@ def reprJSONifiable(self):
         if restrictions:
             reprjson[0]['restrictions'] = restrictions
         if getattr(self, 'owl_uris', None):
-            reprjson[0]['owl_uri'] = self.owl_uris
+            if isinstance(self.owl_uris, str):
+                reprjson[0]['owl_uri'] = [ self.owl_uris ]
```
burnout87, line 465: in case it's a string, how is it going to be formatted ?as a comma-separated list of uris?

perhaps it's safe to `split(',')` it?

dsavchenko: In this case, it's just one URI. This is not intentional, but we already have a situation where `owl_uris` attribute is either tuple or string. 
E.g. here https://github.com/oda-hub/dispatcher-app/blob/420e273c92261c3d112a1db9e2b599b84768247f/cdci_data_analysis/analysis/parameters.py#L559 it's meant to be a tuple, but it's actually a string, because of a missing comma. But everything works already anyway, and moreover, most parameters will have a single URI.

Here I just normalize it to always be a list in the metadata to simplify further checks.

burnout87 (approved):

volodymyrss (approved):

Issue #551, opened 2023-07-11
how to deal with generic file upload


volodymyrss:

volodymyrss:
this is also relevant for https://github.com/oda-hub/mmoda-frontend-module/issues/197

volodymyrss:
@burnout87 this is important now

volodymyrss:
this become more relevant now

PR #649, opened 2024-01-23
Further adaptations to structured input

dsavchenko:


dsavchenko:
Some notes:
1. The value may be encoded all the way through dispatcher, then the required change is only in the Parameter classes to decode before validation. But this makes consistent job_id calculation harder.
2. For job_id calculation, even for decoded value, the row ordering may matter. In the case of phosphoros table this is undesirable, but could be OK in general case. 
3. When requesting backend in plugin's `run_query` the parameter needs to be once again encoded.

dsavchenko:
@volodymyrss

dsavchenko:
We agree to postpone the implementation of fully consistent job_id calculation https://github.com/oda-hub/dispatcher-app/issues/650
As for now, this is very unlikely to cause problems

tests/test_parameters.py
```diff
@@ -647,4 +648,12 @@ def test_invalid_phosphoros_table(tab_value):
...
         PhosphorosFiltersTable(tab_value)
     
+
+def test_structured_get_default_value():
+    value = {"c": [15, 5], "a": [1, 2], "b": ["foo", "bar"]}
+    expected = '{"a": [1, 2], "b": ["foo", "bar"], "c": [15, 5]}'
+    
+    stp = StructuredParameter(value)
+    
+    assert stp.get_default_value() == expected
```
burnout87, line 658: I'd add a test where the value is a `dict`

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -828,8 +828,20 @@ def set_args(self, request, verbose=False):
         else:
             raise NotImplementedError
 
-        self.par_dic = args.to_dict()
-
+        self.par_dic = {}
+        for k, v in args.to_dict().items():
+            if k in ['catalog_selected_objects', 'selected_catalog']:
```
burnout87, line 833: this will change again in the future with #551

burnout87 (changes requested):

tests/test_parameters.py
dsavchenko: Sorry, I don't understand what you mean. It's already a dict...

burnout87: true, I wanted to refer to a test for the new approach to populate `self.par_dic`, which checks if the value is a list or a dict, can we add a test for that ?

cdci_data_analysis/flask_app/dispatcher_query.py
dsavchenko: True, it may change. But it's not always related with file upload. The catalogue may also be added as JSON-encoded string (this happens if the user selects some lines in the frontend "catalog" window).
I decided to completely mask these parameters so that changes are minimal and garanteed to be backward-compatible

tests/test_parameters.py
dsavchenko: The overall flow is tested in plugin, https://github.com/oda-hub/dispatcher-plugin-nb2workflow/blob/d0228f8f1e43da762785479b0ca16b924bf27561/tests/test_plugin.py#L554
It's not completely clear how to write a more localized test here. 

If we want to test only `InstrumentQueryBackEnd.set_args()` method, do you know how to prepare the `request` parameter as an instance of `flask.Request()`?

burnout87: what about a dummy instrument with a product that takes the new parameter as parameter a `StructuredParameter` ?

send a dummy request and then and check the content of the returned object ?

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -828,8 +828,20 @@ def set_args(self, request, verbose=False):
...
+                self.par_dic[k] = v
+                continue
+            try:
+                decoded = json.loads(v)
```
volodymyrss, line 837: it's a bit dangerous to rely on json parsing exception to determine if it should be parsed.

I guess it's hard to distinguish the intended type in here?

dsavchenko: `args` comes here from request, it's an ordered dict of strings. Need to parse them anyway to distinguish types.

If argument is actually a string, it's not a valid json because it isn't enclosed by `"` and we get the parsing exception.
If it's numeral, will parse successfully as int or float, not an object, then keep the previous behaviour.
If an object, decode.

We don't expect any other input here, but just in case, it won't parse, so keep previous behavior again.

volodymyrss: Right. I am concerned that if the json is somehow malformed where we expect it, we end up with a string which will lead later to possibly incomprehensive error. Maybe you can add a test for treating a failure like that?

dsavchenko: Hm, without any schema, the StructuredParameter accepts the string, may it be malformed json or anything. 
Which could be OK. Or we can add a restriction there, I'm not sure...

dsavchenko: So now, by default, StructuredParameter is required to really be structured. 
Added test with malformed structure

burnout87 (approved):

volodymyrss (approved):

PR #660, opened 2024-02-14
check presence `query_output.json` file and more comprehensive sentry message

burnout87:


cdci_data_analysis/flask_app/app.py
```diff
@@ -328,9 +328,17 @@ def push_renku_branch():
...
-
+            query_output_json_content_original = None
+            for scratch_folder in list_scratch_folders:
+                query_output_path = os.path.join(scratch_folder, 'query_output.json')
+                if os.path.exists(query_output_path):
+                    with open(query_output_path) as q_out_f:
+                        query_output_json_content_original = json.load(q_out_f)
+                    break
+            if query_output_json_content_original is None:
+                error_message = (f"Error while posting data in the renku branch: query_output.json file was not found "
```
volodymyrss, line 339: We do not want to send this to the user, this is internal error and user would not know what to do about it.

In this case dispatcher understands the request but can not fulfill it, right?

If so, user should get "internal error, team notified and we are working on it". 
And we should think right away why this is the case.
Do you have any idea what could be the reason?

volodymyrss (changes requested):

cdci_data_analysis/flask_app/app.py
burnout87: > We do not want to send this to the user, this is internal error and user would not know what to do about it.
> 
> In this case dispatcher understands the request but can not fulfill it, right?
> 

It is indeed not sent to the user, rather to sentry, the user, on the frontend will only see the `error_message` defined here: 

https://github.com/oda-hub/dispatcher-app/blob/7a941ade09de12c595c11f984bc9d49f89ae4a88/cdci_data_analysis/flask_app/app.py#L381-L395

> If so, user should get "internal error, team notified and we are working on it". And we should think right away why this is the case. Do you have any idea what could be the reason?

It's quite difficult to understand why it happened, I dont recall any reason why the dispatcher would delete that file specifically

volodymyrss: Ok, it's probably because they are big and cleaned up [here](https://github.com/oda-hub/dispatcher-chart/blob/bf21586379c5a52ba3531083d69e7f4f0181439b/templates/deployment.yaml#L155).

Then let's improve the message here, and in case if this file is gone, say that please resubmit the request.

Also, we could adjust the cleanup so that the latest version of this file remains (there are many I think, which is one of the reasons they are big).

But, do we really need these query output here?

burnout87: > 
> But, do we really need these query output here?

It's needed to extract the `prod_dict` which contains the `api_code` and the `analysis_parameters`  to collect all the info for the commit on the repository

PR #661, opened 2024-02-21
Optimisations in Instrument lists

dsavchenko:
This PR is a supporting for https://github.com/oda-hub/dispatcher-plugin-nb2workflow/pull/90

 - `importer.instrument_factory_list` is now an iterator that stores instr_factory_list per-plugin (to properly deal with the new one in nb2w plugin).
- `Parameter.from_owl_url` now accepts `Ontology` object (for optimisation)
- Optimisation. In all iterations over `instrument_factory_list` the `Instrument` object is not built if we don't really need it, instead attributes of the factory function are checked. (Before, all Instrument instances of nb2w plugin were created in every request).

### DIFF
--- a/cdci_data_analysis/analysis/parameters.py
+++ b/cdci_data_analysis/analysis/parameters.py
@@ -30,6 +30,7 @@ __author__ = "Andrea Tramacere"
 import six
 import decorator
 import logging
+import os
 
 from astropy.time import Time as astropyTime
 from astropy.time import TimeDelta as astropyTimeDelta
@@ -495,21 +496,27 @@ class Parameter:
                                         **kwargs):
         from oda_api.ontology_helper import Ontology
 
-        if ontology_path:
-            onto = Ontology(ontology_path)
+        if ontology_path is None:
+            logger.warning('Ontology path not set in Parameter.from_owl_uri(). '
+                'Trying to find parameter which have %s directly set. '
+                'extra_ttl will be ignored ', owl_uri)
+            parameter_hierarchy = [ owl_uri ]
+            par_format = par_unit = allowed_values = min_value = max_value = None
+        else:
+            if isinstance(ontology_path, str) or isinstance(ontology_path, os.PathLike):
+                onto = Ontology(ontology_path)
+            elif isinstance(ontology_path, Ontology):
+                onto = ontology_path
+            else:
+                raise RuntimeError("Wrong ontology_path")
+            
             if extra_ttl is not None:
                 onto.parse_extra_triples(extra_ttl)
             parameter_hierarchy = onto.get_parameter_hierarchy(owl_uri)
             par_format = onto.get_parameter_format(owl_uri)
             par_unit = onto.get_parameter_unit(owl_uri)
             min_value, max_value = onto.get_limits(owl_uri)
-            allowed_values = onto.get_allowed_values(owl_uri)            
-        else:
-            logger.warning('Ontology path not set in Parameter.from_owl_uri(). '
-                           'Trying to find parameter which have %s directly set. '
-                           'extra_ttl will be ignored ', owl_uri)
-            parameter_hierarchy = [ owl_uri ]
-            par_format = par_unit = allowed_values = min_value = max_value = None
+            allowed_values = onto.get_allowed_values(owl_uri)
         
         for owl_superclass_uri in parameter_hierarchy:
             for python_subclass in subclasses_recursive(cls):
--- a/cdci_data_analysis/analysis/parameters.py
+++ b/cdci_data_analysis/analysis/parameters.py
@@ -496,7 +496,7 @@ class Parameter:
             par_format = par_unit = allowed_values = min_value = max_value = None
         else:
             if ontology_path is not None:
-                if isinstance(ontology_path, str) or isinstance(ontology_path, os.PathLike):
+                if isinstance(ontology_path, (str, os.PathLike)):
                     onto = Ontology(ontology_path)
                 else:
                     raise RuntimeError("Wrong ontology_path")
--- a/cdci_data_analysis/analysis/parameters.py
+++ b/cdci_data_analysis/analysis/parameters.py
@@ -483,17 +483,6 @@ class Parameter:
                      extra_ttl = None,
                      ontology_path = None, 
                      **kwargs):
-        return cls.instance_signature_from_owl_uri(owl_uri,
-                                                   extra_ttl,
-                                                   ontology_path, 
-                                                   **kwargs)[0]
-
-    @classmethod
-    def instance_signature_from_owl_uri(cls,
-                                        owl_uri,
-                                        extra_ttl = None,
-                                        ontology_path = None, 
-                                        **kwargs):
         from oda_api.ontology_helper import Ontology
 
         if ontology_path is None:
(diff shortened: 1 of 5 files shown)


cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -1490,15 +1497,29 @@ def set_instrument(self, instrument_name, roles, email):
...
+                    instr_name = instrument_factory.instr_name
+                else:
+                    _instrument = instrument_factory()
+                    instr_name = _instrument.name
+                
+                if instr_name == instrument_name:
+                    if _instrument is None and hasattr(instrument_factory, 'instrument_query'):
+                        instr_query = instrument_factory.instrument_query
+                    else:
+                        _instrument = instrument_factory()
```
burnout87, line 1512: should this go one line up ?

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -1490,15 +1497,29 @@ def set_instrument(self, instrument_name, roles, email):
...
+                        instr_query = _instrument.instrumet_query
```
burnout87, line 1513:
```suggestion
                        instr_query = _instrument.instrument_query
```

burnout87 (changes requested):

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: I this I am not understand why the `_instrument = instrument_factory()` is executed in case `_instrument is not None`

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -1490,15 +1497,29 @@ def set_instrument(self, instrument_name, roles, email):
         if instrument_name == 'mock':
             new_instrument = 'mock'
```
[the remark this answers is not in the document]
burnout87: totally agree

cdci_data_analysis/plugins/importer.py
```diff
@@ -56,30 +57,30 @@
...
+    instr_factory_iter = InstrumentFactoryIterator()
     
     for plugin_name in cdci_plugins_dict:
         if activate_plugins == 'auto' or plugin_name in activate_plugins:   
             logger.info("found plugin: %s", plugin_name)
 
             try:
                 e = importlib.import_module('.exposer', cdci_plugins_dict[plugin_name].__name__)
-                instr_factory_list.extend(e.instr_factory_list)
+                instr_factory_iter.extend(e.instr_factory_list)
```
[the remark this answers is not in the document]
dsavchenko: This is what we want here, `InstrumentFactoryIterator` keeps a list of per-plugin `instr_factory_list`s and iterates them one-after-another. The idea here is that if they are actually lists, nothing will change. But in nb2workflow plugin it's not a list, but another iter, which reads instruments from the KG and updates itself before being iterated

cdci_data_analysis/flask_app/dispatcher_query.py
dsavchenko: Yep, something wrong here, thanks!

dsavchenko: Fixed. Variants are:

- Old plugins: factory functuin has no `instr_name`, `instrument_query` attributes. Then Instrument is created in the first else clause, regardless of that name is the one we request, getting `name` and `check_instrument_access` from the instance, as before.
- Both `instr_name` and `instrument_query` are attached to factory function. Instrument instance will be created only if one requests it and has access to it. Otherwise only list of names `known_instruments` is needed.
- Only `instr_name` factory attribute is available. Then the instance will be created to check access if one requests this particular instrument. This behaviour is not currently needed, just in case for the future (e.g. migrating plugins with non-trivial InstrumentQuery).

cdci_data_analysis/analysis/parameters.py
```diff
@@ -481,24 +482,37 @@ def from_owl_uri(cls,
...
+            raise RuntimeError("Both ontology_path and ontology_object parameters are set.")
+        elif ontology_path is None and ontology_object is None:
+            logger.warning('Ontology path/object not set in Parameter.from_owl_uri(). '
+                'Trying to find parameter which have %s directly set. '
+                'extra_ttl will be ignored ', owl_uri)
+            parameter_hierarchy = [ owl_uri ]
+            par_format = par_unit = allowed_values = min_value = max_value = None
+        else:
+            if ontology_path is not None:
+                if isinstance(ontology_path, str) or isinstance(ontology_path, os.PathLike):
```
volodymyrss, line 499:
```suggestion
                if isinstance(ontology_path, (str, os.PathLike)):
```

cdci_data_analysis/flask_app/app.py
```diff
@@ -207,8 +207,8 @@ def common_exception_payload():
      
     _l = []
 
-    for instrument_factory in importer.instrument_factory_list:
-        _l.append('%s' % instrument_factory().name)
+    for instrument_factory in importer.instrument_factory_iter:
+        _l.append('%s' % getattr(instrument_factory, 'instr_name', instrument_factory().name))
```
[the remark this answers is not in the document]
volodymyrss: Would this be the same?

```suggestion
    for instrument_factory in importer.instrument_factory_iter:
        _l.append(str(getattr(instrument_factory, 'instr_name', instrument_factory().name)))
```

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -446,11 +446,18 @@ def get_user_specific_instrument_list(app):
...
         out_instrument_list = []
-        for instrument_factory in importer.instrument_factory_list:
-            instrument = instrument_factory()
+        for instrument_factory in importer.instrument_factory_iter:
+            if hasattr(instrument_factory, 'instrument_query'):
+                instrument_query = instrument_factory.instrument_query
+                instr_name = getattr(instrument_factory, 'instr_name', instrument_factory().name)
+            else:
+                instrument = instrument_factory()
+                instrument_query = instrument.instrumet_query
```
volodymyrss, line 455: The `n` is still occasionally missing, but not everywhere.

```suggestion
                instrument_query = instrument.instrument_query
```

dsavchenko: That's because it's missed in `Instrument` class. This should be eventually changed, but let's do it separately. It also affects all the plugins.
I already replied to @burnout87 https://github.com/oda-hub/dispatcher-app/pull/661#discussion_r1502361312

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -446,11 +446,18 @@ def get_user_specific_instrument_list(app):
                 email = tokenHelper.get_token_user_email_address(decoded_token)
 
         out_instrument_list = []
-        for instrument_factory in importer.instrument_factory_list:
-            instrument = instrument_factory()
+        for instrument_factory in importer.instrument_factory_iter:
+            if hasattr(instrument_factory, 'instrument_query'):
```
volodymyrss, line 450: What makes this condition necessary?

volodymyrss: Are there different factories?

volodymyrss (approved):

cdci_data_analysis/flask_app/dispatcher_query.py
dsavchenko: > Are there different factories?

The factories in "old" plugins don't have these attributes attached

burnout87 (approved):

PR #663, opened 2024-02-27
Actually optimise instrument list iteration

dsavchenko:
This was rather dumb mistake, so the idea of optimisation of instruments iteration didn't really work...

### DIFF
--- a/cdci_data_analysis/flask_app/dispatcher_query.py
+++ b/cdci_data_analysis/flask_app/dispatcher_query.py
@@ -60,6 +60,7 @@ from .logstash import logstash_message
 
 from oda_api.data_products import NumpyDataProduct
 import oda_api
+from cdci_data_analysis.timer import block_timer
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


volodymyrss:
Maybe we could add a timer to notice how long it takes?

volodymyrss (dismissed):

dsavchenko:
It shouldn't take too long with optimisation and will also decrease and stabilise when parameter lists are cached. But this logging makes sense, will add it. 

This problem arises only with nb2workflow plugin. So probably better (and easier, there is only one place) to add a timer there.
Does sentry of dispatcher accessible from plugins? I'm just not sure

burnout87 (approved):

volodymyrss (approved):

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -445,19 +446,24 @@ def get_user_specific_instrument_list(app):
...
-        for instrument_factory in importer.instrument_factory_iter:
-            if hasattr(instrument_factory, 'instrument_query'):
-                instrument_query = instrument_factory.instrument_query
-                instr_name = getattr(instrument_factory, 'instr_name', instrument_factory().name)
-            else:
-                instrument = instrument_factory()
-                instrument_query = instrument.instrumet_query
-                instr_name = instrument.name
-                
+        with block_timer(logger=logger,
```
burnout87, line 449: no sentry?

burnout87 (approved):

cdci_data_analysis/flask_app/dispatcher_query.py
dsavchenko: It's [being set by default](https://github.com/oda-hub/dispatcher-app/blob/d841260bcb3b9208440244facc308d83f93e7180/cdci_data_analysis/timer.py#L7) to the dispatcher's main sentry instance

PR #664, opened 2024-02-27
Fix parameter signature check

dsavchenko:
Allow any keyword parameter when instantiating Parameter with **kwargs in from_owl_url

### DIFF
--- a/cdci_data_analysis/analysis/parameters.py
+++ b/cdci_data_analysis/analysis/parameters.py
@@ -39,7 +39,7 @@ from astropy import units as apy_u
 import numpy as np
 
 from typing import Union
-from inspect import signature
+from inspect import signature, Parameter
 from .exceptions import RequestNotUnderstood
 
 from jsonschema import validate, ValidationError, SchemaError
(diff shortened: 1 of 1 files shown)


volodymyrss (approved):

volodymyrss (dismissed):

cdci_data_analysis/analysis/parameters.py
```diff
@@ -520,7 +521,8 @@ def from_owl_uri(cls,
                 if python_subclass.matches_owl_uri(owl_superclass_uri):
                     logger.info("will construct %s by owl_uri %s", python_subclass, owl_superclass_uri)
                     call_kwargs = {}
-                    call_signature = signature(python_subclass)                   
+                    call_signature = signature(python_subclass)
+                    var_kw_signature = call_parameter.VAR_KEYWORD in [x.kind for x in call_signature]
```
burnout87, line 525: can you actually loop over the signature object?

According to https://docs.python.org/3.8/library/inspect.html#inspect.signature

you have to use `call_signatures.parameters.values()`

```suggestion
                    var_kw_signature = call_parameter.VAR_KEYWORD in [x.kind for x in call_signature.parameters.values()]
```

burnout87 (changes requested):

cdci_data_analysis/analysis/parameters.py
dsavchenko: You are right, thank you. It's dict-like. 
I wrote it a bit in a hurry, will fix

burnout87 (approved):

volodymyrss (approved):

PR #666, opened 2024-03-07
Generic file upload

burnout87:


burnout87:
* the method `upload_files_request` within the instrument class now takes care of uploading the files submitted with the request within the scratch_dir, using the name of the field, then calculating a hash of the file itself and adding it to the dict that is then used for the calculation of the `job_id`
* the endpoint  `download_file` has been and it can be used as a url to get the specific file when needed within the backend - it works like `download_products`, with the difference that it is intended to download a single file whereas a whole product

cdci_data_analysis/analysis/hash.py
```diff
@@ -34,3 +34,13 @@ def format_hash(x): return hashlib.md5(
 
     # this takes care of various strange objects which can not be properly represented
     return format_hash(json.dumps(o))
+
+
+def make_hash_file(file_path):
+
+    # TODO to adapt in case of big files and reading in chunks will be necessary
+    def read_file():
```
volodymyrss, line 42: why do you need a function here, you just call it once right below?

cdci_data_analysis/analysis/hash.py
```diff
@@ -34,3 +34,13 @@ def format_hash(x): return hashlib.md5(
...
+        with open(file_path, 'rb') as f_fn:
```
volodymyrss, line 43: just for clarify, why is it called `f_fn`? This is not a filename (`fn`?).

burnout87:
With https://github.com/oda-hub/dispatcher-plugin-nb2workflow/pull/94 the download link will be built, then the nb2service will make the download according to the logic implemented with https://github.com/oda-hub/nb2workflow/pull/141

cdci_data_analysis/analysis/hash.py
```diff
@@ -34,3 +34,12 @@ def format_hash(x): return hashlib.md5(
 
     # this takes care of various strange objects which can not be properly represented
     return format_hash(json.dumps(o))
+
+
+def make_hash_file(file_path):
+
+    # TODO to adapt in case of big files and reading in chunks will be necessary
```
volodymyrss, line 41: I'd just run `md5sum` with subprocess, it's likely to be faster and memory-preserving even for moderately big files.

cdci_data_analysis/analysis/instrument.py
```diff
@@ -716,6 +721,17 @@ def set_input_products(self, par_dic, input_file_path,input_prod_list_name):
...
+    def upload_files_request(self, par_dic, request, temp_dir):
+        if request.method == 'POST':
+            for f in request.files:
+                # TODO needed since those two files are extracted in a previous step
+                if f != 'user_scw_list_file' and f != 'user_catalog_file':
+                    f_path = upload_file(f, temp_dir)
+                    if f_path is not None:
+                        f_hash = make_hash_file(f_path)
+                        par_dic[f] = f_hash
+
```
volodymyrss, line 733:
```suggestion
                        par_dic[f] = make_hash_file(f_path)

```

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -1047,18 +1056,21 @@ def update_token(self, update_email_options=False, refresh_token=False):
...
         try:
             # TODO not entirely sure about these
             self.off_line = False
             self.api = False
 
             self.validate_job_id(request_parameters_from_scratch_dir=True)
             file_list = self.args.get('file_list').split(',')
-            file_name = self.args.get('download_file_name')
+            file_name = self.args.get('download_file_name', None)
+            if file_name is None:
```
volodymyrss, line 1068: is this correctly preventing download of anything outside the product directory?

burnout87: This arg was intended to be used as the name of the archive that will contain all the file of the product.

in the case of wanting to download a single file, with the `download_file` endpoint, the name of the downloaded file will match with the desired file that is the arg `file_list`, and it won't create an archive if wished (using the `return_archive` arg)

burnout87: To better answer your question: the path of the file to download is built here:

https://github.com/oda-hub/dispatcher-app/blob/617110283afddc6e862f0156c5a32d9af3cae62b/cdci_data_analysis/flask_app/dispatcher_query.py#L924-L925

So, do we want to check that each single file is present inside the scratch_dir first ?

dsavchenko: > is built here

Just a side note. There is no need in `+ '/'` here

burnout87: no, you're right. but this was the first time I was looking at that function...I will adjust it

dsavchenko: > is this correctly preventing download of anything outside the product directory?

No, I just checked. We need to fix it for the dispatcher in general

cdci_data_analysis/analysis/hash.py
burnout87: it now opens in a subprocess

cdci_data_analysis/analysis/hash.py
```diff
@@ -34,3 +35,11 @@
 
     # this takes care of various strange objects which can not be properly represented
     return format_hash(json.dumps(o))
+
+
+def make_hash_file(file_path):
+    command = f"md5sum {file_path}"
```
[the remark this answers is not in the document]
dsavchenko: @burnout87 you can avoid this warning by passing command as a list and `shell=False`

cdci_data_analysis/flask_app/dispatcher_query.py
dsavchenko: https://github.com/oda-hub/dispatcher-app/pull/667

cdci_data_analysis/analysis/parameters.py
```diff
@@ -583,6 +583,12 @@ def check_name_value(value, units=None, name=None, par_format=None):
 class Name(String):
     owl_uris = String.owl_uris + ("http://odahub.io/ontology#AstrophysicalObject",)
 
+class FileReference(String):
+    owl_uris = String.owl_uris + ("http://odahub.io/ontology#FileReference",)
```
dsavchenko, line 587: I think `String.owl_uris` is not needed here (and it changes nothing as far as I remember, because of the order the owl class hierarchy is matched with the python class hierarchy)

burnout87: ok, so can it be removed also for the `Name` class?

cdci_data_analysis/analysis/parameters.py
```diff
@@ -583,6 +583,18 @@ def check_name_value(value, units=None, name=None, par_format=None):
...
     owl_uris = String.owl_uris + ("http://odahub.io/ontology#AstrophysicalObject",)
 
+class FileReference(String):
+    owl_uris = "http://odahub.io/ontology#FileReference"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def get_default_value(self):
+        return {'value': self.value, 'downloadable': True}
```
dsavchenko, line 593: You can make this configurable for the future as a parameter (with default True) in the `__init__`.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -583,6 +583,18 @@ def check_name_value(value, units=None, name=None, par_format=None):
...
+
+class POSIXPath(FileReference):
+    owl_uris = FileReference.owl_uris + "http://odahub.io/ontology#POSIXPath"
+
```
dsavchenko: Not sure, which class must override the `get_default_value()` here. 
This depends on the overall interface. Do we always susbstitute the `POSIXPath` with downloadable file? For the moment, yes.  And what do we do with `FileURL` parameter (not really used yet, but annotation exist)? Do we treat it the same way by uploading from frontend? Or we just pass it as an external URL? 
If both are always "downloadable" and are exposed to the frontend by dispatcher - one `FileReference` class will support both. 
We may discuss a bit in the call

cdci_data_analysis/analysis/parameters.py
```diff
@@ -586,11 +586,12 @@ class Name(String):
 class FileReference(String):
     owl_uris = "http://odahub.io/ontology#FileReference"
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, downloadable=True, *args, **kwargs):
```
dsavchenko, line 589: Syntax error here. Named parameter before positional

burnout87: right

cdci_data_analysis/analysis/parameters.py
burnout87: > Do we always susbstitute the `POSIXPath` with downloadable file? For the moment, yes. 

Only in the case this is a URL

> And what do we do with `FileURL` parameter (not really used yet, but annotation exist)? Do we treat it the same way by uploading from frontend? 

Is it intended to be a URL provided from the frontend? Will it refer to a File? In that case it might not be needed to be downloadable, so `downloadable=False`

> If both are always "downloadable" and are exposed to the frontend by dispatcher - one `FileReference` class will support both.

With both you mean PosixPAth and FileURL ? Are they always ? I guess it is up to us to decide, but in case they are, one class will suffice

dsavchenko: Right, that's for us to decide how to provide these urls to the backend (which effectively accepts urls for both). 
Generally speaking, the way frontend + dispatcher deal with files might not be fully determined by the annotations. I could imagine a frontend element which allows to either upload file or provide url, then dispatcher will either expose uploaded file or just pass the url as-is, and it will work for both backend parameter types.
Unfortunately, we didn't really discuss it online. So @volodymyrss what do you think?

volodymyrss: Can we discuss it on a short meeting? Tomorrow 10h30?

burnout87: Ok for me

volodymyrss (commented):
During `__init__` we will store files in a separate directory, keeping track of the file owner in a metadata file.
At this point the entry in the `par_dict` will be replaced with a URL resolvable by the dispatcher.
The rest of the treatment here and in the plugin will be the same as for the case when user provides a URL.

burnout87:
> * During `__init__` we will store files in a separate directory, keeping track of the file owner in a metadata file. 

the separate folder is called `request_files` and inside there is a json file that for each file in the file lists the email addresses of those that can access to it, it can also be `public`

> * At this point the entry in the `par_dict` will be replaced with a URL resolvable by the dispatcher. 

the `products_url` extracted from the configuration is used to build the download url

dsavchenko:
> for each file in the file lists the email addresses of those that can access to it, it can also be `public`

We may probably use role-based access by default to match the way instruments are accessed. 
With user-based, no-one else is able to run the same request. User email may still be included in json, just in case, more meta-data is better. Creation timestamp could also be useful, for example. 

Also, is there any way to force the file to be public while being logged-in?

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -937,11 +1000,13 @@ def prepare_download(self, file_list, file_name, scratch_dir):
             file_list = [file_list]
 
         for ID, f in enumerate(file_list):
-            file_list[ID] = self.validated_download_file_path(scratch_dir, f)
+            file_list[ID] = self.validated_download_file_path(origin_dir, f)
+            if from_request_files_dir:
+                self.verify_access_to_file(f)
 
-        tmp_dir = tempfile.mkdtemp(prefix='download_', dir='./')
+        file_dir = tempfile.mkdtemp(prefix='download_', dir='./')
```
dsavchenko, line 1007: The separate download directory seems to be not needed, at least when `return_archive=False`. 
(But this is a minor note, we can keep it as it is)

dsavchenko (changes requested):

cdci_data_analysis/analysis/parameters.py
```diff
@@ -583,6 +583,16 @@ def check_name_value(value, units=None, name=None, par_format=None):
...
+class FileReference(String):
+    owl_uris = "http://odahub.io/ontology#FileReference"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+
+class POSIXPath(FileReference):
+    owl_uris = FileReference.owl_uris + "http://odahub.io/ontology#POSIXPath"
+
```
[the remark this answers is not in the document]
burnout87: no, could it be anyway useful for the future?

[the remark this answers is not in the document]
dsavchenko: I think we can remove it for now, there is nothing inside anyway

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: it is actually used in case when `return_archive=False and file_name != file_to_download`

cdci_data_analysis/analysis/instrument.py
```diff
@@ -716,6 +731,33 @@ def set_input_products(self, par_dic, input_file_path,input_prod_list_name):
...
+                        file_hash = make_hash_file(f_path)
+                        f_name, f_ext = os.path.splitext(os.path.basename(f_path))
+                        new_file_name = f_name.split('.')[0] + '_' + file_hash + f_ext
+                        new_file_path = os.path.join(upload_dir, new_file_name)
+                        list_uploaded_files.append(new_file_name)
+                        os.rename(f_path, new_file_path)
+                        dpars = urlencode(dict(file_list=new_file_name,
+                                               query_status="ready",
+                                               return_archive=False,
+                                               token=token))
```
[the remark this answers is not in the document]
burnout87: True, good point. Perhaps it can be added within the service nb2workflow ? I'd avoid it in the plugin...

cdci_data_analysis/analysis/instrument.py
```diff
@@ -716,6 +731,33 @@ def set_input_products(self, par_dic, input_file_path,input_prod_list_name):
...
+            else:
+                basepath = os.path.join(products_url, 'download_file')
+            for f in request.files:
+                # TODO needed since those two files are extracted in a previous step
+                if f != 'user_scw_list_file' and f != 'user_catalog_file':
+                    f_path = upload_file(f, upload_dir)
+                    if f_path is not None:
+                        file_hash = make_hash_file(f_path)
+                        f_name, f_ext = os.path.splitext(os.path.basename(f_path))
+                        new_file_name = f_name.split('.')[0] + '_' + file_hash + f_ext
```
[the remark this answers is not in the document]
burnout87: > Just keeping hash here is better. The request must result in the same `jobid` (and it will depend on the constructed url string, right?) for the same file regardless of its name. 

Yes, the file upload, which happens one line above, uses the name of the field in the request files, to which the hash will be added

>  The metatadata file can then be used to relate the hash to human-readable filename.

It is actually used to relate a file to the users that can access it

cdci_data_analysis/analysis/instrument.py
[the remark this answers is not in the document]
dsavchenko: In nb2workflow we don't currently have it. So appending it somewhere in the dispatcher could make sense. 

On the other hand, it will also affect the job identifier on the nb2workflow side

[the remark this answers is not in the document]
burnout87: > In nb2workflow we don't currently have it. So appending it somewhere in the dispatcher could make sense.

like after the job_id generation ? an option we have already discussed

> 
> On the other hand, it will also affect the job identifier on the nb2workflow side

in which way?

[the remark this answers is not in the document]
burnout87: The url is now built without the token, rather with a placeholder, this will be substituted within the plugin (https://github.com/oda-hub/dispatcher-plugin-nb2workflow/pull/94) if needed.

This is for not impacting the `job_id` generation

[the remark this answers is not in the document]
dsavchenko: > in which way?

Imagine the token is changed (e.g. regenerated with the new lifetime). 
On the dispatcher side, the job_id is still the same, because the placeholder is used when calculating it.
But then the placeholder is substituted (in the plugin before requesting the backend), so nb2workflow backend will see this request as a new one (the job_id there is calculated separately) and will start a new job.

cdci_data_analysis/analysis/hash.py
```diff
@@ -34,3 +35,11 @@ def format_hash(x): return hashlib.md5(
 
     # this takes care of various strange objects which can not be properly represented
     return format_hash(json.dumps(o))
+
+
+def make_hash_file(file_path):
+    command = ["md5sum", file_path]
+    process = subprocess.Popen(command, shell=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
volodymyrss, line 42: can't you use `check_output`?

cdci_data_analysis/analysis/instrument.py
```diff
@@ -716,6 +731,32 @@ def set_input_products(self, par_dic, input_file_path,input_prod_list_name):
...
+                        new_file_path = os.path.join(upload_dir, new_file_name)
+                        list_uploaded_files.append(new_file_name)
+                        os.rename(f_path, new_file_path)
+                        dpars = urlencode(dict(file_list=new_file_name,
+                                               return_archive=False,
+                                               token="INSERT_YOUR_TOKEN_HERE"))
+                        download_file_url = f"{basepath}?{dpars}"
+                        par_dic[f] = download_file_url
+        return list_uploaded_files
+
```
volodymyrss, line 758: so this stores hash, not url, right? I think we discussed URL? So that it's universally valid.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -586,6 +586,12 @@ class LongString(String):
 class Name(String):
     owl_uris = String.owl_uris + ("http://odahub.io/ontology#AstrophysicalObject",)
 
+class FileReference(String):
+    owl_uris = "http://odahub.io/ontology#FileReference"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
 class NumericParameter(Parameter):
```
volodymyrss, line 595: didn't we discuss that this can cause issues when the value is stored and passed further?

cdci_data_analysis/flask_app/app.py
```diff
@@ -151,9 +151,14 @@ def meta_data_src():
 @app.route("/download_products", methods=['POST', 'GET'])
 def download_products():
     query = InstrumentQueryBackEnd(app, download_products=True)
-    return query.download_products()
+    return query.download_file()
 
 
+@app.route("/download_file", methods=['POST', 'GET'])
```
volodymyrss, line 157: could you please explain why this is needed?

cdci_data_analysis/analysis/instrument.py
burnout87: the `list_uploaded_files` list stores the filenames, to be then used to update the ownership json file, inside the directory maintaining the uploaded files

the urls are then part of the `par_dic` dict

cdci_data_analysis/analysis/instrument.py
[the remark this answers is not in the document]
burnout87: ok I wasn't aware of this re-calculation of the job_id
so you suggest to leave the placeholder here, and not making any token substitution on the plugin? then the nb2workflow backend will anyway have access to the token, and can itself make the substitution ?

cdci_data_analysis/analysis/parameters.py
burnout87: which kind of issue as I dont remember...

volodymyrss: That this dict will be included in analysis_paramters, API code, product URLs.
Better to immediately introduce a parameter with an (constructed) URL which is also always valid way to pass this parameter.

cdci_data_analysis/flask_app/app.py
burnout87: is your comment about code duplication with the `download_products` endpoint ?

cdci_data_analysis/analysis/instrument.py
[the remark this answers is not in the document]
burnout87: so, are you suggesting of not using the file_name, and just its hash, for storing the file itself?

cdci_data_analysis/analysis/parameters.py
burnout87: > That this dict will be included in analysis_paramters, API code, product URLs.
you mean the `kwargs` dict?

> Better to immediately introduce a parameter with an (constructed) URL which is also always valid way to pass this parameter.

this is what the `FileRefence` Parameter is for, perhaps it can be named as `ParameterURL`?

cdci_data_analysis/analysis/instrument.py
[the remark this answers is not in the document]
burnout87: btw where is the job_id calculation happening in the backend?

cdci_data_analysis/flask_app/app.py
burnout87: the behavior is slightly different, and mainly not to use the `download_products` endpoint when the intention is to download a single file (from the dedicated folder) instead of the the whole products for a given job (from the scratch_dir)

cdci_data_analysis/analysis/parameters.py
volodymyrss: Sorry I don't understand, what FileRefence is for?

http://odahub.io/ontology#FileReference already has a subclass with URL.

volodymyrss: If and when you represent the parameter as file URL, then FileURL seems appropriate.

burnout87: I introduced the FileURL parameter class which extends the FileReference, with the appropriate owl_uris

cdci_data_analysis/analysis/instrument.py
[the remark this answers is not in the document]
dsavchenko: > so you suggest to leave the placeholder here, and not making any token substitution on the plugin? then the nb2workflow backend will anyway have access to the token, and can itself make the substitution ?

I don't fully like it because we try to keep the backend nb2workflow standalone. Like this we bound some of its code to dispatcher. But I don't see any general approach

 
> btw where is the job_id calculation happening in the backend?

It's here https://github.com/oda-hub/nb2workflow/blob/d68c869f138087936cdb02125ba486a46f3834b1/nb2workflow/service.py#L285. It has slightly different implementation but the idea is the same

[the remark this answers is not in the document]
burnout87: ok, and can the token be removed before the generation of the key ? or is it meant to stay there...? we remove on the dispatcher (but they are two separate entities)

https://github.com/oda-hub/dispatcher-app/blob/8d519bf453c42443e22cb5dadcbf24a71e49d706/cdci_data_analysis/flask_app/dispatcher_query.py#L676-L702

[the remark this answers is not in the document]
dsavchenko: > ok, and can the token be removed before the generation of the key ?

It can, in principle. The one that proposed to pass the current token is an underscored argument `_token`. It's removed by default. This one is a part of other argument without fixed name. So the logic can be that we remove "token" arg from any argument that has a form of url before calculating id. Not sure, this won't produce any side effects

cdci_data_analysis/flask_app/app.py
burnout87: it redirects now, adding specific args

cdci_data_analysis/analysis/instrument.py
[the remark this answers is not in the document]
burnout87: as discussed, the token will be injected inside on the nb2workflow service, after the parsing of the url

dsavchenko (dismissed):

cdci_data_analysis/analysis/instrument.py
```diff
@@ -684,6 +700,17 @@ def set_input_products_from_fronted(self, input_file_path, par_dic, verbose=Fals
             else:
                 raise RuntimeError
 
+    def update_par_dic_with_uploaded_files(self, par_dic, uploaded_files_obj, products_url):
+        if validators.url(products_url):
+            basepath = os.path.join(products_url, 'dispatch-data/download_file')
+        else:
+            basepath = os.path.join(products_url, 'download_file')
```
volodymyrss, line 707: why is this needed?

cdci_data_analysis/flask_app/app.py
```diff
@@ -150,8 +150,30 @@ def meta_data_src():
...
+    from_request_files_dir = request.args.get('from_request_files_dir', 'False') == 'True'
+    download_file = request.args.get('download_file', 'False') == 'True'
+    download_products = request.args.get('download_products', 'True') == 'True'
+    query = InstrumentQueryBackEnd(app, download_products=download_products, download_files=download_file)
+    return query.download_file(from_request_files_dir=from_request_files_dir)
+
+
+@app.route("/download_file", methods=['POST', 'GET'])
+def download_file():
+    if app.config['conf'].products_url is not None and validators.url(app.config['conf'].products_url):
```
volodymyrss, line 162:
```suggestion
    if validators.url(app.config['conf'].products_url):
```

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -860,6 +871,50 @@ def set_args(self, request, verbose=False):
...
+        request_files_dir.mkdir()
+        ownership_file_path = os.path.join(request_files_dir.path, '.file_ownerships.json')
+        if not os.path.exists(ownership_file_path):
+            with open(ownership_file_path, 'w') as ownership_file:
+                json.dump({}, ownership_file)
+        return request_files_dir.path
+
+
+    def update_ownership_files(self, uploaded_files_obj):
+        if hasattr(self, 'decoded_token') and self.decoded_token is not None:
```
volodymyrss, line 885:
```suggestion
        if getattr(self, 'decoded_token', None) is not None:
```

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -860,6 +871,50 @@ def set_args(self, request, verbose=False):
...
+            user_email = tokenHelper.get_token_user_email_address(self.decoded_token)
+            user_roles = tokenHelper.get_token_roles(self.decoded_token)
+        else:
+            user_email = 'public'
+            user_roles = []
+
+        update_file = False
+        ownership_file_path = os.path.join(self.request_files_dir, '.file_ownerships.json')
```
volodymyrss, line 893: Why hide the file?

```suggestion
        ownership_file_path = os.path.join(self.request_files_dir, 'file_ownerships.json')
```

volodymyrss: This might not be safe if many updates happen for the same request. Can you make separate metdata file for each data file?

cdci_data_analysis/analysis/instrument.py
burnout87: Simply in case the products_url is a string during tests (ie `PRODUCTS_URL`) 

Same approach has been used for the `instr-list` endpoint

cdci_data_analysis/flask_app/app.py
burnout87: This fails in the case `products_url` is not present in the configuration, which is a case we considered when configuring the redirection for the `instr-list` endpoint.

Do we ever have cases where products_url is not in the configuration?

burnout87:
The test `test_parameter_format_from_owl_uri` is failing for a reason I don't understand, any ideas @dsavchenko  ?

dsavchenko:
> any ideas @dsavchenko ?

I think this is because of https://github.com/oda-hub/ontology/commit/d0aca6d7c9d9c1e70f4c3117b523d7e134933d66
It created a "shortcut" in the class hierarchy. Will need to fix it https://github.com/oda-hub/dispatcher-app/issues/678
(not sure how...)

dsavchenko:
https://github.com/oda-hub/ontology/pull/34 should fix this

dsavchenko (approved):
Please just check once again if we can safely ignore these CodeQL warnings. 
Otherwise, looks good!

burnout87:
> Please just check once again if we can safely ignore these CodeQL warnings. Otherwise, looks good!

We implemented the check on the inputs before actually downloading the file with #667

dsavchenko:
> We implemented the check on the inputs before actually downloading the file with #667

Yes, I remember, this was my PR. Probably, CodeQL isn't so smart, just to be sure that we didn't miss anything there

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: a metadata file per file is now created

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -937,29 +1010,38 @@
...
+                with open(file_list[0], "rb") as f_in:
+                    in_data = f_in.read()
+                with gzip.open(file_path, 'wb') as f:
+                    f.write(in_data)
+            else:
+                file_to_download = file_list[0].split('/')[-1]
+                if file_name == file_to_download:
+                    file_dir = os.path.dirname(file_list[0])
+                else:
+                    shutil.copy(file_list[0], file_path)
```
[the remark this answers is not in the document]
volodymyrss: is this valid @burnout87 ?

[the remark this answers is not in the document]
burnout87: It is pointing at the fact that the input would not be checked before using it (copying etc etc), but we implemented the check on the inputs before actually downloading the file with https://github.com/oda-hub/dispatcher-app/pull/667 .

Maybe CodQL does not see it

dsavchenko (approved):

volodymyrss (approved):
I do not have time to look closely, but I think we discussed it enough, lets see how it behaves.

PR #670, opened 2024-04-15
read and store return-progress output on a dedicated output file

burnout87:


burnout87:
With this, the idea is to write the outputs of the `return-progress` call, in dedicated files, namely:
* `query_output.json.return-progress-query-output` to store the output (eg html  notebook output)
* `query_output.json.return-progress-job-monitor` to store the resulting job_monitor object

In addition to that, the "main" job object won't be overwritten (that is the cause of the reset of the progress bar), and consequently the content of the `job_monitor.json` file. The job output of the return-progress call will be stored in the mentioned above.

burnout87:
tests on the nb2workflow plugin fail, but are addressed with https://github.com/oda-hub/dispatcher-plugin-nb2workflow/pull/100

volodymyrss:
Can it be instead that return_progress just adds some information to query output, not creating an entirely new output structure?

volodymyrss (commented):
Can it be instead that return_progress just adds some information to query output, not creating an entirely new output structure?

burnout87:
> With this, the idea is to write the outputs of the `return-progress` call, in dedicated files, namely:
> 
>     * `query_output.json.return-progress-query-output` to store the output (eg html  notebook output)
> 
>     * `query_output.json.return-progress-job-monitor` to store the resulting job_monitor object
> 
> 
> In addition to that, the "main" job object won't be overwritten (that is the cause of the reset of the progress bar), and consequently the content of the `job_monitor.json` file. The job output of the return-progress call will be stored in the mentioned above.

EDIT

The query output of 'return-progress' is not in a separate file, and the previous behavior has been re-established.

The only difference is that the job-monitor will not be overwritten, thus preventing the progress bar from being reset.

burnout87:
the nb2workflow plugin is now failing in test for the ontology, any ideas why ?

dsavchenko:
> the nb2workflow plugin is now failing in test for the ontology, any ideas why ?

Here is a test adaptation

volodymyrss (approved):

dsavchenko (approved):

PR #675, opened 2024-05-10
setting disp_conf in case of resubmission

burnout87:


burnout87:
integral tests fails because of issues in fetching dummy products

volodymyrss:
so you think this helps?

burnout87:
> so you think this helps?

yes I think so

volodymyrss:
Should we do something about the tests?

burnout87:
> Should we do something about the tests?

those that are failing now ? issue is in fetching the dummy products from your personal URL

dsavchenko (approved):

volodymyrss (approved):

PR #683, opened 2024-05-24
Support head download endpoint

burnout87:


burnout87:
> With [70fbada](https://github.com/oda-hub/dispatcher-app/commit/70fbada53c5daf82fd468531b50eebc4b3011bff) the `download_products` and `download_file` endpoints can be reached also with `dispatch-data/download_products` and `dispatch-data/download_file` respectively.
> 
> The need for this came from some issues I am encountering when doing some local testing. In particular, when running locally, the url for the download would contain `dispatch-data`, and those endpoints would not be originally reachable (which would be when in production though).
> 
> I am thinking that probably this change is not going to have an impact in the current platform since the original endpoint will be still reachable, yet I am wondering if this is too "hacky".
> 
> What do you think @volodymyrss @dsavchenko ?

after a discussion with @dsavchenko , we agreed on:
* adding a special arg to the download_file query built within the `update_par_dic_with_uploaded_files` function, and to adapt accordingly the nb2workflow, this will ensure a more consistent check for mmoda_url on the backend side
* use the bind_host and bind_port configuration in the case of a not url products_url

dsavchenko (dismissed):

burnout87:
now the nb2wporkflow and dispatcher-app use the same version of the library `validators`

dsavchenko (dismissed):

volodymyrss (approved):

PR #693, opened 2024-06-19
numpy, flask, jsonschema versionfreezing

burnout87:


dsavchenko:
- Could you briefly describe what problem does it fix?
- Would be better to have more loose restrictions in `setup.py` and freeze only in `requirements.txt` if possible
- The snyk test is failing (but I can't see the problem, it's only @volodymyrss who has access)
- So bokeh is not finally freezed?

burnout87:
> * Could you briefly describe what problem does it fix?

When using `python 3.10` (and above) it installs the version `2.0.0` of `numpy`, which causes the error seen [here](https://github.com/oda-hub/oda_api/actions/runs/9584155287/job/26427088848), a compatibility one between the version of bokeh (`2.4.2`, which is frozen) and the newer version of `numpy`. So, by freezing the version to `1.24.4`, this does not happen

> 
>     * So bokeh is not finally freezed?

It is actually already frozen, to the version `2.4.2`

burnout87:
also, some issues with the versioning of `flask` and `jsonschema` came out while I was testing,hence the freezing of those libraries too

volodymyrss:
> > * Could you briefly describe what problem does it fix?
> 
> When using `python 3.10` (and above) it installs the version `2.0.0` of `numpy`, which causes the error seen [here](https://github.com/oda-hub/oda_api/actions/runs/9584155287/job/26427088848), a compatibility one between the version of bokeh (`2.4.2`, which is frozen) and the newer version of `numpy`. So, by freezing the version to `1.24.4`, this does not happen

could be better to put `<2.0.0` constraint for numpy.

> 
> > ```
> > * So bokeh is not finally freezed?
> > ```
> 
> It is actually already frozen, to the version `2.4.2`

setup.py
```diff
@@ -49,7 +49,7 @@
     "giturlparse",
     "sentry-sdk",
     "validators==0.28.3",
-    "jsonschema"
+    "jsonschema==3.2.0"
```
volodymyrss, line 52: you are sure this can not be looser?

burnout87: I will try

burnout87: without the restriction, I get this error (using python 3.10):

```bash
ERROR: flask-restx 1.2.0 has requirement jsonschema<=4.17.3, but you'll have jsonschema 4.22.0 which is incompatible.
```
so, I apply this restriction

burnout87: it seems to be ok

burnout87:
I tried to change the restrictions on `flask` (given the warning from snyk) but I am not really successful...

burnout87:
> Ok, well, if that's the best that seems possible.

we'll certainly have to deal again with those dependencies in the future (eg with new versions of other libraries), then we'll see

dsavchenko (approved):

PR #704, opened 2024-08-01
Optional parameters

dsavchenko:


dsavchenko:
Tests in integral plugin fail on metadata requests, an exception occurs in oda_api https://github.com/oda-hub/oda_api/blob/02e47e3670542e763f381364c4a87de99a2a9de9/oda_api/api.py#L920-L924
This is due to a bit weird structure of the response, which is json list, but all queries are represented there by already json-encoded strings. Now, these strings contain `true/false` json representation of boolean, which don't eval as python expression. 

It may be fixed by adding other replaces (true->True), but I don't really understand why it's implemented like this anyway. Probably just legacy. 
@volodymyrss do you have an idea, probably there are additional reasons for this?

volodymyrss:
> Tests in integral plugin fail on metadata requests, an exception occurs in oda_api https://github.com/oda-hub/oda_api/blob/02e47e3670542e763f381364c4a87de99a2a9de9/oda_api/api.py#L920-L924 This is due to a bit weird structure of the response, which is json list, but all queries are represented there by already json-encoded strings. Now, these strings contain `true/false` json representation of boolean, which don't eval as python expression.
> 
> It may be fixed by adding other replaces (true->True), but I don't really understand why it's implemented like this anyway. Probably just legacy. @volodymyrss do you have an idea, probably there are additional reasons for this?

Looks like legacy code, should be adjusted when possible.

cdci_data_analysis/analysis/parameters.py
```diff
@@ -616,7 +621,7 @@ def __init__(self, *args, **kwargs):
             kwargs['allowed_types'] = [int, float]
         
         if kwargs.get('default_type') is None:
-            val = kwargs['value'] if kwargs.get('value') is not None else args[0]
+            val = kwargs['value'] if 'value' in kwargs.keys() else args[0]
```
burnout87, line 624:
```suggestion
            val = kwargs.get('value', args[0])
```

cdci_data_analysis/analysis/parameters.py
```diff
@@ -794,7 +818,13 @@ class TimeDelta(Time):
...
     format_kw = 'delta_T_format'
     
-    def __init__(self, value=None, delta_T_format='sec', name=None, delta_T_format_name=None, par_default_format='sec', extra_metadata=None):
+    def __init__(self, 
+                 value, 
+                 delta_T_format='sec', 
+                 name=None, 
+                 delta_T_format_name=None, 
+                 par_default_format='sec', 
+                 extra_metadata=None):
```
burnout87, line 827: is the `is_optional` arg to be added here?

burnout87 (changes requested):
Why certain classes within the `parameters` module are set as optional whereas others not?

cdci_data_analysis/analysis/parameters.py
dsavchenko: I didn't add it because this parameter class is deprecated. It's there to keep compatibility with plugins, but isn't used in nb2workflow plugin, so will never be declared optional.

dsavchenko:
> Why certain classes within the `parameters` module are set as optional whereas others not?

There are two parameters declared optional by default: `InputProdList` and `UserCatalog`. They are specific to Integral and their value may be set to None. 
Also, token may be None, so it's also "optional". 
All "normal" parameters are not optional by default.

burnout87 (approved):

volodymyrss (approved):

PR #705, opened 2024-08-15
properly handling post request and sanitize request values before logging

burnout87:
it turns out, the gallery-related endpoints are all only `POST` endpoint for the presence of files, so, adaptation is minimal

alongside with https://github.com/oda-hub/oda_api/pull/273

### DIFF
--- a/cdci_data_analysis/flask_app/app.py
+++ b/cdci_data_analysis/flask_app/app.py
@@ -451,9 +451,11 @@ def run_analysis():
     request_summary = log_run_query_request()
 
     try:
+        sanitized_request_values = sanitize_dict_before_log(request.values)
+
         logger.info('\033[32m===> dataserver_call_back\033[0m')
         logger.info('\033[33m raw request values: %s \033[0m',
-                    dict(request.values))
+                    dict(sanitized_request_values))
 
         query_id = hashlib.sha224(str(request.values).encode()).hexdigest()[:8]
 
@@ -532,9 +534,11 @@ def resolve_job_url():
 
 @app.route('/call_back', methods=['POST', 'GET'])
 def dataserver_call_back():
+    sanitized_request_values = sanitize_dict_before_log(request.values)
+
     logger.info('\033[32m===========================> dataserver_call_back\033[0m')
 
-    logger.info('\033[33m raw request values: %s \033[0m', dict(request.values))
+    logger.info('\033[33m raw request values: %s \033[0m', dict(sanitized_request_values))
 
     query_id = hashlib.sha224(str(request.values).encode()).hexdigest()[:8]
 
--- a/cdci_data_analysis/flask_app/app.py
+++ b/cdci_data_analysis/flask_app/app.py
@@ -601,8 +601,11 @@ class Product(Resource):
 
 @app.route('/resolve_name', methods=['GET'])
 def resolve_name():
-    logger.info("request.args: %s ", request.args)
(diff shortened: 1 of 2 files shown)


cdci_data_analysis/flask_app/app.py
```diff
@@ -1020,9 +1022,10 @@
 @app.route('/delete_product_to_gallery', methods=['POST'])
 def delete_product_to_gallery():
     logger.info("request.args: %s ", request.args)
+    logger.info("request.values: %s ", request.values)
```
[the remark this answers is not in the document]
volodymyrss: maybe it should be sanitized, i.e. cleaned of things like CR, new line, etc.

volodymyrss (changes requested):
Sounds good, this should be straightforward,  except for this log sanitization advice form QL.

burnout87:
> Sounds good, this should be straightforward, except for this log sanitization advice form QL.

defined a function that sanitizes the values of the request before logging, applied also to the `run_analysis` and `call_back` endpoints

what do you think?

cdci_data_analysis/flask_app/app.py
```diff
@@ -223,6 +223,16 @@ def remove_nested_keys(D, keys):
     return D
 
 
+def sanitize_dict_before_log(dict_to_sanitize):
+    sensitive_keys = ['token']  # Add any other sensitive keys here
+    sanitized_values = {}
+    for key, value in dict_to_sanitize.items():
+        if key not in sensitive_keys:
+            value = str(value).replace('\n', '').replace('\r', '')
```
volodymyrss, line 231: I would rather replace anything but allowed set of characters.

burnout87: something like using a regex that matches any character that is not a letter, number, or space ?

volodymyrss: yes

cdci_data_analysis/flask_app/app.py
```diff
@@ -223,6 +223,19 @@ def remove_nested_keys(D, keys):
     return D
 
 
+def sanitize_dict_before_log(dict_to_sanitize):
+    sensitive_keys = ['token']  # Add any other sensitive keys here
+    allowed_characters = r'[^A-Za-z0-9 ]'
```
dsavchenko, line 228: This regex matches NOT allowed characters, right?

cdci_data_analysis/flask_app/app.py
```diff
@@ -930,13 +947,17 @@ def get_data_product_list_by_source_name():
...
-    logger.info("request.args: %s ", request.args)
+    par_dic = request.values.to_dict()
+    sanitized_par_dic = sanitize_dict_before_log(par_dic)
+
+    logger.info("request.values: %s ", sanitized_par_dic)
     logger.info("request.files: %s ", request.files)
 
-    token = request.args.get('token', None)
+    token = par_dic.get('token', None)
+    par_dic.pop('token')
```
dsavchenko, line 959: Pop method returns value and has `default` parameter either, so this may be a one-liner

cdci_data_analysis/flask_app/app.py
burnout87: yes,I will rename the regex var

dsavchenko (approved):

volodymyrss (approved):

PR #706, opened 2024-08-21
Plot tools in oda_api

dsavchenko:
Following https://github.com/oda-hub/oda_api/pull/279
Removing code duplication in plot_tools


dsavchenko:
@volodymyrss there are unige-related connection errors in tests

volodymyrss:
Some UNIGE servers are down for an upgrade today. The tests fetch test products from there. Could you maybe add a label and conditionally skip them to see that all else is good?..

dsavchenko:
> Could you maybe add a label and conditionally skip them to see that all else is good?..

Well, this PR is almost empty now, and it is currently to check compatibility of dependencies.
So we can just wait for these servers

volodymyrss:
> > Could you maybe add a label and conditionally skip them to see that all else is good?..
> 
> Well, this PR is almost empty now, and it is currently to check compatibility of dependencies. So we can just wait for these servers

Normally the maintainance should be over, so perhaps there was a problem. I asked.

burnout87:
The failing tests appear to be purely related to a failed import, as expected.

Shall we move on with this change? If so, this will have to be coordinated with a number of other small changes in some plugins, related to the import of the classes that have been moved out from the dispatcher, and would now be in the `oda_api`. As this can be seen in this [PR](https://github.com/oda-hub/oda_api/pull/279), those classes are now there, this would allow us to avoid the `oda_api` to have the dispatcher as a dependency.

Also, the dispatcher still fixes the `bokeh` version to `2.4.2`, to match it with the one in the frontend, but after some tests (eg in the gallery) the upgrade to the latest version (ie `3.6.2`) seems to be working. But that will need an additional coordination for the frontend. Shall we upgrade this too? Or leave it for a later change to apply?

burnout87:
> > Example of a plugin adaptation [oda-hub/dispatcher-plugin-integral#135](https://github.com/oda-hub/dispatcher-plugin-integral/pull/135)
> 
> We can make changes backward-conpatible with plugins by not deleting `cdci_data_analysis.analysis.plot_tools`, rather importing from `oda_api` there

Yes, in this way we avoid having to change the majority of the plugins, but,  it is not cleanest solution...and changing the various plugins is not a big effort. Which consequences could we have?

dsavchenko:
> Yes, in this way we avoid having to change the majority of the plugins, but, it is not cleanest solution...and changing the various plugins is not a big effort. Which consequences could we have?

May be temporary, to simplify coordination. Then plugins may be adapted gradually and a module completely removed afterwards.

burnout87 (dismissed):

burnout87 (approved):

volodymyrss (approved):

burnout87 (approved):

PR #708, opened 2024-08-29
load-frontend-fits-file-uri

burnout87:


dsavchenko:
So the difference from the existing file upload endpoint is that this basically acts as middleware that appends required headers, but doesn't persist the file itself?

dsavchenko (approved):
Well, this approach will work as long as dispatcher and frontend have the same Origin (i.e. served on the same domain. Which is the case in our deployments, the dispatcher is proxied through frontend's webserver at /dispatch-data endpoint.) 
So there is no need to set any additional header.

francoismg (approved):

francoismg:
> Well, this approach will work as long as dispatcher and frontend have the same Origin (i.e. served on the same domain. Which is the case in our deployments, the dispatcher is proxied through frontend's webserver at /dispatch-data endpoint.) So there is no need to set any additional header.

Yes and if it were to change it would be easy to check the origin against a list of authorized origins and set the origin in the header response if there's a match

volodymyrss (approved):

PR #712, opened 2024-10-02
in case local resolver fails, fallback to the external resolver

burnout87:


burnout87:
In case the local resolver is not reachable, then the external resolver is used.

This is the same approach used by the frontend

volodymyrss (approved):

dsavchenko (approved):
I wasn't aware resolver is used from the dispatcher. Is this only gallery-related?

burnout87:
> I wasn't aware resolver is used from the dispatcher. Is this only gallery-related?

This one is, at the moment, used for the gallery

PR #713, opened 2024-10-07
Lock mechanism for scratch_dir creation, small refactoring

burnout87:
Use a lock and loop to attempt to acquire the lock to check and create the scratch dir, also with a small refactoring


dsavchenko (approved):

burnout87:
This approach has the downside of creating several empty lock files. After discussing with @dsavchenko  about it, we agreed on extending the `free-up-space` to use the same logic to delete the lock files

volodymyrss (approved):

PR #718, opened 2024-11-15
passing token to the funciotn that builds the uploaded file url

burnout87:


burnout87:
needs https://github.com/oda-hub/dispatcher-plugin-nb2workflow/pull/119

volodymyrss:
Wasn't the mechanism that the token is added by nb2workflow?

burnout87:
> Wasn't the mechanism that the token is added by nb2workflow?

Yes, but as noted [here](https://github.com/oda-hub/nb2workflow/issues/206), we were wondering how to handle that corner case.

volodymyrss (approved):

PR #731, opened 2025-01-20
Special treatment empty `POSIXPath` parameters

burnout87:


burnout87:
Along with  https://github.com/oda-hub/nb2workflow/pull/229

burnout87:
With this PR, the arg `force_default_value` for the `Parameter` class is introduced: if this is set, then we the default valus is forced and applied, but only in the case such parameter is not part of the input form.

burnout87:
The goal, is to obtain this

image

burnout87:
Following https://github.com/oda-hub/nb2workflow/pull/229  with this PR, we set a default value for the POSIXPath parameters in the case this is not included in the request.

This is, because we want to avoid the situation described in https://github.com/esg-epfl-apc/galaxy/issues/4#issuecomment-2582976841

cdci_data_analysis/plugins/dummy_plugin/empty_instrument_with_posix_path.py
```diff
@@ -0,0 +1,144 @@
...
+                                  catalog_name='user_catalog')
+
+    empty_query = EmptyProductQuery('empty_with_posix_path_parameters_dummy_query',)
+    failing_query = FailingProductQuery('failing_parameters_dummy_query',)
+    # let's build a simple parameter to its list
+    p = BoundaryFloat(value=10., name='p', units='W',)
+    numerical_query = DataServerNumericQuery('numerical_parameters_dummy_query',
+                                             parameters_list=[p])
+
+    f = POSIXPath(value=None, name='dummy_POSIX_file', is_optional=True)
```
dsavchenko, line 85: Do we really need all other parameters/queries in this dummy instrument?

dsavchenko (changes requested):

cdci_data_analysis/plugins/dummy_plugin/empty_instrument_with_posix_path.py
burnout87: not really, I just copied from another instrument

volodymyrss (approved):

PR #733, opened 2025-01-27
Error message for done query

burnout87:


burnout87:
Given the decision taken in https://github.com/oda-hub/dispatcher-plugin-nb2workflow/pull/134 , another PR, specific to sentry filtering for `AnalysisError` will be created

PR #734, opened 2025-01-27
Open gallery endpoints

burnout87:
Following the request from @francoismg  and @ferrigno , the following endpoints no longer require the token:

* `resolve_name`
* `get_data_product_list_with_conditions` (it will be optional as a further check will be implemented in the gallery)
* `get_data_product_list_by_source_name` (along with a small refactoring: it uses now, the more general drupal endpoint to get a list of `data_product`s given a certain set of conditions)


volodymyrss (dismissed):

volodymyrss (approved):

ferrigno (approved):
Cannot comment the code, ok for the principle

volodymyrss (approved):

PR #739, opened 2025-02-03
Increase time and the number of attempts for the scratch dir creation

burnout87:


cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -939,13 +941,14 @@ def set_scratch_dir(self, session_id, job_id=None, verbose=False):
                     break
             except (OSError, IOError) as io_e:
                 scratch_dir_created = False
-                self.logger.warning(f'Failed to acquire lock for the scratch directory creation, attempt number {attempt + 1} ({scratch_dir_retry_attempts - (attempt + 1)} left), sleeping {scratch_dir_retry_delay} seconds until retry.\nError: {str(io_e)}')
+                self.logger.warning(f'Failed to acquire lock for the scratch directory "{wd}" creation, attempt number {attempt + 1} ({scratch_dir_retry_attempts - (attempt + 1)} left), sleeping {scratch_dir_retry_delay} seconds until retry.\nError: {str(io_e)}')
+                scratch_dir_retry_delay *= 2
```
dsavchenko, line 945: To increase after sleep (otherwise, the first delay is already 0.2*2).

Also, the overall time to wait for this loop in the worst case is very long `\sum_0^9 (0.2 * 2**n) = 0.2*(2**(9+1)-1) = 204.8 s`. May lead to request timeout

cdci_data_analysis/flask_app/dispatcher_query.py
```diff
@@ -924,6 +924,8 @@ def set_scratch_dir(self, session_id, job_id=None, verbose=False):
         if job_id is not None:
             wd += '_jid_'+job_id
 
+        scratch_dir_retry_delay = starting_scratch_dir_retry_delay
```
dsavchenko, line 927: Minor, but you don't need an additional variable here

dsavchenko (changes requested):

cdci_data_analysis/flask_app/dispatcher_query.py
burnout87: good point, better to reduce the number of attempts

dsavchenko (approved):

volodymyrss (approved):

PR #740, opened 2025-02-03
inspect also the LogRecord within the hint when filter for sentry

burnout87:


dsavchenko (approved):
So it was sent twice because of sentry-logger integration.
Do we need a `send_message` then overall? 

Anyway, approving as it works.

burnout87:
> Do we need a `send_message` then overall?
> 

Good question, probably not really, in this case, but then it would imply  refactoring at many levels

PR #747, opened 2025-03-14
handling case of `dx` and `dy` None or all nan

burnout87:


burnout87:
With @dsavchenko , we implemented the handling of this corner case....what do you think @volodymyrss @ferrigno  ?

burnout87:
example

image

dsavchenko:
Meanwhile, @andriineronov, you may want to inspect the workflow notebook itself. 
In the output, with all defaults, there is only one valid point, but there are more elements in the array, with NaN values. There is also no time bin "errorbars" set in the lightcurve object. 
Not sure, if this is really expected

ferrigno:
I think that it would be easier to set `dx` and `dy` to zeros
```
if dx is None or np.isnan(dx).all():
    dx = np.zeros(len(x))
```
while leaving the x_range unchanged, and equivalent for dy, 
This will also preserve the ability of plotting error bars without additional tests.

ferrigno (approved):

volodymyrss (approved):

PR #750, opened 2025-05-16
Extended inspect state

burnout87:


burnout87:
A new endpoint is defined here, `inspect-user-state`, this behaves like the `inspect-state` one, with the difference that it returns the content of the scratch_dirs related only to the requests submitted by a certain user.  For that, the email address within the token is checked.

dsavchenko (approved):
Not sure what is the use-case. It's not very broad, right? If you want to debug a problem in someone's jobs, you will need to pass this person's token.

But OK overall

burnout87:
> Not sure what is the use-case. It's not very broad, right? 

It comes from a request from @francoismg, it is needed for the development of the new mmoda front-end

> If you want to debug a problem in someone's jobs, you will need to pass this person's token.
> 

And it is indeed what the new endpoint expects

PR #753, opened 2025-06-06
downgrade psycopg2 version

burnout87:


burnout87:
not needed, since the fix is https://gitlab.astro.unige.ch/oda/flux-mmoda-dstic/-/merge_requests/6

PR #755, opened 2025-06-30
endpoint using cors

burnout87:


cdci_data_analysis/flask_app/app.py
```diff
@@ -672,6 +677,38 @@ def load_frontend_fits_file_url():
...
 
+@app.route('/oauth_access_token_request', methods=['GET'])
+def oauth_access_token_request():
+    par_dic = request.values.to_dict()
+    logger.info('\033[32m===========================> oauth_access_token_request\033[0m')
+
+    client_id = par_dic.get('client_id', None)
+    code = par_dic.get('code', None)
+    redirect_uri = par_dic.get('redirect_uri', None)
+    client_secret = par_dic.get('client_secret', None)
```
dsavchenko, line 688: Given that this endpoint is meant to be used by the new client-only frontend, the client secret should never be there. 
If using the dispatcher as a backend helper in the authentication process anyway, the secret must be stored here.

cdci_data_analysis/flask_app/app.py
```diff
@@ -604,6 +604,11 @@ def validate_schema(response):
             'error': repr(e),
             'invalid_response': response.json
         }), 500
+    # TODO improve this
+    response.headers.add('Access-Control-Allow-Origin', '*')
+    response.headers.add('Access-Control-Allow-Headers', 'Content-Type,Authorization')
+    response.headers.add('Access-Control-Allow-Methods', 'GET,POST,PUT,DELETE,OPTIONS')
```
dsavchenko, line 610: I would be extra-careful when setting CORS headers for every response of the app. Better to set for the one that really requires it.

dsavchenko (changes requested):
I'm not sure if dispatcher is really a place for this kind of helper. Some thin dedicated backend component of the frontend may serve better.
Anyway, we need to carefully review the whole auth machinery of the app first. 

At least, see the comments above

cdci_data_analysis/flask_app/app.py
burnout87: this is, obviously, a first version, that I right now use only for tests, of course it has to better handled

cdci_data_analysis/flask_app/app.py
burnout87: secrets is now extracted fro the dispatcher config

dsavchenko (dismissed):

dsavchenko (approved):

PR #766, opened 2025-10-24
Attempt to mitigate the file locking conflict 

dsavchenko:
It may be that file listing in `get_existing_job_ID_path` takes time when there are many scratch dirs.
This PR moves all no-write operations outside the "file locked" block, so only write operation (directory creation) is done with file lock.

### DIFF
--- a/cdci_data_analysis/flask_app/dispatcher_query.py
+++ b/cdci_data_analysis/flask_app/dispatcher_query.py
@@ -946,17 +946,18 @@ class InstrumentQueryBackEnd:
         if job_id is not None:
             wd += '_jid_'+job_id
 
+        alias_workdir = self.get_existing_job_ID_path(wd=FilePath(file_dir=wd).path)
+        if alias_workdir is not None:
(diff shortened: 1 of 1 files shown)


burnout87:
> The cause of locking issues seems to be really very long globbing of the scratch directory. It became so due to a big number of the integral backend log/exception files stored there. For now, it's mitigated in prod by just moving them into subfolder.
> 
> Here, the test case may not really reproduce the situation we have in reality, so the exception produced in this PR is about aliasing conflict. This aliasing is a bit convoluted, I mark this PR as draft for a moment
> 
> Some cleanup/archiving of these files could be useful. Also, just having a subdir will already help a lot [oda-hub/dispatcher-plugin-integral#142](https://github.com/oda-hub/dispatcher-plugin-integral/issues/142).
> 
> Better solution could be to rely on an object db instead of filesystem, but this implies bigger refactoring. Need to explore this option

interesting, thanks for looking into it more in depth

